PR #23, opened 2015-08-27
Do not install requirements_dev.txt in HBP CI.

juanchopanza:
These dependencies should already get installed by the
ContinuousIntegration package.


juanchopanza:
retest this please

PR #30, opened 2015-08-28
Refactoring of neuron classes.

juanchopanza:
Make standard and ezy neurons more uniform. Viewing functions now work for both.

### DIFF
--- a/examples/log_neuron.py
+++ b/examples/log_neuron.py
@@ -99,7 +99,7 @@ if __name__ == '__main__':
                                        for p in nrn.soma.iter()])
 
     LOG.info('Neuron tree init points, types')
-    for tt in nrn.neurite_trees:
+    for tt in nrn.neurites:
         LOG.info('%s, %s', tt.value[COLS.ID], tt.value[COLS.TYPE])
 
     LOG.info('Making neuron 2')
@@ -108,10 +108,10 @@ if __name__ == '__main__':
     LOG.debug('Neuron 2 soma points %s', [as_point(p)
                                           for p in nrn2.soma.iter()])
     LOG.info('Neuron 2 tree init points, types')
-    for tt in nrn2.neurite_trees:
+    for tt in nrn2.neurites:
         LOG.info('%s %s', tt.value[COLS.ID], tt.value[COLS.TYPE])
 
     LOG.debug('Print neuron leaves as points')
-    for tt in nrn2.neurite_trees:
+    for tt in nrn2.neurites:
         for p in point_iter(tree.ileaf(tt)):
             LOG.debug(p)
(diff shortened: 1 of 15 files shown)


juanchopanza:
@mgeplf Can you have a look? This is re-factoring. We can leave more fundamental design issues for later.

examples/log_neuron.py
```diff
@@ -99,7 +99,7 @@ def point_iter(iterator):
                                        for p in nrn.soma.iter()])
 
     LOG.info('Neuron tree init points, types')
-    for tt in nrn.neurite_trees:
+    for tt in nrn.neurites:
```
mgeplf: fine w/ a rename, not sure I need to review that

neurom/core/neuron.py
```diff
@@ -156,8 +155,4 @@ def i_neurite(self, iterator_type, mapping=None, tree_filter=None):
             mapping: optional function to apply to the iterator's target.
             tree_filter: optional top level filter on properties of neurite tree objects.
         '''
-        nrt = (self.neurite_trees if tree_filter is None
-               else filter(tree_filter, self.neurite_trees))
-
-        chain_it = chain(*imap(iterator_type, nrt))
-        return chain_it if mapping is None else tr.imap_val(mapping, chain_it)
+        return tr.i_chain(self.neurites, iterator_type, mapping, tree_filter)
```
mgeplf: this is just proxy, right?  I would be tempted to have people be explicit, but not a big deal

neurom/ezy/neuron.py
```diff
@@ -109,6 +109,8 @@ class Neuron(object):
     def __init__(self, filename, iterable_type=np.array):
         self._iterable_type = iterable_type
         self._nrn = load_neuron(filename, set_tree_type)
+        self.soma = self._nrn.soma
```
mgeplf: depending on the type here, I would be concerned that someone might expect to change _nrn, and see show up in soma.  Or is it a container so they point to the same object?

mgeplf:
+1

neurom/ezy/neuron.py
juanchopanza: Same object, and this class owns it.

neurom/core/neuron.py
juanchopanza: I didn't get that. What do you mean?

mgeplf: typing tr.i_chain(self.neurites, ...) vs self.i_neurites_chains(....) is about the same, and is more explicit, imo.  No biggie though

mgeplf: OTOH, it's nice to hint at how an interface works, so this makes sense in that context

PR #31, opened 2015-08-28
Add neuron bounding box function.

juanchopanza:
Implemented in neurom.core.neuron.
Imported into neurom.ezy.

### DIFF
--- a/neurom/analysis/morphtree.py
+++ b/neurom/analysis/morphtree.py
@@ -231,17 +231,17 @@ def n_sections(tree):
 
 def get_bounding_box(tree):
     """
-    Returns the boundaries of the tree
-    in three dimensions:
-    [[xmin, ymin, zmin],
-    [xmax, ymax, zmax]]
+    Returns:
+        The boundaries of the tree in three dimensions:
+            [[xmin, ymin, zmin],
+            [xmax, ymax, zmax]]
     """
 
-    min_x = min(p[COLS.X] for p in val_iter(tr.ipreorder(tree)))
-    min_y = min(p[COLS.Y] for p in val_iter(tr.ipreorder(tree)))
-    min_z = min(p[COLS.Z] for p in val_iter(tr.ipreorder(tree)))
-    max_x = max(p[COLS.X] for p in val_iter(tr.ipreorder(tree)))
-    max_y = max(p[COLS.Y] for p in val_iter(tr.ipreorder(tree)))
-    max_z = max(p[COLS.Z] for p in val_iter(tr.ipreorder(tree)))
+    min_xyz, max_xyz = (np.array([np.inf, np.inf, np.inf]),
+                        np.array([np.NINF, np.NINF, np.NINF]))
 
-    return np.array([[min_x, min_y, min_z], [max_x, max_y, max_z]])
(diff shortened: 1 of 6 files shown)


juanchopanza:
@mgeplf Please have a look if you have a moment. I'm sure this can be simplified.

neurom/analysis/morphtree.py
```diff
@@ -237,11 +237,11 @@ def get_bounding_box(tree):
     [xmax, ymax, zmax]]
     """
 
-    min_x = min(p[0] for p in val_iter(tr.ipreorder(tree)))
-    min_y = min(p[1] for p in val_iter(tr.ipreorder(tree)))
-    min_z = min(p[2] for p in val_iter(tr.ipreorder(tree)))
-    max_x = max(p[0] for p in val_iter(tr.ipreorder(tree)))
-    max_y = max(p[1] for p in val_iter(tr.ipreorder(tree)))
-    max_z = max(p[2] for p in val_iter(tr.ipreorder(tree)))
+    min_x = min(p[COLS.X] for p in val_iter(tr.ipreorder(tree)))
```
mgeplf: This seems inefficient, couldn't you do something like:
min_value, max_value = np.array([np.INF, np.INF, np.INF), np.array([np.NINF, np.NINF, np.NINF)
for p in val_iter(tr.ipreorder(tree)):
    min_value = np.minimum(p, min_value)
    max_value = np.maximum(p, max_value)
return min_value, max_value

That what it's one traversal instead of 6

neurom/core/neuron.py
```diff
@@ -156,3 +157,38 @@ def i_neurites(self, iterator_type, mapping=None, tree_filter=None):
...
+
+def bounding_box(nrn):
+    '''Return 3D bounding box of a neuron
+
+    Returns:
+        tuple of ((min_x, min_y, min_z), (max_x, max_y, max_z))
+    '''
+
+    # Get the bounding coordinates of the neurites
+    nmin_xyz = (min(c for c in tr.i_chain(nrn.neurites, tr.ipreorder,
```
mgeplf: if c is a (x, y, z) point, the same as above applies, imo

neurom/analysis/morphtree.py
juanchopanza: At the cost of building a numpy array. I'll try profiling it. The thing is, I don't know what kind of data sizes to expect. I also intend to memoize most Neuron methods, because it is supposed to be read-only.

neurom/core/neuron.py
juanchopanza: Will give it a shot.

neurom/analysis/morphtree.py
juanchopanza: Ah, that is lida's code, BTW, and not sure it will be used now that I have a neuron-wide bounding box.

neurom/core/neuron.py
mgeplf: I'm not sure the cost of making a numpy array is near the same as iterating so much.  Also, does the iteration make reference or return a new np array?  Finally, you can save one creation by doing:
np.minimum(min_value, pt, min_value)

PR #32, opened 2015-08-31
Add helper functions to load all morpholigies in a directory.

juanchopanza:
* ezy.load_neurons(directory_path) returns a list of ezy.Neuron objects,
  one for each morphology in directory_path.


neurom/ezy/__init__.py
```diff
@@ -47,6 +47,14 @@
     >>> fig3d, ax3d = ezy.view3d(nrn)
     >>> fig3d.show()
 
+    Load neurons from a directory. This loads all SWC or HDF5 files it finds\
+    and returns a list of neurons
+
+    >>> import numpy as np  # For mean value calculation
+    >>> nrns = ezy.load_neurons('some/data/directory')
+    >>> for nrn in nrns:
+            print 'mean section length', np.mean([l for l in nrn.get_section_lengths()])
```
liesbethvanherpe: Maybe be you should write "len" instead of "l" to avoid confusion here.

juanchopanza:
I went for `n` instead of `l`. I don't want that line to become too long.

PR #43, opened 2015-09-18
Simplify sum expressions.

juanchopanza:
sum(i for i in iterator) -> sum(iterator)


juanchopanza:
@liesbethvanherpe Could you have a quick look? I removed some anti-patterns from the examples.

liesbethvanherpe:
+2

PR #47, opened 2015-09-24
Add segment_taper_rate function to morphmath

juanchopanza:
* Add usage example to examples/ezy_advanced.py
* Add unit test


juanchopanza:
@liesbethvanherpe FYI, taper rate example

PR #49, opened 2015-09-24
Example: calculate and plot end-to-end distance of neurites.

liesbethvanherpe:


examples/end_to_end_distance.py
```diff
@@ -0,0 +1,112 @@
...
+    neurite trunk.'''
+    trunk = neurite.value[0:3]
+    furthest = trunk
+    max_distance = 0
+    for leaf in tree.val_iter(tree.ileaf(neurite)):
+        new_distance = morphmath.point_dist(leaf[0:3], trunk)
+        if new_distance > max_distance:
+            max_distance = new_distance
+            furthest = leaf[0:3]
+    return furthest
```
juanchopanza: I think this can be simplified to

trunk = neurite.value
return min((morphmath.point_dist(l, trunk) for l in tree.val_iter(tree.ileaf(n0)))

examples/end_to_end_distance.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def segment_vector(seg):
+    '''Calculate and return segment vector.'''
+    return np.subtract(seg[1][0:3], seg[0][0:3])
+
+
+def neurite_end_to_end_distance(neurite):
+    '''Calculate and return end-to-end-distance of a given neurite.'''
+    return morphmath.point_dist(furthest_leaf(neurite), neurite.value[0:3])
```
juanchopanza: You can just pass tree values to morphmath functions, you don't need to slice with [0:3].

examples/end_to_end_distance.py
```diff
@@ -0,0 +1,112 @@
...
+    '''Calculate and plot the end-to-end distance vs the number of segments for
+    an increasingly larger part of a given neurite.'''
+    n = sum(1 for _ in tree.isegment(neurite))
+    end_to_end_distance = np.zeros(n)
+    index = 0
+    max_so_far = 0
+    for segment in tree.val_iter(tree.isegment(neurite)):
+        new_distance = morphmath.point_dist(segment[1][0:3], neurite.value[0:3])
+        if new_distance > max_so_far:
+            max_so_far = new_distance
```
juanchopanza: This looks like a max calculation. Can it be simplified?

examples/end_to_end_distance.py
```diff
@@ -0,0 +1,112 @@
...
+
+def calculate_and_plot_end_to_end_distance(neurite):
+    '''Calculate and plot the end-to-end distance vs the number of segments for
+    an increasingly larger part of a given neurite.'''
+    n = sum(1 for _ in tree.isegment(neurite))
+    end_to_end_distance = np.zeros(n)
+    index = 0
+    max_so_far = 0
+    for segment in tree.val_iter(tree.isegment(neurite)):
+        new_distance = morphmath.point_dist(segment[1][0:3], neurite.value[0:3])
```
juanchopanza: No need for the [0:3] here.

examples/end_to_end_distance.py
juanchopanza: Maybe not, are you trying to calculate tha max out of all neurites? Then use the expression above inside another max.

liesbethvanherpe: I can see how this finds the maximum distance, but how can I get the "l" at this maximum distance?

examples/end_to_end_distance.py
```diff
@@ -0,0 +1,97 @@
...
+    '''Calculate and plot the end-to-end distance vs the number of segments for
+    an increasingly larger part of a given neurite.'''
+    n = sum(1 for _ in tree.isegment(neurite))
+    end_to_end_distance = np.zeros(n)
+    index = 0
+    max_so_far = 0
+    for segment in tree.val_iter(tree.isegment(neurite)):
+        max_so_far = max(max_so_far, morphmath.point_dist(segment[1], neurite.value))
+        end_to_end_distance[index] = max_so_far
+        index += 1
```
liesbethvanherpe: I'd like to make this as compact as the one on line 46, but I don't know how I can select an increasingly larger part of a neurite...

liesbethvanherpe:
Is this OK now?

examples/end_to_end_distance.py
```diff
@@ -0,0 +1,90 @@
...
+    plt.ylabel('End-to-end distance')
+    plt.show()
+
+
+def calculate_and_plot_end_to_end_distance(path):
+    '''Calculate and plot the end-to-end distance vs the number of segments for
+    an increasingly larger part of a given path.
+
+    Note that the plots are not very meaningful for bifurcating trees.'''
+    n = sum(1 for _ in tree.isegment(path))
```
juanchopanza: You could increment n in the loop to avoid doing two iterations.

PR #63, opened 2015-10-02
Fix SWC loader and neuron viewers for soma only data.

juanchopanza:


juanchopanza:
@liesbethvanherpe @lidakanari FYI, minor fixes to allow loading single-point somas and viewing neurons with no neurites.

PR #65, opened 2015-10-04
fix for 'deep' neurons hitting recursion limit

mgeplf:
* neurons with many segments/sections hit the
  python recursion limit, fix this
* add test & data


juanchopanza:
pep8 fail, according to travis log.

mgeplf:
Oops, not my normal setup.  Stuck in Madrid.

mgeplf:
I need to go through for other recursive calls, I think, but these are the ones I hit

PR #69, opened 2015-10-07
histogram example

eleftherioszisis:
Test merging the fork with one commit of a histogram example.


juanchopanza:
It seems it failed the pep8 linting stage. You can test this locally before pull requesting to save some time:

    make test

This will run the linting and only if this passes, the unit tests.

eleftherioszisis:
ok, I think now it's good

PR #71, opened 2015-10-07
Example: plot several somas in same plot, each with different color.

liesbethvanherpe:


examples/plot_somas.py
```diff
@@ -0,0 +1,70 @@
...
+    for file_name in file_names:
+        nrn = ezy.Neuron(file_name)
+        somas.append(nrn.soma)
+    return somas
+
+
+def plot_somas(somas):
+    '''Plot set of somas on same figure as spheres, each with different color'''
+    fig = plt.figure()
+    ax = fig.add_subplot(111, projection='3d', aspect='equal')
```
lidakanari: You can also use the common.get_figure() to create the figure:

fig, ax = common.get_figure(new_fig=True, subplot=111, params={'projection':'3d', 'aspect':'equal'}

examples/plot_somas.py
```diff
@@ -0,0 +1,79 @@
...
+        somas.append(neuron.soma)
+    return somas
+
+
+def plot_somas(neurons):
+    '''Plot set of somas on same figure as spheres, each with different color'''
+    somas = extract_somas(neurons)
+
+    fig = plt.figure()
+    ax = fig.add_subplot(111, projection='3d', aspect='equal')
```
lidakanari: You can also use common.get_figure() function to create the figure:

fig, ax = common.get_figure(new_fig=True, subplot=111, params={'projection':'3d', 'aspect':'equal'}

examples/plot_somas.py
```diff
@@ -0,0 +1,79 @@
...
+
+def load_neurons(file_names):
+    '''Load neurons from list of file names'''
+    neurons = []
+    for file_name in file_names:
+        neurons.append(ezy.Neuron(file_name))
+    return neurons
+
+
+def extract_somas(neurons):
```
juanchopanza: You can simply say

return list(neuron.soma for neuron in neurons)

instead of repeatedly appending to a list

juanchopanza: Or even return a tuple, if you don't want to modify it.

examples/plot_somas.py
```diff
@@ -0,0 +1,79 @@
...
+        center = s.center
+        radius = s.radius
+        plot_sphere(fig, ax, center, radius, color=random_color(), alpha=1)
+    plt.show()
+
+
+if __name__ == '__main__':
+    #  define set of files containing relevant neurons
+    file_nms = []
+    file_nms.append('test_data/swc/Soma_origin.swc')
```
juanchopanza: It would be good to get the path of the tests WRT this file, so that it can be run from different directories. See how it is done in tests.

liesbethvanherpe:
And now?

PR #74, opened 2015-10-07
Further improvements in HDF5 reading performance.

juanchopanza:


juanchopanza:
@mgeplf You may find this "interesting". The biggest improvement comes from using `numpy.searchsorted` instead of `bisect.bisect_right`. But I think we're starting to hit the limits of the design.

trialsbiker:
Yeah, that's not surprising.  bisect is using individual accesses to the numpy array, which will be much slower than the 'native' method.

Cool.

PR #75, opened 2015-10-08
Add filename in ezy.neuron

lidakanari:


lidakanari:
@liesbethvanherpe Could you have a look at this?

neurom/ezy/neuron.py
```diff
@@ -111,6 +111,7 @@ def __init__(self, filename, iterable_type=np.array):
         self._nrn = load_neuron(filename, set_tree_type)
         self.soma = self._nrn.soma
         self.neurites = self._nrn.neurites
+        self.filename = self._nrn.id
```
juanchopanza: I think this should be called something like "label" or "name". Also, don't we want to strip it? It could get quite long if the filename path is long.

lidakanari: Yes, this is literally the filename without the extension. If we want to keep the name and remove the path of the file I can rename this to name and keep only the name striping the path data. Is this better?

juanchopanza: Yes, I think that would be a better option. You could even do it in `core.neuron.Neuron`.

neurom/ezy/tests/test_neuron.py
```diff
@@ -73,6 +73,10 @@ def setUp(self):
         self.filename = os.path.join(SWC_PATH, 'Neuron.swc')
         self.neuron = ezy.Neuron(self.filename)
 
+    def test_name(self):
+        print self.neuron.filename
+        nt.assert_true(self.neuron.filename == os.path.join(SWC_PATH, 'Neuron'))
```
liesbethvanherpe: Maybe I'm missing something, but why 'Neuron' instead of ' Neuron.swc'?

neurom/ezy/neuron.py
lidakanari: Yes, I think it is better if we pass the path strip to the initial stage. I will update the core.neuron.Neuron instead.

neurom/ezy/tests/test_neuron.py
lidakanari: We remove the extension and keep the "real name". This comes from the neuron.id (check io.utils.load_neuron code, where this comes from.)

lidakanari:
@juanchopanza Is this good like this?

neurom/io/utils.py
```diff
@@ -116,7 +116,7 @@ def load_neuron(filename, tree_action=None):
         raise NonConsecutiveIDsError('Non consecutive IDs found in raw data')
 
     nrn = make_neuron(data, tree_action)
-    nrn.id = os.path.splitext(filename)[0]
+    nrn.id = os.path.splitext(filename)[0].split('/')[-1]
```
juanchopanza: You can use `os.path.basename` instead of `split` to make it platform independent.

lidakanari: Can I use both os.path.splitext and os.path.basename in one operation or should I use them consecutively?

juanchopanza: You can use both in one expression. 

    os.path.splitext(os.path.basename(filename))[0]

Issue #76, opened 2015-10-09
Some docstrings cause severe sphinx errors


juanchopanza:
e.g. in `NeuroM/neurom/view/common.py`:

    SEVERE: Unexpected section title.

    Options
    ----------

It seems sphinx doesn't like `----------`.

juanchopanza:
Resolved.

PR #80, opened 2015-10-09
Fix issue #76

juanchopanza:
https://github.com/BlueBrain/NeuroM/issues/76


juanchopanza:
@lidakanari FYI, I changed the format of some of the viewer docstrings while fixing a spinx error.

lidakanari:
Thanks! That seems good to me. I will apply same rules for future commits.

PR #81, opened 2015-10-09
Add example of distribution fit for soma radii.

lidakanari:
Simplified example to fit soma radii.

### DIFF
--- a/examples/soma_radius_fit
+++ b/examples/soma_radius_fit
@@ -28,7 +28,7 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-'''Extract a distribution for a selected feature of the population (list) of neurons.'''
+'''Extract a distribution for the soma radii of the population (list) of neurons.'''
 
 from neurom import ezy
 from scipy import stats
--- a/examples/soma_radius_fit
+++ b/examples/soma_radius_fit
@@ -1,0 +1,107 @@
+#!/usr/bin/env python
+
+# Copyright (c) 2015, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
+# All rights reserved.
+#
+# This file is part of NeuroM <https://github.com/BlueBrain/NeuroM>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#     1. Redistributions of source code must retain the above copyright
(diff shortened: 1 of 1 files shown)


examples/soma_radius_fit
```diff
@@ -0,0 +1,107 @@
...
+from scipy import stats
+import numpy as np
+import argparse
+import os
+
+
+def parse_args():
+    '''Parse command line arguments'''
+    parser = argparse.ArgumentParser(description='Morphology statistics extractor',
+                                     epilog='Note: Outputs json')
```
juanchopanza: Does this really output JSON?

lidakanari: Not yet. This just prints out results for the moment. Will fix this. 

Or should I write the output in json for this example instead?

juanchopanza: Up to you. We can always add that later.

lidakanari: I leave this for later, when we decide about the json format. Fixed the description and epilog accordingly.

examples/soma_radius_fit
```diff
@@ -0,0 +1,112 @@
...
+from scipy import stats
+import numpy as np
+import argparse
+import os
+
+
+def parse_args():
+    '''Parse command line arguments'''
+    parser = argparse.ArgumentParser(description='Morphology fit distribution extractor',
+                                     epilog='Note: Prints the optimal distribution and corresponding parameters.')
```
liesbethvanherpe: Could you add the list of candidates here (so norm, expon, uniform), or add that it only tests for a limited set of distributions?

examples/soma_radius_fit
```diff
@@ -0,0 +1,112 @@
...
+    '''
+    params = distribution_fit(data, distribution=distribution)
+    return getattr(stats, distribution).fit(data)[0]
+
+
+def test_multiple_distr(filepath):
+    '''Runs the distribution fit for multiple distributions and returns
+       the optimal distribution along with the corresponding parameters.
+    '''
+    #  load a neuron from an SWC file
```
liesbethvanherpe: It can be h5 too.

lidakanari: True! Actually I'm using it for h5 data ;) Thanks!

examples/soma_radius_fit
lidakanari: Should this be added here, or in the comments of the function? This can be extended to any number of distributions available in scipy.stats but I keep the distributions that are available in synthesis for the moment.

liesbethvanherpe: Maybe it should be added to line 31: this way, the user of the code is reminded of the fact that there may be another distribution out there that is even a better fit.

PR #83, opened 2015-10-09
ezy population class and respective tests

eleftherioszisis:
more functions to be added

### DIFF
--- a/neurom/ezy/population.py
+++ b/neurom/ezy/population.py
@@ -1,0 +1,75 @@
+# Copyright (c) 2015, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
+# All rights reserved.
+#
+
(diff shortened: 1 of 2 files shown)


neurom/ezy/population.py
```diff
@@ -0,0 +1,75 @@
...
+        return sum(nrn.get_n_neurites(neurite_type=neurite_type) for nrn in self.iter_neurons())
+
+    def iter_neurites(self):
+        '''
+        Iterate over the neurites
+
+            Returns:
+                Iterator of neurite tree iterators
+        '''
+        return chain(self.neurites)
```
juanchopanza: Hasn't neurites already been flattened in `neurom.core.Population`?

eleftherioszisis: Indeed they have already been. changed to iter

PR #93, opened 2015-10-13
Fix the example soma_radius_fit

lidakanari:
I removed the logging, since this is taken care by the load_neurons and there is no need to check the directory again.


juanchopanza:
This needs a new travis environment. Closing, so it can be re-submitted.

PR #99, opened 2015-10-14
Update requirements.txt to include scipy.

lidakanari:
Minor change in requirements.txt

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,6 +30,7 @@
 #numpy>=1.8.0
 #h5py>=2.2.1
 #matplotlib>=1.3.1
+#scipy>=0.13.3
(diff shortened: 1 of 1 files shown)


juanchopanza:
Can you change the commit message? It is confusing because that requirement is commented out. It is only there for documentation purposes.

lidakanari:
Agreed. Just please let me know how can I do this. 

I tried to 

git commit --amend -m "..."

but the push failed.

juanchopanza:
I think you need a new commit. Or maybe you can edit the comment here.

lidakanari:
Edited here. Let me know if it is more comprehensible now.

PR #100, opened 2015-10-14
dendrogram example

eleftherioszisis:
A first approach to a dendrogram plot where diameters can be shows as well if desired. Let me know if changes are required. Not validated against other dendrogram plots, thus I am not sure that it works 100% correctly.

### DIFF
--- a/examples/dendrogram.py
+++ b/examples/dendrogram.py
@@ -1,22 +1,54 @@
+#!/usr/bin/env python
+# Copyright (c) 2015, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
+# All rights reserved.
+#
+# This file is part of NeuroM <https://github.com/BlueBrain/NeuroM>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#     1. Redistributions of source code must retain the above copyright
+#        notice, this list of conditions and the following disclaimer.
+#     2. Redistributions in binary form must reproduce the above copyright
+#        notice, this list of conditions and the following disclaimer in the
+#        documentation and/or other materials provided with the distribution.
+#     3. Neither the name of the copyright holder nor the names of
+#        its contributors may be used to endorse or promote products
+#        derived from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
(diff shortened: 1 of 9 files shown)


examples/dendrogram.py
```diff
@@ -0,0 +1,163 @@
...
+    segments_dict = {tree_object.value[-2]: [(0., 0.), 0.]}
+
+    positions = []
+    diameters = []
+
+    for seg in isegment(tree_object):
+
+        tr_pos = get_transformed_position(seg, segments_dict, y_length)
+
+        seg_diam = seg[0].value[3] + seg[1].value[3]
```
juanchopanza: We have symbols for the elements of tree values defined somewhere (search for `COLS.X`, `COLS.R` etc.). These could be used to avoid magic numbers.

examples/dendrogram.py
```diff
@@ -0,0 +1,163 @@
...
+
+'''Moprhology Dendrogram Functions Example'''
+
+from neurom.view import common
+from neurom.core.tree import isegment
+from matplotlib.collections import LineCollection
+from numpy import sqrt
+
+
+def get_length(segment):
```
juanchopanza: We already have a segment length function somewhere, so this one can be removed.

eleftherioszisis:
No sqrt in numpy module. This is interesting. :P 
Btw if anyone knows how I can easily rotate a line collection in order to have a rotated dendrogram, please let me know.

examples/dendrogram.py
```diff
@@ -0,0 +1,163 @@
...
+            The first item is either a figure object (if subplot is False) or an
+            axis object. The second item is an object containing the data used to
+            generate the figure. The final item is text used in report generation
+            as a figure legend. This text needs to be manually entered in each
+            figure file.
+    '''
+    # get line segment positions and respective diameters
+    positions, linewidths = dendro_transform(tree_object)
+
+    linewidths = linewidths / min(linewidths)
```
juanchopanza: This could be

    linewidths = linewidths / min(linewidths) if show_diameters else 1

That way you make only one assignment.

lidakanari:
If you want to have an y-x instead of x-y graph simply change the order of the coordinates, when you create the segment list:

segments_dict[end_node_id] = [(x1, y1), ...   -> segments_dict[end_node_id] = [(y1, x1), ...

juanchopanza:
Re. sqrt, it might be a `pylint` problem. You may have to disable the error (or I might have to force a newer version of `plyint`. Does it work on your desktop?

eleftherioszisis:
Yes it works in my desktop but I removed the sqrt anyway because I used the morphmath segment_length function.

eleftherioszisis:
don't merge it yet. the segment length function throws an error. brb

eleftherioszisis:
ok, now it works. Is there a prettier way than segment_length(list(val_iter(segment))) ?

examples/dendrogram.py
```diff
@@ -0,0 +1,150 @@
...
+    root_id = tree_object.value[-2]
+
+    y_length = {root_id: 1.}
+
+    segments_dict = {tree_object.value[-2]: [(0., 0.), 0.]}
+
+    positions = []
+    diameters = []
+
+    for seg in isegment(tree_object):
```
juanchopanza: you could apply the value transformation here, instead of using x.value and so on further down the line. Something like

    for seg in val_iter(isegment(tree_object)):

        seg_diam = seg[0][3] + seg[1][3]

eleftherioszisis: Well the problem of transforming to the value arrays from the beginning is that I will lose the info about the children which I need in get_transformed_position.

juanchopanza: Yes, I see that now. I think for now you can leave it like this (although it makes more sense to use a tuple than a list). We can see later if there is an easier way of doing this.

juanchopanza:
I will merge it as is. We can refine it later.

PR #101, opened 2015-10-14
Example to extract distribution of feature in json format.

lidakanari:
This example can be called to extract the fitted distribution from a selected feature:
./examples/distribution_to_json.py /path/to/data/ section_lengths

Sample outputs of three distribution types

{
  "type": "uniform",
  "min": 0.88831300789755407,
  "max": 647.43531343775476,
  "description": "Random text"
}

{
  "type": "normal",
  "mu": 7.4974040499781536,
  "sigma": 0.69349297860265591,
  "min": 6.7262619633790655,
  "max": 8.407764268817246,
  "description": "Random text"
}

{
  "type": "exponential",
  "lambda": 66.351635544015153,
  "min": 0.10295630140984792,
  "max": 941.35384973935652,
  "description": "Random text"
}

### DIFF
--- a/examples/distribution_to_json.py
+++ b/examples/distribution_to_json.py
@@ -56,32 +56,28 @@ def parse_args():
     return parser.parse_args()
 
 
-def distribution_fit(data, distribution='norm'):
-    '''Calculates and returns the parameters of a distribution'''
-    return getattr(stats, distribution).fit(data)
-
-
-def distribution_error(data, distribution='norm'):
-    '''Calculates and returns the distance of a fitted distribution
-       from the initial data.
+def distribution_fit_error(data, distribution='norm'):
+    '''Calculates and returns the parameters and the ks-distance
+        of a fitted distribution from the initial data.
     '''
-    params = distribution_fit(data, distribution=distribution)
-    return stats.kstest(data, distribution, params)[0]
+    params = getattr(stats, distribution).fit(data)
+    return params, stats.kstest(data, distribution, params)[0]
 
 
 def test_multiple_distr(data):
     '''Runs the distribution fit for multiple distributions and returns
        the optimal distribution along with the corresponding parameters.
+       Fit normal returns (mean, std).
+       Fit exponential returns (loc, scale=1/lambda).
+       Fit uniform returns (mean, std).
     '''
     distr_to_check = ['norm', 'expon', 'uniform']
 
-    fit_data = {d: distribution_fit(data, d) for d in distr_to_check}
+    fit_all = {d: distribution_fit_error(data, d) for d in distr_to_check}
 
-    fit_error = {distribution_error(data, d): d for d in distr_to_check}
+    optimal = fit_all.keys()[np.argmin([error[1] for error in fit_all.values()])]
 
-    optimal = fit_error.values()[np.argmax(fit_error.iterkeys())]
-
-    return optimal, fit_data[optimal]
+    return optimal, fit_all[optimal][0]
 
 
 def extract_data(data_path, feature):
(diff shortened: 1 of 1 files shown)


lidakanari:
@juanchopanza @liesbethvanherpe Could you please check that this conforms to the json format that we want to use for each distribution type? When we finalize this, I can build from this example to generate the initial feature extractor.

examples/distribution_to_json.py
```diff
@@ -0,0 +1,148 @@
...
+    data_dict = OrderedDict()
+
+    if data["type"] == 'norm':
+        data_dict.update({"type": "normal"})
+        data_dict.update({"mu": data["params"][0]})
+        data_dict.update({"sigma": data["params"][1]})
+
+    elif data["type"] == 'expon':
+        data_dict.update({"type": "exponential"})
+        data_dict.update({"lambda": data["params"][1]})
```
juanchopanza: Just curious. What is `data["params"][0]` for an exponential?

examples/distribution_to_json.py
```diff
@@ -0,0 +1,148 @@
...
+
+    elif data["type"] == 'uniform':
+        data_dict.update({"type": "uniform"})
+
+    data_dict.update({"min": data["min"]})
+    data_dict.update({"max": data["max"]})
+    data_dict.update({"description": "Random text"})
```
juanchopanza: Note the description isn't necessary, and the `min` and `max` parameters are optional.

juanchopanza:
The format looks fine to me.

examples/distribution_to_json.py
```diff
@@ -0,0 +1,148 @@
...
+
+def test_multiple_distr(data):
+    '''Runs the distribution fit for multiple distributions and returns
+       the optimal distribution along with the corresponding parameters.
+    '''
+    distr_to_check = ['norm', 'expon', 'uniform']
+
+    fit_data = {d: distribution_fit(data, d) for d in distr_to_check}
+
+    fit_error = {distribution_error(data, d): d for d in distr_to_check}
```
juanchopanza: Question: does this require that we fit the distribution twice? I don't see the result of `distribution_fit` being used in `distribution_error`.

examples/distribution_to_json.py
lidakanari: Ok, this needs to be hardcoded depending on the distribution. The problem is that each distribution fit in scipy returns a list of parameters. What each parameter means is unfortunately defined by the distribution you fit and there is no global way of defining this concept. Please don't merge this change before I make sure that for the exponential and normal I use the output of scipy in the right order. I will include a reference for each.

examples/distribution_to_json.py
lidakanari: Ok, I add the min, max for completeness. Do you want me to include a description only if the user adds a description upon the call of the script?

examples/distribution_to_json.py
lidakanari: The problem here is that we need to fit the data and find the error of the fit. I could merge the distribution fit and the distribution error into one that will return both. 

I think that would be optimal.

juanchopanza: Right. I don't see a problem there. You can do both in one call.

lidakanari: Ok, will add that.

examples/distribution_to_json.py
lidakanari: So for the exponential fit the first parameter is the location (not sure here, if it starts from zero???) and the second the scale. The scale is supposed to be 1/lambda, right?

juanchopanza: I'm not sure about scipy. We expect lambda to be stored in the synthesizer json input file. Shall I merge this?

lidakanari: We are not using the 'loc' parameter for the model. If the synthesizer needs the lambda then the output 1/scale is ok. Yes, this can be merged.

PR #104, opened 2015-10-14
added rotation to dendrogram and histogram for features over populations

eleftherioszisis:


eleftherioszisis:
Travis does not like numpy :(

examples/dendrogram.py
```diff
@@ -135,6 +155,31 @@ def dendrogram(tree_object, show_diameters=False, new_fig=True, subplot=False, *
...
     linewidths = linewidths / min(linewidths) if show_diameters else 1
 
+    xlabel = 'Length (um)'
+    ylabel = ''
+
+    if rotation == 'left':
+
+        angle = pi
+
+        affine2D_transfrom(positions, cos(angle), -sin(angle), sin(angle), cos(angle))
```
juanchopanza: This line is repeated in 3 places. I think it can be moved out of the if-else blocks.

examples/histogram.py
```diff
@@ -84,10 +85,88 @@ def histogram(neurons, feature, new_fig=True, subplot=False, **kwargs):
...
+        Number of bins for the histogram.
+
+        cumulative : bool
+        Sets cumulative histogram on.
+
+        subplot : bool
+            Default is False, which returns a matplotlib figure object. If True,
+            returns a matplotlib axis object, for use as a subplot.
+
+    Returns
```
juanchopanza: Parameters, Options and Returns will need `:` in order for the docs to be build correctly. I am not sure if "Options" will work either.

juanchopanza:
I updated the version of pylint. Hopefully that will fix the numpy problems next time around.

PR #105, opened 2015-10-14
Mock out h5py, scipy for doc build. Use newer pylint.

juanchopanza:


juanchopanza:
@mgeplf It looks like this fixed the readthedocs problem.

mgeplf:
Nice, good catch.

PR #107, opened 2015-10-15
Add simple stats module.

juanchopanza:
Consists of helper functions that invoke scipy stats analysis
functionality.


juanchopanza:
@lidakanari It is pretty simple, but does it look OK?

PR #108, opened 2015-10-15
Add optimal distribution fit in stats and Fix pylint errors with numpy in local make test

lidakanari:


neurom/stats.py
```diff
@@ -48,3 +49,28 @@ def fit(data, distribution='norm'):
...
+    Returns:
+        tuple (optimal, parameters) of data to fitted distribution
+
+    Note:
+        Uses Kolmogorov-Smirnov test to estimate distance and p-value.
+
+        The distributions that are tested are uniform, normal and exponential.
+    '''
+
+    distr_to_check = ['norm', 'expon', 'uniform']
```
juanchopanza: Maybe this should be an optional input?

neurom/stats.py
```diff
@@ -48,3 +49,28 @@ def fit(data, distribution='norm'):
...
+
+    fit_d = {d: fit(data, d) for d in distr_to_check}
+
+    optimal = fit_d.keys()[_np.argmin([error[1][0] for error in fit_d.values()])]
+
+    return optimal, fit_d[optimal][0]
```
juanchopanza: Maybe it makes sense to return some estimate of the goodness of fit?

neurom/tests/test_stats.py
```diff
@@ -63,3 +63,9 @@ def test_fit_default_is_normal():
     p1, e1 = st.fit(NORMAL, 'norm')
     nt.assert_items_equal(p0, p1)
     nt.assert_items_equal(e0, e1)
+
+def test_optimal_distribution():
```
juanchopanza: I think here it makes sense to test whether it will find the optimal distribution, e.g. if you pass an exponential or a uniform you get that back (assuming a large enough sample.)

neurom/stats.py
```diff
@@ -48,3 +49,28 @@ def fit(data, distribution='norm'):
...
+        Uses Kolmogorov-Smirnov test to estimate distance and p-value.
+
+        The distributions that are tested are uniform, normal and exponential.
+    '''
+
+    distr_to_check = ['norm', 'expon', 'uniform']
+
+    fit_d = {d: fit(data, d) for d in distr_to_check}
+
+    optimal = fit_d.keys()[_np.argmin([error[1][0] for error in fit_d.values()])]
```
juanchopanza: I think the name "error" is confusing here, because `fit_d.values()` holds the results of the fits, right?

neurom/stats.py
lidakanari: Yes, that could be useful for generalization.

neurom/tests/test_stats.py
lidakanari: That's what line 63 is doing for normal. I will add the same for exponential and uniform.

neurom/stats.py
lidakanari: True, it is confusing this way. The error is actually the error[1] part, so I will rename this one.

neurom/stats.py
lidakanari: Does this really worths it? Should I return the fit_d[optimal] instead then?

juanchopanza: Well, I think it is good to know if the fit is really bad. In fact, I don't like this idea of throwing distributions at the data until we find one that matches very much, but that is (almost) a different issue.

lidakanari: Yes, I agree, but this is a different issue :) If you want to use one of the predefined distributions you need to have a measure of it. So I'll just return the full fit data here.

neurom/stats.py
```diff
@@ -48,3 +51,26 @@ def fit(data, distribution='norm'):
     '''
     params = getattr(_st, distribution).fit(data)
     return params, _st.kstest(data, distribution, params)
+
+
+def optimal_distribution(data, distr_to_check=('norm', 'expon', 'uniform')):
```
juanchopanza: I have to figure out if there are any potential problems with this default parameter, but if there are we can fix it later. So +2 from me. Shall I merge?

lidakanari: OK from me. If you have a better way we can improve it later.

PR #111, opened 2015-10-15
Example json extractor for an mtype.

lidakanari:
Needs refinement and cleaning, but please check if it produces the correct output that is recognized by the synthesizer.

### DIFF
--- a/examples/synthesis_json.py
+++ b/examples/synthesis_json.py
@@ -81,13 +81,11 @@ def extract_data(data_path, feature, params=None):
 
     results["type"] = opt_fit[0]
     results["params"] = opt_fit[1][0]
-    results["min"] = np.min(feature_data)
-    results["max"] = np.max(feature_data)
 
     return results
 
 
-def transform_distribution(data):
+def transform_distribution(data, datamin=None, datamax=None):
     '''Transform optimal distribution data into correct format
        based on the distribution type.
     '''
@@ -105,8 +103,10 @@ def transform_distribution(data):
     elif data["type"] == 'uniform':
         data_dict.update({"type": "uniform"})
 
-    data_dict.update({"min": data["min"]})
-    data_dict.update({"max": data["max"]})
+    if datamin is not None:
+        data_dict.update({"min": datamin})
+    if datamax is not None:
+        data_dict.update({"max": datamax})
 
     return data_dict
 
@@ -131,7 +131,7 @@ def transform_package(data_path, feature_list, feature_names, component):
 
         data_dict["components"].setdefault(comp)
 
-    for feature, comp, name in zip(feature_list, component, feature_names):
+    for feature, name, comp in zip(feature_list, component, feature_names):
 
         result = transform_distribution(extract_data(data_path, feature))
         data_dict["components"][comp] = {name: result}
--- a/examples/synthesis_json.py
+++ b/examples/synthesis_json.py
@@ -133,9 +133,9 @@ def transform_package(data_path, components, feature_list):
     '''
     data_dict = transform_header(os.path.basename(data_path), components)
 
-    for feature, name, comp, fmin, fmax in feature_list:
+    for feature, name, comp, fmin, fmax, fparam in feature_list:
 
-        result = transform_distribution(extract_data(data_path, feature), fmin, fmax)
+        result = transform_distribution(extract_data(data_path, feature, fparam), fmin, fmax)
         data_dict["components"][comp] = {name: result}
 
     return data_dict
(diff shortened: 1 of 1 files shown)


examples/synthesis_json.py
```diff
@@ -0,0 +1,152 @@
...
+    try:
+        opt_fit = stats.optimal_distribution(feature_data)
+    except ValueError:
+        from itertools import chain
+        feature_data = list(chain(*feature_data))
+        opt_fit = stats.optimal_distribution(feature_data)
+
+    results["type"] = opt_fit[0]
+    results["params"] = opt_fit[1][0]
+    results["min"] = np.min(feature_data)
```
juanchopanza: I think we need a flag to say if we want to keep min and max. It doesn't always make sense to take these from the data (they could for example mark biological or physical boundaries). But if they get put in the file, they get used. Technically, I think min and max should be used as input to the fit, not be an output.

lidakanari: Ok, so the default will be without min and max and those will be inserted only in case the user defines a min and max.

examples/synthesis_json.py
```diff
@@ -0,0 +1,159 @@
...
+        data_dict.update({"sigma": data["params"][1]})
+
+    elif data["type"] == 'expon':
+        data_dict.update({"type": "exponential"})
+        data_dict.update({"lambda": 1. / data["params"][1]})
+
+    elif data["type"] == 'uniform':
+        data_dict.update({"type": "uniform"})
+
+    if datamin is not None:
```
juanchopanza: I think the uniform distribution should always have min and max, because those are the fitter parameters, no?

lidakanari: That's right! Thanks, I will fix this.

juanchopanza: Do you think this can be done in time for the demo?

lidakanari: Yes, I'm working on this now. However, I don't think that we should have the same name for the uniform parameters and the min max that are defined by the user. To avoid confusion, I will use [lower bound, upper bound] for now for the uniform and we can discuss this later. Is that ok for you?

lidakanari:
Please don't merge yet. I need to properly handle the Inf values that might occur in distributions.

examples/synthesis_json.py
```diff
@@ -0,0 +1,162 @@
...
+    elif data["type"] == 'expon':
+        data_dict.update({"type": "exponential"})
+        if data["params"][1] != 0:
+            data_dict.update({"lambda": 1. / data["params"][1]})
+        else:
+            data_dict.update({"scale": 0.})
+
+    elif data["type"] == 'uniform':
+        data_dict.update({"type": "uniform"})
+        data_dict.update({"lower_bound": data["params"][0]})
```
juanchopanza: How does this get changed to "min" and "max" in the resulting JSON?

lidakanari: They shouldn't be changed. If the min max are user defined parameters, then they should not change the values you get for the distribution fit. We need to define a new name for the parameter of uniform distribution.

juanchopanza: I don't agree. We have min/max as input parameters for morphine, and there really is no reason to change that.

juanchopanza: In any case, we shouldn't be changing the structure of the JSON without discussion, just before a demo :-)

lidakanari: Those are parameters outputted from the distribution fit. When min max were defined by the data, those two were the same, so there was no reason for a different name. 

Now, there is a new decision that min and max are defined by the user. It doesn't make sense to have the same name for user defined parameters and data extracted parameters. Which one is more important to be outputted and how can the program judge which one is more important the real data or the user's decision?

juanchopanza: Simple: a user should not be allowed to enter min and max when fitting a uniform distribution. That should be an error because it makes no sense.

lidakanari: But the user does not define the distribution. The optimal is selected according to minimum ks-distance.

juanchopanza: OK, so the range parameters can be ignored if the optimal distribution is uniform.

lidakanari: Ok, that is one option. However, this is an important decision and it will have a significant impact on the output. 

I can save it as (min, max) for now, so that it works with morphine, but I strongly believe that we should not mix the names of user decisions with the extracted from biological data parameters.

examples/synthesis_json.py
```diff
@@ -0,0 +1,162 @@
...
+            data_dict.update({"lambda": 1. / data["params"][1]})
+        else:
+            data_dict.update({"scale": 0.})
+
+    if datamin is not None:
+        data_dict.update({"min": datamin})
+    if datamax is not None:
+        data_dict.update({"max": datamax})
+
+    elif data["type"] == 'uniform':
```
juanchopanza: Now if datamin or datamax are not None and the distrubution is normal, you set and unset things twice. There is no need for that. Just restrict the datamin/datamax thing to distrutions that are not "normal".

juanchopanza: I'll merge this and make changes later.

PR #114, opened 2015-10-16
Fix uniform paramaters fit.

lidakanari:

PR #115, opened 2015-10-16
Avoid some unnecessary calls to OrderedDict.update and simplify

juanchopanza:
min/max handling of normal distribution.

### DIFF
--- a/examples/synthesis_json.py
+++ b/examples/synthesis_json.py
@@ -92,18 +92,18 @@ def transform_distribution(data, datamin=None, datamax=None):
     data_dict = OrderedDict()
 
     if data["type"] == 'uniform':
-        data_dict.update({"type": "uniform",
-                          "min": np.min(data["params"]),
-                          "max": np.max(data["params"])})
+        data_dict["type"] = "uniform"
+        data_dict["min"] = np.min(data["params"])
+        data_dict["max"] = np.max(data["params"])
         return data_dict
 
     elif data["type"] == 'norm':
-        data_dict.update({"type": "normal",
-                          "mu": data["params"][0],
-                          "sigma": data["params"][1]})
+        data_dict["type"] = "normal"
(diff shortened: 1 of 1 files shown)


examples/synthesis_json.py
```diff
@@ -91,27 +91,28 @@ def transform_distribution(data, datamin=None, datamax=None):
...
 
-    if data["type"] == 'norm':
-        data_dict.update({"type": "normal"})
-        data_dict.update({"mu": data["params"][0]})
-        data_dict.update({"sigma": data["params"][1]})
+    if data["type"] == 'uniform':
+        data_dict.update({"type": "uniform",
+                          "min": np.min(data["params"]),
+                          "max": np.max(data["params"])})
+        return data_dict
```
juanchopanza: The logic here is that min/max are the parameters of the uniform distribution, so we ignore datamin/datamax.

examples/synthesis_json.py
```diff
@@ -91,27 +91,28 @@ def transform_distribution(data, datamin=None, datamax=None):
...
     data_dict = OrderedDict()
 
-    if data["type"] == 'norm':
-        data_dict.update({"type": "normal"})
-        data_dict.update({"mu": data["params"][0]})
-        data_dict.update({"sigma": data["params"][1]})
+    if data["type"] == 'uniform':
+        data_dict.update({"type": "uniform",
+                          "min": np.min(data["params"]),
+                          "max": np.max(data["params"])})
```
juanchopanza: This use of `np,min`, `np.max` is taken from #114

juanchopanza:
That is fine, because JSON objects are unordered anyway. But it might be nice to keep an order for presentation purposes. I'll change this.

juanchopanza:
Is this OK?

lidakanari:
This is ok I guess, if the output that it produces is good. However, you will have to show me how to erase my last commits without breaking the merge!

Issue #116, opened 2015-10-16
neurom.ezy.load_neurons to work with list of filenames


juanchopanza:
Currently we have

* `neurom.ezy.load_neuron(filename)`: loads from a file name
* `neurom.ezy.load_neurons(directory_name)`: loads set of neurons from all morphology files in a directory.

It would be very useful to have `neurom.ezy.load_neurons` work with a list of filenames.

juanchopanza:
Pull request #118 implements this and similar functionality for `ezy.load_population`.

PR #118, opened 2015-10-19
ezy.load_neurons and ezy.load_population also work with lists of file…

juanchopanza:
…names.

* Allow optional name parameter for load_population


juanchopanza:
Implements issue #116.

PR #122, opened 2015-10-21
examples/synthesis_json.py can take multiple file path inputs.

juanchopanza:
Runs analysis for each file path separately.

TODO: Write out JSON files for each iteration
TODO: Find naming scheme for when files do not have m-type in their name
TODO: Store results of analyses in list to user can use them in
interactive session

### DIFF
--- a/examples/synthesis_json.py
+++ b/examples/synthesis_json.py
@@ -51,8 +51,9 @@ def parse_args():
         epilog='Note: Outputs json of the optimal distribution \
                 and corresponding parameters.')
 
-    parser.add_argument('datapath',
-                        help='Path to morphology data file or directory')
+    parser.add_argument('datapaths',
+                        nargs='+',
+                        help='Morphology data directory paths')
 
     return parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


juanchopanza:
@lidakanari This might be of interest.

lidakanari:
TODO: Find naming scheme for when files do not have m-type in their name

-> That would be my comment as well, in order to make it work for biological data. Could we make the example take as input a datapath (or multiple) and the corresponding type for each directory if it is provided?

juanchopanza:
Yes, I am trying to figure out the least confusing way of providing this functionality via the commandline. I hope to have something ready later today.

examples/synthesis_json.py
```diff
@@ -113,31 +114,33 @@ def transform_package(mtype, files, components, feature_list):
...
-    d_path = args.datapath
+    data_dirs = args.datapaths
 
     mtype_files = defaultdict(list)
 
-    for f in get_morph_files(d_path):
-        mtype_files[get_mtype(f)].append(f)
-
     flist = [["soma_radius", "radius", "soma", None, None, None],
              ["n_neurites", "number", "basal_dendrite", 1, None,
```
liesbethvanherpe: For my education: what do "n_neurites" and "number" mean here?

(n_neurites is probably number of neurites, but what is number?)

lidakanari: The first is the NeuroM name, the second is the morphine readable name chosen in this commit:

https://bbpcode.epfl.ch/code/#/c/17824/

They both mean the same thing, but we need different names to be recognizable by the corresponding software.

PR #128, opened 2015-10-21
Fix bug in view and improve documentation

lidakanari:
View documentation might still needs improvements, but this can be done in a separate merge. However, comments of what can be improved will be useful.

### DIFF
--- a/neurom/view/view.py
+++ b/neurom/view/view.py
@@ -60,67 +60,58 @@ def tree(tr, plane='xy', new_fig=True, subplot=False, **kwargs):
     '''Generates a 2d figure of the tree.
 
     Parameters:
-        tr: Tree
-        neurom.Tree object
+        tr: Tree \
+            neurom.Tree object
 
     Options:
-        plane: str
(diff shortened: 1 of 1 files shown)


neurom/view/view.py
```diff
@@ -186,60 +170,47 @@ def _seg_2d(seg):
...
-        If True the figure will be scaled according to tree limits.
-        Default value is False.
-
-    Returns
-    --------
-    A 2D matplotlib figure with a soma view, at the selected plane.
-
-    """
-
+    Options:
```
juanchopanza: This is better, but see http://sphinxcontrib-napoleon.readthedocs.org/en/latest/example_google.html. Also, I don't think "Options" will be recognised. These should be in "Parameters", with an indication that they are optional. If in doubt, you can build the docs locally and see how it looks. Anyway, I am OK with merging this and fixing it later.

lidakanari: Yes, that will be the next step. Thanks!

PR #129, opened 2015-10-22
Correct the subplot option in neuron viewer

lidakanari:
The subplot option was not propagating to the soma and tree level correctly, when the neuron was plotted. This fixes the subplot problem.


juanchopanza:
We need to figure out a way of testing this and/or re-factoring it.

lidakanari:
Yes, indeed! This needs complete tests in place...

PR #133, opened 2015-10-22
Add constant distribution in synthesis_json

lidakanari:
This should fix artificial normal distributions with zero sigma.

### DIFF
--- a/examples/synthesis_json.py
+++ b/examples/synthesis_json.py
@@ -96,10 +96,8 @@ def transform_package(mtype, files, components, feature_list):
                                            fmin, fmax)
 
         # When the distribution is normal with sigma = 0 it will be replaced with constant
-        if result['type'] == 'normal' and result['sigma']==0.0:
-            replace_result = OrderedDict()
-            replace_result['type'] = 'constant'
-            replace_result['val'] = result['mu']
+        if result['type'] == 'normal' and result['sigma'] == 0.0:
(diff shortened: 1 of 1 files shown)


examples/synthesis_json.py
```diff
@@ -95,6 +95,13 @@ def transform_package(mtype, files, components, feature_list):
         result = stats.fit_results_to_dict(extract_data(files, feature, fparam),
                                            fmin, fmax)
 
+        # When the distribution is normal with sigma = 0 it will be replaced with constant
+        if result['type'] == 'normal' and result['sigma']==0.0:
+            replace_result = OrderedDict()
```
juanchopanza: You can simplify this a bit and avoid some `__setitem__` calls:

    replace_result = OrderedDict((('type', 'constant'), ('val', result['mu'])))

examples/synthesis_json.py
```diff
@@ -95,6 +95,13 @@ def transform_package(mtype, files, components, feature_list):
         result = stats.fit_results_to_dict(extract_data(files, feature, fparam),
                                            fmin, fmax)
 
+        # When the distribution is normal with sigma = 0 it will be replaced with constant
+        if result['type'] == 'normal' and result['sigma']==0.0:
```
juanchopanza: I guess this will fail the linting. A good reminder to run locally first!

examples/synthesis_json.py
lidakanari: Ok, fixed this.

PR #135, opened 2015-10-23
Ipython Notebook from neurom demo documentation

eleftherioszisis:
Please let me know what else should I include. I believe that the installation part must be removed from the notebook and instead to put how to install notebook and run the kernels.


juanchopanza:
I think this shouldn't go in the top level. Maybe put it in an ipynb directory?

PR #136, opened 2015-10-23
Trunk diameter and length ready to be used.

lidakanari:


neurom/analysis/morphtree.py
```diff
@@ -247,6 +247,16 @@ def n_terminations(tree):
     return sum(1 for _ in tr.ileaf(tree))
 
 
+def trunk_diameter(tree):
```
juanchopanza: diameter != radius

neurom/analysis/morphtree.py
```diff
@@ -247,6 +247,16 @@ def n_terminations(tree):
...
+    '''Return the initial diameter of the tree segments'''
+    return i_segment_radius(tree).next()
+
+
+def trunk_length(tree):
```
juanchopanza: Is the trunk really defined as the first segment? I thought it was the first section.

PR #138, opened 2015-10-23
bugfix: analysis.trunk_radius and analysis.trunk_length

juanchopanza:
Fix and improve tests.


examples/synthesis_json.py
```diff
@@ -69,7 +72,6 @@ def extract_data(files, feature, params=None):
     try:
```
juanchopanza: There is a problem with this bit of code, which has prevented me from fitting trunk_radius.

lidakanari: The problem is in the exception it catches...

juanchopanza: The problem was in `ezy.Neuron`. Fixed.

PR #139, opened 2015-10-23
Add taper rate to JSON synthesizer file maker

juanchopanza:


juanchopanza:
@lidakanari @liesbethvanherpe This gives 0.0 for neurons synthesized with the latest synthesizer. Other than that, it isn't tested. I'll merge unless there are any objections.

PR #141, opened 2015-10-27
Add trunk vector, elevation and azimuth functions.

lidakanari:
@juanchopanza @liesbethvanherpe 

Please check the correctness of the implementation and the 'special cases' : 

1. In trunk elevation I need to throw an Exception when the soma center coincides with the first point of the tree!
2. In trunk azimuth I need to return pi/2 when the x coordinate is the same, to avoid the division by zero error.

### DIFF
--- a/neurom/analysis/morphtree.py
+++ b/neurom/analysis/morphtree.py
@@ -273,7 +273,10 @@ def trunk_vector(tree, soma):
 
 def trunk_elevation(tree, soma):
     '''Angle between x-axis and vector defined by (initial tree point - soma center)
-       on the x-y plane.
+       on the x-y half-plane.
+
+       Returns:
+           Angle in radiants between -pi/2 and pi/2
     '''
     vector = trunk_vector(tree, soma)
 
@@ -286,6 +289,9 @@ def trunk_elevation(tree, soma):
 def trunk_azimuth(tree, soma):
     '''Angle between x-axis and vector defined by (initial tree point - soma center)
        on the x-z plane.
+
+       Returns:
+           Angle in radiants between -pi and pi
     '''
     vector = trunk_vector(tree, soma)
 
--- a/neurom/analysis/morphtree.py
+++ b/neurom/analysis/morphtree.py
@@ -264,9 +264,9 @@ def trunk_length(tree):
         return 0.0
 
 
-def trunk_vector(tree, soma):
-    '''Vector of trunk defined as (initial tree point - soma center)
-       of the tree.
+def trunk_direction(tree, soma):
+    '''Vector of trunk direction defined as
+       (initial tree point - soma center) of the tree.
     '''
     return mm.vector(tree.value, soma.center)
 
@@ -276,9 +276,9 @@ def trunk_elevation(tree, soma):
        on the x-y half-plane.
 
        Returns:
-           Angle in radiants between -pi/2 and pi/2
+           Angle in radians between -pi/2 and pi/2
     '''
-    vector = trunk_vector(tree, soma)
+    vector = trunk_direction(tree, soma)
 
     if np.linalg.norm(vector) != 0:
         return np.arcsin(vector[COLS.Y] / np.linalg.norm(vector))
@@ -291,9 +291,9 @@ def trunk_azimuth(tree, soma):
        on the x-z plane.
 
        Returns:
-           Angle in radiants between -pi and pi
+           Angle in radians between -pi and pi
     '''
-    vector = trunk_vector(tree, soma)
+    vector = trunk_direction(tree, soma)
 
     return np.arctan2(vector[COLS.Z], vector[COLS.X])
 
--- a/neurom/analysis/morphtree.py
+++ b/neurom/analysis/morphtree.py
@@ -252,6 +252,13 @@ def trunk_radius(tree):
     return tree.value[COLS.R]
 
 
+def trunk_vector(tree, soma):
+    '''Vector of trunk defined as (soma center - initial point)
+       of the tree.
+    '''
+    return mm.vector(tree.value, soma.center)
+
+
 def trunk_length(tree):
     '''Length of the initial tree section
 
--- a/neurom/analysis/morphtree.py
+++ b/neurom/analysis/morphtree.py
@@ -265,19 +265,22 @@ def trunk_length(tree):
 
 
 def trunk_vector(tree, soma):
-    '''Vector of trunk defined as (soma center - initial point)
+    '''Vector of trunk defined as (initial tree point - soma center)
        of the tree.
     '''
     return mm.vector(tree.value, soma.center)
 
 
 def trunk_elevation(tree, soma):
-    '''Angle between x-axis and vector defined by (soma center - initial point)
+    '''Angle between x-axis and vector defined by (initial tree point - soma center)
        on the x-y plane.
     '''
     vector = trunk_vector(tree, soma)
 
-    return np.arcsin(vector[COLS.Y] / np.linalg.norm(vector))
+    if np.linalg.norm(vector) != 0:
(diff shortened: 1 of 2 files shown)


neurom/analysis/morphtree.py
```diff
@@ -264,6 +264,37 @@ def trunk_length(tree):
...
+def trunk_azimuth(tree, soma):
+    '''Angle between x-axis and vector defined by (initial tree point - soma center)
+       on the x-z plane.
+    '''
+    vector = trunk_vector(tree, soma)
+
+    if vector[COLS.X] != 0.0:
+        return np.arctan(vector[COLS.Z] / vector[COLS.X])
+    else:
+        return np.pi / 2
```
liesbethvanherpe: It could also be -np.pi/2 here.

neurom/analysis/morphtree.py
```diff
@@ -264,6 +264,37 @@ def trunk_length(tree):
...
+
+
+def trunk_azimuth(tree, soma):
+    '''Angle between x-axis and vector defined by (initial tree point - soma center)
+       on the x-z plane.
+    '''
+    vector = trunk_vector(tree, soma)
+
+    if vector[COLS.X] != 0.0:
+        return np.arctan(vector[COLS.Z] / vector[COLS.X])
```
liesbethvanherpe: The range of arctan is -pi/2, +pi/2, and we should cover a full circle... (+ pi depending on signs?)

lidakanari: Good point! I will use this instead:

http://docs.scipy.org/doc/numpy/reference/generated/numpy.arctan2.html

juanchopanza: It might be a good idea to specify this corner case in the docstring.

neurom/analysis/morphtree.py
juanchopanza: Whatever it is, it might also be a good idea to specify this corner case in the docstring.

lidakanari: Yes! I will correct that.

neurom/analysis/morphtree.py
```diff
@@ -264,6 +264,40 @@ def trunk_length(tree):
         return 0.0
 
 
+def trunk_vector(tree, soma):
```
juanchopanza: Wouldn't `trunk_direction` make more sense?

neurom/analysis/morphtree.py
```diff
@@ -264,6 +264,40 @@ def trunk_length(tree):
...
+    '''
+    return mm.vector(tree.value, soma.center)
+
+
+def trunk_elevation(tree, soma):
+    '''Angle between x-axis and vector defined by (initial tree point - soma center)
+       on the x-y half-plane.
+
+       Returns:
+           Angle in radiants between -pi/2 and pi/2
```
liesbethvanherpe: * radians

neurom/analysis/morphtree.py
```diff
@@ -264,6 +264,40 @@ def trunk_length(tree):
...
 
 
+def trunk_vector(tree, soma):
+    '''Vector of trunk defined as (initial tree point - soma center)
+       of the tree.
+    '''
+    return mm.vector(tree.value, soma.center)
+
+
+def trunk_elevation(tree, soma):
```
liesbethvanherpe: (The following may be a question of taste) Actually, I use trunk elevation and trunk azimuth for the orientation of the first segment, and initial elevation and initial azimuth for the positioning of the first cross-section of a neurite.

neurom/analysis/morphtree.py
```diff
@@ -264,6 +264,40 @@ def trunk_length(tree):
...
+    else:
+        raise ValueError("Vector between soma center and tree is [0, 0, 0]")
+
+
+def trunk_azimuth(tree, soma):
+    '''Angle between x-axis and vector defined by (initial tree point - soma center)
+       on the x-z plane.
+
+       Returns:
+           Angle in radiants between -pi and pi
```
liesbethvanherpe: radiants -> radians

neurom/analysis/morphtree.py
lidakanari: Thanks!

neurom/analysis/morphtree.py
lidakanari: No problem for me. If this makes more sense I can change it to trunk_direction.

neurom/analysis/morphtree.py
lidakanari: This should be the same (assuming that trunk is tangent to the sphere of course), so I don't see any reason to distinguish them.

juanchopanza: I think `trunk_direction` is easier to understand. It is a direction vector, it isn't an angle WRT another vector.

lidakanari: Ok, coming in next commit.

liesbethvanherpe: I meant that for me, trunk_direction is vector(second_cross_section - first_cross_section), and initial_direction is vector(first_cross_section - soma.center).

If I understand correctly, both will be the same?

lidakanari: I though you were referring to the trunk_elevation and trunk_azimuth (see your comment above) and those should be the same. If your question refers to trunk_direction or initial_direction I am open to more changes depending on what everyone prefers. After all it is a matter of taste (and definition) as you put it and I don't have a problem with either of the proposed names. I would prefer to keep trunk_elevation and trunk_azimuth though.

liesbethvanherpe: Better keep "trunk_elevation", "trunk_azimuth" and "trunk_direction" together and not mix with "initial"! :-)

So, for my education: "trunk" means "the first cross section", and not "the first segment".

juanchopanza: Trunk should not mean first cross section. If this is what we think it means, then we need to use a word other than trunk. In fact, the work trunk makes almost no sense in this context.

lidakanari: Well trunk is a commonly used term in neuroscience (at least to my knowledge, so please correct me if this is not accurate). It is True that the way this term is used is generally very confusing...

So, in general...
Trunk is the initial section (so no segment and no point). When we refer to the orientation of a tree (elevation and azimuth) we generally refer to the vector (initial point - soma center) so the initial orientation. So when you refer to trunk elevation and trunk azimuth to me this is equivalent to the tree elevation and azimuth which implies a reference to the first point.

So, we can use any of the following, as long as the definitions are clear!
trunk_ (elevation, azimuth, direction)
tree_ (elevation, azimuth, direction)
initial_ (elevation, azimuth, direction)

neurom/analysis/morphtree.py
```diff
@@ -264,6 +264,40 @@ def trunk_length(tree):
...
+def trunk_elevation(tree, soma):
+    '''Angle between x-axis and vector defined by (initial tree point - soma center)
+       on the x-y half-plane.
+
+       Returns:
+           Angle in radians between -pi/2 and pi/2
+    '''
+    vector = trunk_direction(tree, soma)
+
+    if np.linalg.norm(vector) != 0:
```
berthelm: Is it a good idea to compare the magnitude of a vector with exactly 0?  There should be a precision involved for such a test I think. 

if np.linalg.norm(vector) > epsilon
...

lidakanari: Thanks! So a common practice in python seems to be the use of sys.float_info.epsilon . Is it better if I use

np.linalg.norm(vector) <=  sys.float_info.epsilon

instead of comparison to zero?

juanchopanza: It is better than 0, but you should check if numpy provides similar epsilon numbers that depend on the type of the expression being evaluated (`np.linalg.norm(v)`)

PR #142, opened 2015-10-30
neuron equality operator

eleftherioszisis:
Implementation of equality operator for trees and neurons. If two trees or neurons are identical. then n1 == n2 returns true. Because there is no sorting in the order that the trees are created the comparison finds first the pair of trees that have identical leaves and checks the nodes in an upstream manner.

It is slow when comparing big trees (eg axon) and I think the only way to become faster is if a sorting rule is applied when trees are generated from the file. Then, there will be no need to check if the children of a node are swapped depending on which is read first from the file.

I needed this comparison operator in order to make sure that a neuron that goes through a process remains unchanged (for example in the repair process where everything is switched off) and also it might come handy in order to check if different formats that describe the same neuron generate the same internal representation in NeuroM


neurom/ezy/neuron.py
```diff
@@ -110,6 +112,12 @@ def __init__(self, neuron, iterable_type=np.array):
         super(Neuron, self).__init__(neuron.soma, neuron.neurites, neuron.name)
         self._iterable_type = iterable_type
 
+    def __eq__(self, other):
+        return False if not isinstance(self, type(other)) else \
+               all(self._compare_neurites(other, ttype) for ttype in
```
juanchopanza: Why do you have to split them by type? Wouldn't it be easier to compare all neurites?

eleftherioszisis: Because this is faster due to the fact that if I compare the neurites of one neuron and the neurites of another neuron then I will have (NBASAL + NAPICAL + NUNDEFINED + 1)^2 possible combinations (I do not assume that the order the neurites are stored is sorted), while if I separate them by time I will not have for example to compare the axon with the basals etc. having thus NBASAL^2 + NAPICAL^2 + NUNDEFINED^2 + 1 combinations. I hope I did the math correctly :P

juanchopanza: That makes sense. I wonder if there are other ways to segment the neurons further. I'll merge this.

PR #143, opened 2015-11-04
Add partition in morphtree

lidakanari:


neurom/analysis/morphtree.py
```diff
@@ -300,6 +300,16 @@ def trunk_azimuth(tree, soma):
     return np.arctan2(vector[COLS.Z], vector[COLS.X])
 
 
+def partition(tree):
+    '''Ratio of maximum number of sections
```
juanchopanza: Could you explain better what this returns?

juanchopanza:
Which lines are you referring to? Could you add an inline comment in the right place?

neurom/analysis/morphtree.py
```diff
@@ -300,6 +300,16 @@ def trunk_azimuth(tree, soma):
...
+       over minimum number of sections at
+       each bifurcation point.
+    '''
+    return [np.max([n_sections(i.children[0]), n_sections(i.children[1])]) /
+            np.min([n_sections(i.children[0]), n_sections(i.children[1])])
```
lidakanari: Is there a better way to minimize the computation and repetition in the previous two lines?

lidakanari: Lines 308-309 since it doesn't seem to put it in the right place.

juanchopanza: You can define a helper function inside the function, to call n_sections only once per child:

    def partition_at_point(bif_point):
        n = n_sections(bif_point.children[0])
        m = n_sections(bif_point.children[1])
        return max(n, m) / min(n,m)

Also, no idea why you are using `np.max`. All it seems to do is force you to build lists for no reason.

lidakanari: Ok, I was thinking to add this into morphmath if it makes sense and then call it from morphtree.

juanchopanza: I think it relies on the tree structure, so it should probably stay in `morphtree`.

PR #144, opened 2015-11-04
Example script to plot histogram of some features

juanchopanza:
The idea is to have separate plots for each feature/mtype combination, and superpose the simulation's input distribution. Only the first part is done. Kind of.


lidakanari:
Looks ok to me!

PR #145, opened 2015-11-05
Feature plotter with uniform distribution

juanchopanza:
Note: both the histogram and the distribution are normalized. If the histogram diverges wildly from the distribution it may be difficult to make out one or the other.

### DIFF
--- a/examples/plot_features.py
+++ b/examples/plot_features.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-
 # Copyright (c) 2015, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
 # All rights reserved.
 #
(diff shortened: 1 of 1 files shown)


juanchopanza:
@lidakanari, @liesbethvanherpe this could be of interest. It is still quite rough. If you have any ideas how to show a constant "distribution" nicely it would be good to know.

juanchopanza:
@lidakanari @liesbethvanherpe Currently this makes an array of plot objects called `plots`. If anyone knows how to make figures with many plots in them out of this, please let me know how (or make the necessary changes if this is easier.)

lidakanari:
I'm not sure how one can make a figure out of subfigures... However, there is a way to combine axes into one large figure. I can show you how to do this if you think it worths it, but when many sub-plots are combined into one figure it gets kind of messy and the individual plots are usually not presentable...

PR #146, opened 2015-11-06
Example to check if neurites in a neuron are flat within a tolerance in microns

eleftherioszisis:


examples/check_planar.py
```diff
@@ -0,0 +1,111 @@
...
+        Input
+            tree : a tree object
+
+            tol : tolerance in microns
+
+        Returns
+            extend : int
+    '''
+    # extract the x,y,z coordinates of all the points in the tree
+    points = np.array([value[0:3] for value in val_iter(ipreorder(tree))])
```
juanchopanza: Here you could use symbols instead of magic numbers, e.g.

    ....
    points = np.array([value[COLS.X: COLS.R] .... 

Which reminds me, I should put this COLS thing in a higher-level module.

examples/check_planar.py
```diff
@@ -0,0 +1,111 @@
...
+    '''
+
+    # calculate the covariance of the points
+    cov = np.cov(points)
+
+    # find the principal components
+    return np.linalg.eig(cov)
+
+
+def extend_of_tree(tree):
```
juanchopanza: Is it "extent" rather than "extend"?

PR #147, opened 2015-11-06
Feature plotter: save all plots to PDF file.

juanchopanza:


juanchopanza:
@lidakanari @liesbethvanherpe I decided to dump everything in a PDF. This means we can use this even if plotting doesn't work on the macbooks!

lidakanari:
I think that is a good idea indeed! Pdf is a much safer solution.

PR #148, opened 2015-11-06
example to check if a neuron is monotonic.

eleftherioszisis:


examples/dendrogram.py
```diff
@@ -37,6 +37,8 @@
 from neurom.analysis.morphtree import n_segments, n_bifurcations, n_terminations
 
 from matplotlib.collections import PolyCollection
+import sys
+sys.setrecursionlimit(10000)
```
juanchopanza: We should try to avoid this.

examples/dendrogram.py
```diff
@@ -146,8 +148,7 @@ def _dendrogram(neuron_, show_diameters=True):
     '''Main dendrogram function
     '''
     from copy import deepcopy
-    import sys
-    sys.setrecursionlimit(3000)
+
     # neuron is copied because otherwise the tree modifications that follow
     # will be applied to our original object
     neuron = deepcopy(neuron_)
```
juanchopanza: Maybe we can explore copying the raw data array, then re-building the neuron. This should avoid the recursion limit problem.

examples/check_monotonicity.py
```diff
@@ -0,0 +1,72 @@
...
+        Input
+            neuron : neuron object
+
+        tol
+            numerical precision
+    '''
+
+    for i, neurite in enumerate(neuron.neurites):
+
+        print '{0} {1:30} {2}'.format(i, neurite.type, is_monotonic(neurite, tol))
```
juanchopanza: I know this is an example, but it would make sense to make it into a checker app, or part of the morph_check app at some point. So better to get used to using loggers and log levels for output.

eleftherioszisis:
Please ignore the failing of checks. Pylint fails on my mac, thus I cannot complete all the tests. I will do it tomorrow through my desktop machine.

This is the error:
  File "/Users/zisis/Projects/NeuroM/neurom_test_venv/lib/python2.7/locale.py", line 475, in _parse_localename
    raise ValueError, 'unknown locale: %s' % localename
ValueError: unknown locale: UTF-8

examples/check_monotonicity.py
eleftherioszisis: ok. I will remove then the function and after we discuss its proper location I will write it properly there, using the logger.

eleftherioszisis: (*the check monotonicity function*)

juanchopanza:
I think I had this once. See [this stackoverflow post for a possible solution](http://stackoverflow.com/questions/19961239/pelican-3-3-pelican-quickstart-error-valueerror-unknown-locale-utf-8).

eleftherioszisis:
Thanks, that did the trick. However, pylint now throws an error for importing h5py:

************* Module neurom.io.hdf5
F: 45, 0: Unable to import 'h5py' (import-error)
************* Module neurom.view.common
I: 44, 0: Locally disabling unused-import (W0611) (locally-disabled)
************* Module read_h5_v1
F:  3, 0: Unable to import 'h5py' (import-error)

h5py imports normally through ipython

juanchopanza:
That should not be a problem if you have installed h5py in your system. If this works

    python -c "import h5py"

then `make test` should not fail in that manner.

examples/dendrogram.py
eleftherioszisis: Yep, I agree. I also do not like the recursive nature of the dendrogram creation and I will probably make it iterative at some point if possible but it will need a significant amount of time.

Can we leave the recursion limit like this for the time being (I think it is probably needed by the dendrogram as well) and for example first implement the copying of the neuron in the core folder?

juanchopanza: Yes, that is fine for now.

eleftherioszisis:
Yes, it works, but pylint still gives that error :/

eleftherioszisis:
Ok, at least here the tests are passed :P

PR #157, opened 2015-11-17
make_copy function example

eleftherioszisis:
I created a function to copy trees as an example. @juanchopanza I would like your input for my implementation. I used the logic that is used in make_tree, thus I am parsing the original tree and creating the copy in the same time.

Please let me know if changes are needed in order to make it more efficient. Unfortunately, I cannot think of a way to do it with iterators as it seems that the behavior is peculiar when I add to a tree while its iterator runs in a for loop.

Finally, as far as its placement is concerned, should it go in the core.tree? I think in this implementation there is no assumptions for the tree values or number of chidlren.


juanchopanza:
Wouldn't this leave two trees referring to the same node data?

eleftherioszisis:
Yes, you are right. I ll copy the value.

juanchopanza:
You could add some tests to a main function to ensure everything works OK.

eleftherioszisis:
Well, I will anyway implement the tests when I ll move it inside neurom. Should I also add some here?

juanchopanza:
I think it will make your life easier, but it is up to you.

eleftherioszisis:
So, is it ok to move it to core.tree?

juanchopanza:
Yes, with tests.

neurom/analysis/tests/test_morphtree.py
```diff
@@ -445,11 +445,18 @@ def test_principal_direction_extent():
     tree.children[0].add_child(Tree(np.array([points[2][0], points[2][1], points[2][2], 1., 0., 0.])))
     tree.children[0].add_child(Tree(np.array([points[3][0], points[3][1], points[3][2], 1., 0., 0.])))
 
-    extent = principal_direction_extent(tree)
+    extent = mtr.principal_direction_extent(tree)
 
     nt.assert_true(np.allclose(extent, [20., 0., 0.]))
 
 
+def test_compare_trees():
```
juanchopanza: Are there any tests for trees that don't compare equal?

neurom/ezy/neuron.py
```diff
@@ -322,7 +323,8 @@ def _compare_neurites(self, other, neurite_type):
 
             return True if len(neurites1) == 0 and len(neurites2) == 0 else \
                    len(neurites1) - sum(1 for neu1, neu2 in
-                                        product(neurites1, neurites2) if neu1 == neu2) == 0
+                                        product(neurites1, neurites2)
+                                        if compare_trees(neu1, neu2)) == 0
```
juanchopanza: I think it would be better to pass the comparison logic as an argument

neurom/core/tree.py
```diff
@@ -253,38 +248,31 @@ def i_chain(trees, iterator_type, mapping=None, tree_filter=None):
     return chain_it if mapping is None else imap_val(mapping, chain_it)
 
 
-def compare_trees(tree1, tree2):
+def make_copy(tree):
     '''
-    Comparison between all the nodes and their respective radii between two trees.
-    Ids are do not have to be identical between the trees, and swapping is allowed
+    Copy a tree structure
```
juanchopanza: Maybe some more explanations about what it means to copy a tree structure?

neurom/core/tests/test_tree.py
```diff
@@ -393,3 +377,23 @@ def _build_tuple(tree):
 
     nt.assert_equal(_build_tuple(REF_TREE2.children[0].children[0].children[0]),
                     (11111, 11112, 11113))
+
+
+def test_make_copy():
```
juanchopanza: We need a test that shows that modifying a copied tree does not modify the original

neurom/core/tree.py
eleftherioszisis: k, tried to make it more clear.

neurom/core/tests/test_tree.py
eleftherioszisis: added

neurom/ezy/neuron.py
eleftherioszisis: something like compare_trees(tree1, tree2, function=check_identical) ?

juanchopanza: I mean in ezy neuron, whoever wants to know if trees are identical should be able to pass a binary predicate (such as the `compare_trees function`.)

eleftherioszisis: Ah, I see. ok

PR #158, opened 2015-11-18
tree transformations

eleftherioszisis:
Created basic tree transformations (rotation, translation) which copy the tree every time they are applied. Rotation and translation are already tested separately. I will have to implement a thorough test for affineTransform when I move it inside neurom.

@lidakanari and everyone please let me know if other functionality is required as well.

@juanchopanza I am unsure where these functions should be moved. I would say morphtree but morphtree is in analysis and this is a little bit more invasive than analysis.


examples/transform.py
```diff
@@ -0,0 +1,199 @@
...
+
+    Returns:
+
+        A copy of the tree with the applied translation.
+    '''
+    # no rotation -> identity matrix
+    return _affineTransform(np.identity(3), t, tree)
+
+
+def rodriguezToRotationMatrix(axis, angle):
```
juanchopanza: Isn't this "Rodrigues" with an "s"?

juanchopanza:
It looks generally OK to me, but I haven't checked all the maths. I think it will need a lot of tests, not only of rotated quantities but that all kinds of invariants have been maintained.

examples/transform.py
eleftherioszisis: Yes. Corrected.

eleftherioszisis:
I agree that it needs extensive testing. So, were should I move the functions? morphtree is the most popular option but I am not sure if I should mix analysis functions with the transformation ones.

juanchopanza:
Maybe in a sub-module `geom`, and call the file `transform.py`?

eleftherioszisis:
I think that geom is too abstract and it does not give away the role of its contents.

I have some suggestions for the submodule name:
* transform, transformation
* modify, modification
 
Is any of the above more appealing than the others? Personally, I prefer modify/modification because it describes clearly what its contents are about and it will permit adding more functions in the future which will modify trees if required.

juanchopanza:
I think `geom` is fine, and the submodule `transform` should make it clear that something is getting modified. We could move other geometrical stuff to `geom` too. I don't like `analysis.morph*` etc.

lidakanari:
The problem is that geom is not saying anything about what the module does and it is counter-intuitive for the user to search for geometry if he wants to modify the cell. 

For the morph part I guess we could have transform instead.

juanchopanza:
But it is a geometrical transformation. To me it doesn't make sense to have a module for things that can change a tree.

eleftherioszisis:
I named it geom and transform.py respectively. 

So how do we differentiate between functions that apply transforms on neurons and on trees?

Also I want to address the issue of numerical precision. It seems that if the rotation origin x,y,z values differ from each other 3-4 orders of magnitude (e.g. (-1., -10., 1000.)) then the test on the affine transform fails due to the fact that the accuracy has dropped to 1e-5 or lower as we increase the distance among the values x.y.z. This can be reproduced if for example in test_affineTransform new_orig is set to (-1. , 2., -600.).

eleftherioszisis:
If instead of the np.dot I do the calculations analytically in _affineTransform, the error reduces significantly and the tests pass for the aforementioned case.

juanchopanza:
"So how do we differentiate between functions that apply transforms on neurons and on trees?"

I'm not sure of the best way to do this in python (I am too used to function overloading). But assuming a transformation function that returns a new object, then one way of dispatching is to use member functions:

    def transform(stuff, transf):
        return stuff.affine_transform(transf)

where `stuff` can be a neuron or a neurite.

eleftherioszisis:
Ok. I like this sim of overloading behavior as it makes the code simpler to use.

eleftherioszisis:
It can be merged now. If more tests are required, let me know.

PR #160, opened 2015-11-20
Remove duplicate points from h5 format morphologies and corresponding tests

lidakanari:
The tests are not yet fully complete to test all the aspects of the loading due to the inconsistency between swc and h5 files. They will be extended in a future commit.


neurom/io/tests/test_h5_reader.py
```diff
@@ -90,31 +90,52 @@ def test_unpack_h2():
...
+
+
+def test_removed_duplicates():
+    v1_data = readers.RawDataWrapper(readers.H5.read_v1(
+            os.path.join(H5V1_PATH, 'Neuron.h5')))
+    v2_data = readers.RawDataWrapper(readers.H5.read_v2(
+            os.path.join(H5V2_PATH, 'Neuron.h5')))
+    for i in v1_data.get_fork_points()[1:]:
+        nt.ok_(not np.allclose(v1_data.get_row(i)[0:4],
+                               v1_data.get_row(v1_data.get_children(i)[0])[0:4]))
```
juanchopanza: Shouldn't this compare with all the children?

neurom/io/hdf5.py
```diff
@@ -81,7 +81,7 @@ class H5(object):
     def read_v1(filename):
```
juanchopanza: I think it would be good to have an option to switch this duplicate removal on, set by default to True

lidakanari: True. Note that as it is it checks if the duplicate point is present and then removes it, but I will change this to be optional.

neurom/io/tests/test_h5_reader.py
lidakanari: Also true. I will add this.

neurom/io/hdf5.py
```diff
@@ -81,7 +81,7 @@ class H5(object):
...
         '''Read an HDF5 v1 file and return a tuple of data, offset, format.'''
         points, groups = _unpack_v1(h5py.File(filename, mode='r'))
-        data = H5.unpack_data(points, groups)
+        data = H5.unpack_data(*H5.remove_duplicate_points(points, groups))
```
juanchopanza: I think we still need to make this optional.

neurom/io/hdf5.py
```diff
@@ -146,6 +146,40 @@ def find_parent_id(point_id):
                           find_parent_id(i))
                          for i, p in enumerate(points)])
 
+    @staticmethod
+    def remove_duplicate_points(points, groups):
+        ''' Removes the duplicate points
+            from the beginning of a section,
+            if they are present in points-groups representation.
+            Returns points, groups with unique points.
```
juanchopanza: Could you make the docstring compatible with the google style so it renders well?

neurom/io/hdf5.py
```diff
@@ -146,6 +146,40 @@ def find_parent_id(point_id):
...
+                                          points[_find_last_point(g[2], groups)]):
+                # Remove duplicate from list of points
+                to_be_removed.append(g[0])
+                # Reduce the id of the following sections
+                # in groups structure by one
+                for igg in range(ig + 1, len(groups)):
+                    to_be_reduced[igg] = to_be_reduced[igg] + 1
+
+        groups = np.array([np.subtract(i, [j, 0, 0])
+                           for i, j in zip(groups, to_be_reduced)])
```
juanchopanza: maybe use `itertools.izip` here to avoid building large tuples.

neurom/io/hdf5.py
lidakanari: YES, this is already optional in the upcoming commit... Just fixing tests before sending everything up.

neurom/io/hdf5.py
lidakanari: Ok, will fix this.

neurom/io/hdf5.py
lidakanari: done!

lidakanari:
Ok, now it is ready for review @juanchopanza. I added many more tests and I hope I improved the documentation accordingly. If I'm still missing some checks please let me know.

juanchopanza:
It all looks good. But I think we should keep the non-duplicate removal tests too, to test that this works the same as before.

lidakanari:
Ok, sure makes sense! So, I will add the previous tests to the testing process and resubmit.

juanchopanza:
Did you remove the duplicate tests? I can't find them, but I'm finding it hard to follow all the changes.

neurom/io/tests/test_h5_reader.py
```diff
@@ -40,6 +40,7 @@
 H5_PATH = os.path.join(DATA_PATH, 'h5')
 H5V1_PATH = os.path.join(H5_PATH, 'v1')
 H5V2_PATH = os.path.join(H5_PATH, 'v2')
+SWC_PATH = os.path.join(DATA_PATH, 'swc')
 
 
 def test_read_h5v1_basic():
```
juanchopanza: I think these test_h5v* are missing the no duplicate removal tests. Other than that it looks fine!

lidakanari: I hope that I understood correctly what you meant. Check my last commit.

juanchopanza:
Perfect, +2 and merging.

PR #162, opened 2015-11-23
Make soma center a property

juanchopanza:


eleftherioszisis:
Apart from Leda's point I think it looks good. And yes it is possible to use make_soma to create the new soma, but I wanted it to be clear and not create an extra copy of the points, transform it and the make the new soma which would require to put building code in the transform.py.

lidakanari:
That's true @juanchopanza so let's keep it as it is and fix this when more information are provided for soma types.

PR #163, opened 2015-11-23
Neuron copy and affine transformations

eleftherioszisis:
Uploading it to run the tests..

### DIFF
--- a/neurom/geom/tests/test_transform.py
+++ b/neurom/geom/tests/test_transform.py
@@ -29,28 +29,28 @@
 import neurom.geom.transform as gtr
 from neurom.core.dataformat import COLS
 from neurom.core.tree import val_iter, ipreorder, Tree, make_copy
-from neurom.core.neuron import Neuron
+from neurom.core.neuron import Neuron, make_soma
 from neurom.ezy import load_neuron
 from nose import tools as nt
 from itertools import izip
 import numpy as np
 from copy import copy
 
-TREE = Tree([0.0, 0.0, 0.0, 1.0, 1, 1, 2] )
-T1 = TREE.add_child(Tree([0.0, 1.0, 0.0, 1.0, 1, 1, 2]))
-T2 = T1.add_child(Tree([0.0, 2.0, 0.0, 1.0, 1, 1, 2]))
-T3 = T2.add_child(Tree([0.0, 4.0, 0.0, 2.0, 1, 1, 2]))
-T4 = T3.add_child(Tree([0.0, 5.0, 0.0, 2.0, 1, 1, 2]))
-T5 = T4.add_child(Tree([2.0, 5.0, 0.0, 1.0, 1, 1, 2]))
-T6 = T4.add_child(Tree([0.0, 5.0, 2.0, 1.0, 1, 1, 2]))
-T7 = T5.add_child(Tree([3.0, 5.0, 0.0, 0.75, 1, 1, 2]))
-T8 = T7.add_child(Tree([4.0, 5.0, 0.0, 0.75, 1, 1, 2]))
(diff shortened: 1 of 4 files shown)


eleftherioszisis:
Should I add a soma copy function as well. Currently, I am using deep copy to copy the soma.

juanchopanza:
After the soma center changes, this might require some merging...

juanchopanza:
I think `soma.copy` would be fine. Then `neuron.copy` could invoke that. But I think it is OK to leave things like this for now.

eleftherioszisis:
I give up. I will debug it tomorrow from my linux machine...

eleftherioszisis:
Just for the record, Daniel found what I did wrong which resulted in the import h5py error. Tests are creating a new virtualenv and I was installing h5py in the nrm one. Thus, by installing it globally the problem was solved.

neurom/geom/tests/test_transform.py
```diff
@@ -28,14 +28,30 @@
...
-from neurom.core.tree import val_iter, ipreorder
+from neurom.core.tree import val_iter, ipreorder, Tree, make_copy
+from neurom.core.neuron import Neuron, make_soma
 from neurom.ezy import load_neuron
 from nose import tools as nt
 from itertools import izip
 import numpy as np
+from copy import copy
 
+TREE = Tree(np.array([0.0, 0.0, 0.0, 1.0, 1, 1, 2]))
```
juanchopanza: Could we have a copy of this test using plain arrays or tuples?

eleftherioszisis: I am not sure that I understand the question. Transformations will fail if sliceable values are present.

juanchopanza: What do you mean by sliceable? lists and tuples are sliceable by pythons definition of that term. numpy. Anyway, there is no requirement that the neuron points be stored as numpy arrays. So if this doesn't work with lists and tuples, we have to re-think it.

eleftherioszisis: I think I only need to do the calculations in affine transform explicitly in order to work with lists and tuples and assign the respective type to the value variable. I think there will be a minor issue with tuple's immutability but I will find a way.

eleftherioszisis: I have a question. Is there an advantage behind having multiple type support for the neurites and neurons? I mean why someone would need to use tuples instead of numpy arrays, for example, for the neurite values? 

I can see why the generalised tree structure can store a variety of types, but the neurites are a specific version of the tree and the transformations are intended for the latter which also need to start with x, y, z. Thus, not every tree can be transformed.

juanchopanza: I didn't see a good reason to limit it to numpy arrays. The data representation is an implementation detail. The only requirements on the neurite tree values are that they be indexable and sliceable, and that the elements correspond to (x, y, z, r).

eleftherioszisis: Can we at least assume mutability as well? Otherwise I will have to perform explicit type checking..

juanchopanza:
I think it is OK to fail if the value is not mutable. The way I would probably have done it is make a big data block from the original tree, transform that in one pass, then build a tree out of it, referring to the block.

eleftherioszisis:
Well, the thing is that I wanted to make it as clean as possible, without merging code for making trees/neurons with transformations. This is why I implemented the copy function which duplicates the objects and let you then do whatever changes on the object copy.

juanchopanza:
Fair enough. So, if the node's value is immutable, the transformation fails, which makes sense.

eleftherioszisis:
@juanchopanza you can merge it if you think it is fine. I do not have anything else to add.

PR #167, opened 2015-11-24
do not try to open config file if not provided.

jdcourcol:
fix attempt to access res variable not assigned if exception is raised in main.

### DIFF
--- a/apps/morph_check
+++ b/apps/morph_check
@@ -196,15 +196,14 @@ if __name__ == '__main__':
 
     config = {}
 
-    try:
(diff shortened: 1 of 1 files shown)


juanchopanza:
That looks fine to me, +2. Maybe @eleftherioszisis can have a second look. Re. testing, we don't test the apps stuff at the moment. The idea is that most of the code should be in tested modules. That is something we need to work on.

eleftherioszisis:
It looks fine to me as well.

PR #169, opened 2015-11-25
fix unraveled h5v2 unpack

jdcourcol:


juanchopanza:
Yes, I like the ternary. The problem is with the comment.

Issue #168, opened 2015-11-25
morph_check fails if no config file is provided.


jdcourcol:
As per the documentation:
```
morph_check some/data/path/morph_file.swc
```
when you do that, morph_check will fail at https://github.com/BlueBrain/NeuroM/blob/master/apps/morph_check#L144

because the config dict does not contain a 'neuron_checks' key.

may be there should be a default config when the file is not provided.

juanchopanza:
Yes, there should have been a documented default. @eleftherioszisis can you look at this?

eleftherioszisis:
Sure, I initialized it to an empty list. Is there a default set of tests that I should put instead?

juanchopanza:
Whatever was being tested before you added the config file would be a reasonable starting point. We can re-evaluate it later.

lidakanari:
Yes, I think a reasonable starting point for the set of default checks is:

Does the cell have a soma?
Does it have basals?
Does it have axons?
Does it have apicals?
Does it contain zero length sections?
Does it contain zero length segments?
Does it contain zero diameter segments? (or points instead?)

eleftherioszisis:
So, if no config is provided, then the above will always be checked. Is that ok with everyone?

juanchopanza:
Sounds fine to me.

PR #170, opened 2015-11-25
init of config in morph_check

eleftherioszisis:


juanchopanza:
Does this fix #168?

eleftherioszisis:
Yes. This basically prevents the test function to break, and by extension throw an exception. This means that it does not do any tests, except from the valid soma which is not included and checked always.

juanchopanza:
OK, good. I will merge it. Apparently, if you'd put "fixes #168" in the commit message, the issue would get closed when this gets merged. See https://help.github.com/articles/closing-issues-via-commit-messages/

eleftherioszisis:
Ah, thanks for the link. I wasn't aware of that.

Issue #130, opened 2015-10-22
Disable h5py loading when instantiating swc morphology ?


wvangeit:
Would it be possible to only load h5py when loading .h5 morphologies ? 
I was trying to load an swc morphology, but I got as error that h5py is not installed. I guess h5py is not necessary in code that only uses swc morphologies ?

h5py is also not set as dependency for pip, so it generated an error after installing NeuroM using pip.

juanchopanza:
Fixed by #171

PR #171, opened 2015-11-25
Make h5py imports fail late.

juanchopanza:
This may sound crazy but there is a rationale behind it: in the current
neurom.io design, SWC and HDF5 readers are imported early. But HDF5 may
be considered an internal BBP format, and most external users may not
need it. Ideally, the IO should use a plug-in architecture to make the
optionality of readers more explicit.

TODO: This isn't correctly tested in the absence of h5py. Figure out how
to test it.


juanchopanza:
@mgeplf @wvangeit this is for issue #130. It is an awful hack, so if you have better ideas on how to do this I am all ears. I would like to make the IO use plugins at some point in the future, but we're not there yet.

wvangeit:
Is this mock H5 class necessary ? Can't you just check if h5py is available during import in hdf5.py, and print a clear error message saying that neurom has to be install with 'pip install neurom[hdf5] to enable this functionality ? 
To prevent hdf5.py being loaded unwillingly reader.py could only execute 'from neurom.io.hdf5 import H5' when somebody attemps to load an h5 file.

juanchopanza:
The mock is so I can keep the structure of the reader unchanged. I will have a look at what you suggest but I am not sure it is worth the changes it would require.

juanchopanza:
@wvangeit I followed your advice in the end and simplified this. It seems to work fine, although it isn't fully tested.

juanchopanza:
Unless I have any objections by 11AM, I will merge this.

mgeplf:
+2 from me

wvangeit:
Looks ok to me

juanchopanza:
Fixes #130

PR #172, opened 2015-11-26
fix unpack for unraveled h5v2 

jdcourcol:


juanchopanza:
Do we have unraveled test files to test this on?

PR #173, opened 2015-11-26
Add default options to morph_checks

lidakanari:
Make default options of checks more clear. Note that this could be replaced by a default yaml file.

### DIFF
--- a/apps/morph_check
+++ b/apps/morph_check
@@ -54,7 +54,7 @@ morphology files.
 Files are unloaded into neuron objects before testing. This means they must
 have a soma and no format errors.
 
-Errors checked for
+Default errors checked for
 ------------------
 * No soma
 * No axon
@@ -63,6 +63,12 @@ Errors checked for
 * Zero radius points
 * Zero length segments
 * Zero length sections
+
+Default options
+--------------
+* nonzero_neurite_radii: 0.007
+* nonzero_segment_lengths: 0.01
+* nonzero_section_lengths: 0.01
 
 Examples
 --------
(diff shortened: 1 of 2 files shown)


apps/morph_check
```diff
@@ -193,7 +193,7 @@ if __name__ == '__main__':
     yaml_path = args.config
     SEPARATOR = '=' * 40
 
-    config = {'neuron_checks': [], 'options': []}
+    config = {'neuron_checks': [], 'options': {}}
```
juanchopanza: Why not simply set the dictionary to the default value in the else block?

apps/morph_check
```diff
@@ -206,12 +206,16 @@ if __name__ == '__main__':
     else:
```
juanchopanza: Here you can say
```python
config = {'neuron_checks': [....], 'options': {....}}
```

apps/morph_check
lidakanari: Is this going to work correctly if you provide a yaml file? I'll check if it works.

eleftherioszisis: I think it can go inside the else statement. I just left it there just in case something unexpected happens. Is there any way to put tests in the apps?

juanchopanza: It should. `if-else` blocks don't have their own scope in python, so anything you declare in them exists from there on.

juanchopanza: The way to test the apps is to put most of the code in (tested) modules, then the app is just a few trivial lines that call a single function and pass the command line arguments to it.

PR #176, opened 2015-11-26
dendrogram class

eleftherioszisis:
I put the dendrogram in analysis and there is the example dendrogram2.py which is the visualization of the dendrogram. The latter will probably migrate in view.


examples/dendrogram2.py
```diff
@@ -0,0 +1,111 @@
...
+
+def _displace(rectangles, t):
+    '''Displace the collection of rectangles
+    '''
+    n, m, _ = rectangles.shape
+
+    for i in xrange(n):
+
+        for j in xrange(m):
+
```
juanchopanza: Shouldn't there be an efficient way of doing this with `numpy`?

neurom/analysis/dendrogram.py
```diff
@@ -0,0 +1,256 @@
...
+import sys
+
+
+def _total_rectangles(tree):
+    '''
+    Calculate the total number of segments that are required
+    for the dendrogram. There is a vertical line for each segment
+    and two horizontal line at each branching point
+    '''
+    return n_segments(tree) + n_bifurcations(tree) * 2
```
juanchopanza: Note that this will ignore forks with more than two branches (not common, but worth keeping in mind)

neurom/analysis/dendrogram.py
```diff
@@ -0,0 +1,256 @@
...
+    Calculate the total number of rectangles with respect to
+    the type of the object
+    '''
+    if isinstance(obj, Tree):
+
+        return _total_rectangles(obj)
+
+    elif isinstance(obj, Neuron):
+
+        return sum([_total_rectangles(neu) for neu in obj.neurites])
```
juanchopanza: You shouldn't need to create a list for the `sum` argument. Just pass the generator expression.

```python
return sum(_total_rectangles(neu) for neu in obj.neurites)
```

neurom/analysis/dendrogram.py
```diff
@@ -0,0 +1,256 @@
...
+        self._dims = []
+
+        # initialize the number of rectangles
+        self._rectangles = np.zeros([_n_rectangles(self._obj), 4, 2])
+
+    def generate(self):
+        '''Generate dendrogram
+        '''
+
+        sys.setrecursionlimit(100000)
```
juanchopanza: We have to find a way of avoiding hitting large recursion depths. In the meantime, you should at lease make sure this function leaves `sys` like it found it, i.e. make sure the limit is set to what it was originally. You could implement this in a decorator to make it more transparent.

neurom/analysis/dendrogram.py
```diff
@@ -0,0 +1,256 @@
...
+
+        spacing = (40., 0.)
```
juanchopanza: magic number?

neurom/analysis/dendrogram.py
eleftherioszisis: My knowledge in recursion is limited but I have the following idea: Given that the recursion is applied on the nodes of the tree, then wouldn't it be possible to predict the recursion depth by calculating how deep a tree goes?

Does this makes sense?

neurom/analysis/dendrogram.py
eleftherioszisis: I am thinking of determining this number with respect to the radii. When it is small the extent of the horizontal lines is small and the radii appear huge which is what Werner wants. Thus, it is in my TODO to find a way around this.

neurom/analysis/dendrogram.py
eleftherioszisis: yes, I will fix this.

examples/dendrogram2.py
eleftherioszisis: Yes, there is. I didn't optimize the example code, because it's there temporarily. I will take care of all these when I will migrate it into view.

neurom/analysis/dendrogram.py
eleftherioszisis: Hmm, this is interesting. You are right, I haven't taken trifuractions into account at all.

neurom/analysis/dendrogram.py
juanchopanza: Probably, but this doesn't scale well. In any case, you should at least make sure to re-set the limit to what is was before this function gets called.

eleftherioszisis:
This is peculiar. All the tests pass on my laptop.

lidakanari:
Try to remove coverage file:

rm .coverage

eleftherioszisis:
Still passes the tests. :/

juanchopanza:
I suggest checking whether there has been some change in `pylint` which could cause these warnings to appear. If there has been, then maybe we need evaluate whether to fix the "problems" or disable the warnings.

juanchopanza:
This is most likely because of `pylint 5.0`, which is a newer version. I am trying to use an older version but ran into other problems.

neurom/analysis/dendrogram.py
```diff
@@ -0,0 +1,295 @@
...
+    by finding the max number of nodes of the given tree or from the biggest neurite in
+    the given neuron.
+    '''
+
+    def depth(tree):
+        ''' Returns the depth which is defined as the number of nodes in a tree
+        '''
+        return sum(1 for _ in ipreorder(tree))
+
+    def set_limit(self):
```
juanchopanza: I think it is confusing to use `self` in a non-member function

neurom/analysis/dendrogram.py
```diff
@@ -0,0 +1,295 @@
...
+        ''' Set the recursion limit, run the function and restore its initial value
+        '''
+        neurites = self.obj.neurites if hasattr(self.obj, 'neurites') else [self.obj]
```
juanchopanza: And this dependency on `self.obj` makes quite a strong coupling between the decorator and the thing being detorated.

neurom/analysis/dendrogram.py
eleftherioszisis: Should I pass only the recursion depth in the decorator?

juanchopanza: I think that would make the decorator more general and useful.

eleftherioszisis: ok!

eleftherioszisis:
I simplified the setting of the recursion limit. I see no point behind having a decorator for the recursion limit at all, thus I just put it inside the function generate. If in the future there will be another function with recursion then I will refactor this part and create a decorator or a context class (with Context: etc). However, because recursion is not preferred in general, I believe that this not very probable.

juanchopanza:
I am merging this now. I think it needs some re-factoring, especially to avoid deep recursions. But that can be done at a later stage.

PR #181, opened 2015-11-30
Restrict pylint version to < 1.5

juanchopanza:
New version 1.5 is causing new problems and seems itself to
have unstable builds. Restrict version to be in the 1.4 series.
Requires explicitly restricting astroid to versions < 1.4.


juanchopanza:
@eleftherioszisis This should fix the build problems you were seeing in #176. I'll merge this now.

PR #183, opened 2015-12-02
Fixing viewer problems: one point tree viewer and documentation of viewers. Fixes #166

### LINKED ISSUE #166: View of tree fails if tree has only one point if diameter = True option is selected
This can be fixed with a try-except statement in the tree viewer.

lidakanari:


juanchopanza:
This fixes #166 right?

PR #186, opened 2015-12-08
NeuroM refactoring proposal example

eleftherioszisis:
This is a proposal for refactoring NeuroM. The motives behind this design are:

* implementanion of a neurite class which makes clear how it differs from a regular tree
* Avoid writing the same feature functions for neurite, neuron and population classes
* avoid polution of the classes' scopes with the addition of more and more features
* and separate the feature implementation from all the other functions in the neurite, neuron, population classes such as iterators, copies etc.
* probably move away from the ezy module

My idea was to instantiate a NeuriteFeatures object in the neurite class, through which one can call the feature functions. Then, the neuron object instantiates a NeuronFeatures object which inherits the functions from NeuriteFeatures in order share the functions that are shared from the two objects. The population in a similar manner creates the PopulationFeatures object which currently is a placeholder in case that there are features that are sensible only on a population level.

In order for the feature classes to handle multiple objects, some duck typing has been considered in order to accept always iterables, even if rhis is only one object.

All the feature functions return a flattened list which correspond to the calculated value on each neurite. For example:

pop.features.section_lengths()

will return a list with all the section lengths in all neurites in the neurons of the populations without nesting, ready to use for fitting etc. If per neurite features are required, this does not need to be implemented as an explicit function, because the users will be able ro do:

sec_lenths_per_neurite = [neuron.features.section_lengths() for neuron in pop.neurons]

which will provide the same results but now nested per neuron. This is the same for neurites as well.

Of course if there is need for these functions to exist, they can always be implemented.

Please let me know what do you think about this example. I think it would be nice if we could discuss, refine and change it in order to have a nice and useful design in NeuroM.


juanchopanza:
I am not sure the problems this is trying to fix are the problems we should be fixing. This looks more in the direction of the `ezy` module, which I think was a bad idea in the first place. Regarding some of the "problems" it solves:

* Avoid writing the same feature functions for neurite, neuron and population classes
* avoid polution of the classes' scopes with the addition of more and more features

We don't need to do either of these. We don't need to write functions for every single feature, and there is no reason we should be writing them twice anyway. I have said many times that it doesn't scale, which is why I think the `ezy` module is a bad idea and I would like to remove it one day. What I think needs to be done is to simplify the iteration mechanisms such that a user can simply say (ignoring filtering for simplicity)

```python
stuff = [my_function(x) for x in neuron.neurites]
stuff = [my_function(x) for x in population.neurites]
```

Of course, this can be put in a helper function. The point is that there is no difference between a neuron and a population when it comes to neurites.

What is missing here is a means to bind functions to iteration modes. I have to think about how to do this nicely. Of course, the user could still say something like

```python
stuff = [my_segment_fun(s) for s in neuron.segments]
stuff = [my_segment_fun(s) for s in population.segments]
```
but it would be nice to have a mechanism by which they don't have to do that, at least for NeuroM provided functions.

eleftherioszisis:
Thanks @juanchopanza  for the reply.

Yes, I see your point of exposing the elements of each abstraction level and I admit that it is simpler than my implementation. My sole objection on that is that the user should be aware on which objects (trees, segments, sections, neurons, etc.) he/she should apply the analysis functions, and this is by no means obvious. (is this what you mean by binding functions to iteration modes?)

If we can substitute the feature inheritance schema with e.g. a clear namespace separation of different types of helper analysis functions then I think this would be equivalent to my complicated approach.

I still however believe that Neurite must be either a different object than Tree or inherit from it, where its type is assigned in a clear way.

I would like to play a little bit with the example after your reply, so do not merge it yet (I might even remove it completely in the end :P).

juanchopanza:
> My sole objection on that is that the user should be aware on which objects (trees, segments, sections, neurons, etc.) he/she should apply the analysis functions, and this is by no means obvious. (is this what you mean by binding functions to iteration modes?)

Yes, that's it. If a user is using a `NeuroM` function, say `segment_colour`, then they shouldn't have to ask for the segment iterator. They would just say `segment_colour(c) for c in blah]` or something like that. I have to think of the best way to do this. But this is likely to require that "the system" know about `segment_colour`. So if a user writes their own `NeuroM`-independent function, they would have to ask for the right kind of iterator (unless they "register" it as a segment function somehow.)

>I still however believe that Neurite must be either a different object than Tree or inherit from it, where its type is assigned in a clear way.

Yes, that might be a good idea.

eleftherioszisis:
I changed the structure. Is this close to what you suggested @juanchopanza? The chaining of iterators is still not very pretty.

Also I was not sure if I should put the same function _i_neurites inside the population Class (unless we make it a helper function), thus I called the _i_neurites  (which i simplified a little for the sake of the example :P) from within the Neuron class.

I have a question: Is it efficient to use the asterisk to pass the results as lists of arguments for the chain function etc?

examples/refactoring.py
```diff
@@ -0,0 +1,168 @@
...
+                                tree_filter=lambda t: checkTreeType(neurite_type, t.type))
+
+    def sections(self, neurite_type=TreeType.all):
+        '''Returns sections
+        '''
+        return self._i_neurites(lambda n: n.sections(),
+                                tree_filter=lambda t: checkTreeType(neurite_type, t.type))
+
+
+class Population(object):
```
juanchopanza: I still don't understand why the population has to have different code to the neuron when it comes to getting any neurite related information. Both can store collections of neurites, so both could use exactly the same code.

eleftherioszisis: Yes, indeed. However, in order to do that I need to make _i_neurites a helper function and remove it from the Neuron class, no?

juanchopanza: I am not thinking of any particular solution. Just that

```python
def do_stuff(thing_with_neurites):
    ....

pop = Population(...)
nrn = Neuron(...)
do_stuff(pop)
do_stuff(nrn)
```
should work and be trivial because Population and Neuron look the same when it comes to neurites (or at least, they should, and I think they did when both had `neurites` lists.

eleftherioszisis: sure, there is no reason behind putting neurites as a property!

eleftherioszisis:
Ok, I put i_neurites outside Neuron class. I still need to replcate segments and sections code twice.

juanchopanza:
I can merge this, but I am not sure we will keep any of it in `neurom`. We may have to remove it or heavily modify it in the future to avoid confusion.

eleftherioszisis:
I agree. We can discuss it after hackathon.

PR #188, opened 2015-12-14
Refactor some morpology checks.

juanchopanza:
Modified neurite monotonicity and flatness checks woth the following
considerations:

* Tests with names such as has_x return True or False.
* Where possible, write has_x tests for which True is good, False is
  bad.
* Provide getter functions to get lists of "bad" neurites

### DIFF
--- a/apps/config.yaml
+++ b/apps/config.yaml
@@ -48,8 +48,8 @@ neuron_checks :
     - nonzero_neurite_radii
     - nonzero_segment_lengths
     - nonzero_section_lengths
-    - has_monotonic_neurites
-    - has_flat_neurites
+    - has_all_monotonic_neurites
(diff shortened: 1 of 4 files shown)


juanchopanza:
@jdcourcol @eleftherioszisis @lidakanari This might be of interest. I will add a test for number of flat neurites in a future commit. That will allow for different configurations to test for different flatness criteria.

juanchopanza:
Any comments / objections?

lidakanari:
It seems ok to me, but I would wait for Eleftherios to check the details.

eleftherioszisis:
I think it is good. Now it is much clearer to the user, what the functions do!

juanchopanza:
OK, thanks. I will merge this.

PR #189, opened 2015-12-15
Add helper neurom.check.ok

juanchopanza:
Returns sensible values of True or False for morphology checks:

True / False ==> True / False
Iterable ==> True if empty else False

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
 install:
   - pip install --upgrade pip
   - pip install --ignore-installed -r requirements_dev.txt
-  - pip install matplotlib==1.3.1
+  - pip install matplotlib==1.4.1
   - pip install codecov
   - pip install .
 cache:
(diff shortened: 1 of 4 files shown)


juanchopanza:
@jdcourcol This should simplify the checks.

jdcourcol:
agree. thank you.

juanchopanza:
@eleftherioszisis @lidakanari FYI some more simplifications to morphology checking.

.travis.yml
```diff
@@ -11,7 +11,7 @@ before_install:
 install:
   - pip install --upgrade pip
   - pip install --ignore-installed -r requirements_dev.txt
-  - pip install matplotlib==1.3.1
+  - pip install matplotlib==1.4.1
```
lidakanari: Out of curiosity, why do we modify the matplotlib version?

lidakanari:
I like the structure with the tests for this :+1:

.travis.yml
juanchopanza: Travis moved their infrastructure, and the matplotlib build started failing. There was an issue with it not working with the installed numpy, which had been fixed in newer versions. I don't know how or why it worked before.

lidakanari: Ok, we need to check if all the plotting functionality is compatible with this version, but if no tests are failing that's a good start. Good to know.

PR #192, opened 2015-12-18
Remove non-consecutive IDs restriction for SWC data.

juanchopanza:
This is overly restrictive, and isn't specified in

* http://neuromorpho.org/myfaq.jsp
* http://www.sciencedirect.com/science/article/pii/S0165027098000910


juanchopanza:
@lidakanari @a1eko This should fix the non-consecutive IDs issue.

juanchopanza:
I replace the consecutive IDs check for one that checks that the ID sequence is increasing. I think this fits the neuromorpho.org "specification" better.

lidakanari:
I think this is ok!

PR #195, opened 2016-01-06
fix: morphmath.angle_3points could not cope with collinear vectors

juanchopanza:
Add new tests for pi radians and zero angles and change the
implementation of angle calculation to a more robust one
using atan2 instead of acos..

### DIFF
--- a/neurom/analysis/morphmath.py
+++ b/neurom/analysis/morphmath.py
@@ -83,12 +83,11 @@ def angle_3points(p0, p1, p2):
 
     Returns:
         Angle in radians between (p1-p0) and (p2-p0).
-        nan if p0==p1 or p0==p2.
+        0.0 if p0==p1 or p0==p2.
     '''
     vec1 = vector(p1, p0)
     vec2 = vector(p2, p0)
-    return math.acos(np.dot(vec1, vec2) /
-                     (np.linalg.norm(vec1) * np.linalg.norm(vec2)))
+    return math.atan2(np.linalg.norm(np.cross(vec1, vec2)), np.dot(vec1, vec2))
 
 
 def polygon_diameter(points):
--- a/neurom/analysis/tests/test_morphmath.py
+++ b/neurom/analysis/tests/test_morphmath.py
@@ -78,16 +78,16 @@ def test_angle_3points_half_pi():
     orig = (0.0, 0.0, 0.0)
     vec1 = (1.0, 0.0, 0.0)
     vec2 = (0.0, 2.0, 0.0)
-    nt.ok_(angle_3points(orig, vec1, vec2)==pi/2.0)
+    nt.eq_(angle_3points(orig, vec1, vec2), pi / 2.0)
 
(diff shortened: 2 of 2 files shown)


neurom/analysis/morphmath.py
```diff
@@ -83,12 +83,11 @@ def angle_3points(p0, p1, p2):
...
     Returns:
         Angle in radians between (p1-p0) and (p2-p0).
-        nan if p0==p1 or p0==p2.
+        0.0 if p0==p1 or p0==p2.
     '''
     vec1 = vector(p1, p0)
     vec2 = vector(p2, p0)
-    return math.acos(np.dot(vec1, vec2) /
-                     (np.linalg.norm(vec1) * np.linalg.norm(vec2)))
+    return math.atan2(np.linalg.norm(np.cross(vec1, vec2)), np.dot(vec1, vec2))
```
mgeplf: huh,  math.atan2(0., 0.) == 0.0
weird

neurom/analysis/tests/test_morphmath.py
```diff
@@ -74,21 +74,82 @@ def test_point_dist():
...
-def test_angle_3points():
+def test_angle_3points_half_pi():
+    orig = (0.0, 0.0, 0.0)
     vec1 = (1.0, 0.0, 0.0)
-    vec2 = (0.0, 1.0, 0.0)
-    orig = (0.0,0.0,0.0)
-    angle=angle_3points(orig, vec1, vec2)
-    nt.ok_(angle==pi/2.0)
+    vec2 = (0.0, 2.0, 0.0)
+    nt.ok_(angle_3points(orig, vec1, vec2)==pi/2.0)
```
mgeplf: use nt.eq_ so it will give you the numbers if/when it fails?

neurom/analysis/tests/test_morphmath.py
```diff
@@ -74,21 +74,82 @@ def test_point_dist():
...
+
+    vec1 = (0.0, 4.0, 0.0)
+    nt.ok_(angle_3points(orig, vec1, vec2)==pi/2.0)
+
+
+def test_angle_3points_quarter_pi():
+    orig = (0.0, 0.0, 0.0)
+    vec1 = (1.0, 0.0, 0.0)
+    vec2 = (2.0, 2.0, 0.0)
+    nt.assert_equal(angle_3points(orig, vec1, vec2), pi / 4.0)
```
mgeplf: I also prefer eq_ to assert_equal, as that actually comes from unittest via a weird set of imports, fyi

mgeplf:
+2 from me

neurom/analysis/tests/test_morphmath.py
juanchopanza: Will do. I've been using `nt.assert_equal` for that.

neurom/analysis/tests/test_morphmath.py
juanchopanza: Any benefits in doing that?

PR #197, opened 2016-01-08
fix: non-connected points are a fatal error

juanchopanza:
neuron loading was ignoring disconnected points, simply
truncating trees at the point of disconnection. This is
now an error that makes the loading fail.


juanchopanza:
@lidakanari disconnected stuff wasn't an error, it is now!

lidakanari:
I think this is good!

PR #198, opened 2016-01-11
New structure for neurite segment morphometrics calculators.

juanchopanza:
General idea: put functions and their compatible iterators
under the same module to add some kind of semantic association
between them. Iteration source is a neuron or population, not
a single tree.

Initial test case:
* module neurom.segments with functors and a segment iterator.

Examples:

    import neurom.segments as seg
    nrn = ... # core or ezy neuron, or population
    # simple case
    seg_len = [l for l in seg.itr(nrn, seg.length)]
    # less simple: radial_dist requires an origin WRT which the distance is calculated
    seg_rad_dist = [d for d in seg.itr(nrn, seg.radial_dist(nrn.soma.center()))]

### DIFF
--- a/neurom/segments.py
+++ b/neurom/segments.py
@@ -38,13 +38,14 @@ iter_type = tr.isegment
 
 
 def itr(neuron, mapping=None, filt=None):
-    '''Iterator to neuron's segments
+    '''Iterator to a neuron or neuron population's segments
 
-    Applies a filter function and a mapping.
+    Applies a neurite filter function and a segment mapping.
 
     Example:
         Get the lengths of segments in a neuron and a population
 
+        >>> from neurom import segments as seg
         >>> neuron_lengths = [l for l in seg.itr(nrn, seg.length)]
         >>> population_lengths = [l for l in seg.itr(pop, seg.length)]
 
--- a/neurom/segments.py
+++ b/neurom/segments.py
@@ -37,8 +37,8 @@ import neurom.analysis.morphmath as mm
 iter_type = tr.isegment
 
 
-def itr(neuron, mapping=None, filt=None):
-    '''Iterator to a neuron or neuron population's segments
+def itr(obj, mapping=None, filt=None):
+    '''Iterator to a neurite, neuron or neuron population's segments
 
     Applies a neurite filter function and a segment mapping.
 
@@ -48,9 +48,12 @@ def itr(neuron, mapping=None, filt=None):
         >>> from neurom import segments as seg
         >>> neuron_lengths = [l for l in seg.itr(nrn, seg.length)]
         >>> population_lengths = [l for l in seg.itr(pop, seg.length)]
-
+        >>> neurite = nrn.neurites[0]
+        >>> tree_lengths = [l for l in seg.itr(neurite, seg.length)]
     '''
-    return tr.i_chain(neuron.neurites, iter_type, mapping, filt)
+    #  TODO: optimize case of single neurite and move code to neurom.core.tree
+    neurites = [obj] if isinstance(obj, tr.Tree) else obj.neurites
+    return tr.i_chain(neurites, iter_type, mapping, filt)
 
 
 length = mm.segment_length
--- a/neurom/segments.py
+++ b/neurom/segments.py
@@ -1,0 +1,73 @@
+# Copyright (c) 2015, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
+# All rights reserved.
+#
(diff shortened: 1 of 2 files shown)


eleftherioszisis:
Is it necessary for the iterators to be exposed to the user?

juanchopanza:
@eleftherioszisis I am not sure what you mean. Generator expressions require an iterator, so at some level there must be something that produces one. But the user now doesn't need to know much about the particular type of iterator. Only that if they want to use function `seg.foo`, they need `seg.itr`.

eleftherioszisis:
Yes, I agree that generators are essential but I am also aware that iterator logic is not easily understandable by non-programmers. Thus, the question was mainly on the interfacing with the user and this is why I presented some more complicated designs where the user does not have to think with iterators in order to extract his favorite feature.

juanchopanza:
I have some other ideas that I may try out later, but this is the simplest one and doesn't require any changes. It is basically re-packaging and it solves the neuron/population problem for most cases. However, I can't really say what will confuse non-programmers so I don't want to drive the design based on that. Intuitively, I would expect a non-programmer with a mathematical background to understand functional-style python more than, say, OO python.

eleftherioszisis:
Well, I am not sure how biologists and especially the experimentalists fit in that requires a mathematical background :P 

As far as I know from people of the lab that tried to use NeuroM, iterators were not intuitive and by extension hard to use. This is the main reason that some prefer to stick with Pneumatk, even though it is not tested...

Unfortunately, whether we like it or not, we have to take into account the difficulty of the software for the groups that we are targeting, because NeuroM is not destined only for interior use. And to be honest I think that the user should not go around searching for the correct function of the correct object to apply the calculation of choice.. :/ (I like the semantic association though for it links the functions with their respective objects!)

juanchopanza:
In my mind these types of users will be running scripts to get pre-packaged sets of data, in which case they don't really need to understand anything other than how to pass commandline arguments or modify configuration files. How these map to `neurom` code is not for the non-programmers. And it should be kept as simple as possible, which is not the case right now.

juanchopanza:
I have another scheme where the function knows the type of iterator it requires. This means there is only a single general "iterator" function, so the expression would become something like

    seg_len = [l for l in nrn.itr(seg.length)]
    sec_vol = [v for v in nrn.itr(sec.volume)]

This in principle means the user doesn't have to worry about calling the right `itr` function.

eleftherioszisis:
Is it possible to hide the iterator entirely? i.e.:
```python
seg_len = [l for l in seg.length(nrn)]
```
How about this "slightly" extreme example of a factory using a functor objects:

```python
class Neuron(object):
    def __init__(self, ...):
        ...
        self._primitives = Primitives(self.neurites)
    @property
    def segments(self):
        return self._primitives('segments')
    @property
    def sections(self):
        return self._primitives('sections')
```
And then the Elements class that dispatches each module:

```python
# functors as well, i.e. they implement the operator ()
from neurom.primitives import Segments, Sections  

class_dict = {'segments': Segments, 'sections': Sections}
class Elements(object):
    def __init__(self, objects):
        self._objects = objects

    def operator()(self, key):
        return class_dict['key'](self._objects)
```

Which you can use:

```
nrn.segments.my_favorite_segment_function(args)
nrn.sections.my_favorite_section_function(args)
```
and the same logic would be with the population and a single neurite. Isn't it pretty (provided it works)? :P

juanchopanza:
I was actually thinking of a variant that would allow something like what you said here:

    seg_len = [l for l in seg.length(nrn)]

I may have more on that tomorrow.
As for this

    nrn.segments.my_favorite_segment_function(args)

I it doesn't scale well, which is why I try to keep the functions decoupled from the neuron. This allows the user to define and use their own segment function without having to do anything special (except maybe decorate their function.)

eleftherioszisis:
Cool, I think that approach is informative enough for the user. It would also be nice if it can work with a single neurite as well.

neurom/segments.py
```diff
@@ -0,0 +1,73 @@
...
+
+def itr(neuron, mapping=None, filt=None):
+    '''Iterator to neuron's segments
+
+    Applies a filter function and a mapping.
+
+    Example:
+        Get the lengths of segments in a neuron and a population
+
+        >>> neuron_lengths = [l for l in seg.itr(nrn, seg.length)]
```
mgeplf: >> import neurom.segments as seg
?

juanchopanza: I'll add that.

mgeplf:
Looks good to me

PR #199, opened 2016-01-11
Table of plots for feature and cells

eleftherioszisis:
This example  covers for:

A) A table where each row is a feature and the columns are graphs of that feature for the cell created by different reconstruction algorithms (this compares many features from many different reconstruction algorithms for the same cell)

B) Superimposition of a selected feature graph for a set of reconstructed cells  (this could be either one cell reconstructed by many different algorithms, or different instances of the same algorithm on different image stacks). 

Please let me now if power ups are needed!


juanchopanza:
Looks fine to me but I don't know much matplotlib. @lidakanari can you have a look?

eleftherioszisis:
I will need to do some changes tomorrow, so do not merge it yet.

lidakanari:
Ok, so we need to add a couple of more features-capabilities but it is good in general. We can work on the details later!

eleftherioszisis:
@juanchopanza I will remove the main and enable the user to call it without editing the file, so do not merge yet. (I have to meet with Giuseppe so it might take some time)

eleftherioszisis:
Ok, I added the argparse stuff. @lidakanari  is it sufficient or should I add sth else as well?

lidakanari:
Looks good to me! @juanchopanza you can merge it and we correct details in the future versions according to our needs.

PR #200, opened 2016-01-12
Add iteration helpers and functions for section morphometrics.

juanchopanza:
* Minor fixes to segment morphometrics code.

### DIFF
--- a/neurom/core/tree.py
+++ b/neurom/core/tree.py
@@ -248,6 +248,25 @@ def i_chain(trees, iterator_type, mapping=None, tree_filter=None):
     return chain_it if mapping is None else imap_val(mapping, chain_it)
 
 
+def i_chain2(trees, iterator_type, mapping=None, tree_filter=None):
+    '''Returns a mapped iterator to a collection of trees
+
+    Provides access to all the elements of all the trees
+    in one iteration sequence.
+
+    Parameters:
+        trees: iterator or iterable of tree objects
+        iterator_type: type of the iteration (segment, section, triplet...)
+        mapping: optional function to apply to the iterator's target.
+        tree_filter: optional top level filter on properties of tree objects.
+    '''
+    nrt = (trees if tree_filter is None
+           else filter(tree_filter, trees))
+
+    chain_it = chain(*imap(iterator_type, nrt))
+    return chain_it if mapping is None else imap(mapping, chain_it)
+
+
 def make_copy(tree):
     '''
     Copies a tree structure. A new tree is generated with the copied values
--- a/neurom/core/tree.py
+++ b/neurom/core/tree.py
@@ -178,10 +178,10 @@ def itriplet(tree):
 
     Post-order iteration yielding tuples with three consecutive sub-trees
     '''
-    return chain(
-        *imap(lambda n: zip(repeat(n.parent), repeat(n), n.children),
-              ifilter(lambda n: not is_root(n) and not is_leaf(n),
-                      ipreorder(tree))))
+    return chain.from_iterable(
+        imap(lambda n: zip(repeat(n.parent), repeat(n), n.children),
+             ifilter(lambda n: not is_root(n) and not is_leaf(n),
+                     ipreorder(tree))))
 
 
 def i_branch_end_points(fork_point):
@@ -244,7 +244,7 @@ def i_chain(trees, iterator_type, mapping=None, tree_filter=None):
     nrt = (trees if tree_filter is None
            else filter(tree_filter, trees))
 
-    chain_it = chain(*imap(iterator_type, nrt))
+    chain_it = chain.from_iterable(imap(iterator_type, nrt))
     return chain_it if mapping is None else imap_val(mapping, chain_it)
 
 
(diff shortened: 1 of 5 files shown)


neurom/sections.py
```diff
@@ -0,0 +1,98 @@
...
+def radial_dist(pos, use_start_point=False):
+    '''Return a function that calculates radial distance for a section
+
+    Radial distance calculater WRT pos
+
+    TODO: Can we simplify this to use the middle point?
+    '''
+    sec_idx = 0 if use_start_point else -1
+
+    def _dist(section):
```
juanchopanza: If somebody knows how to do this with `functools` it would be nice to know.

neurom/sections.py
```diff
@@ -0,0 +1,98 @@
...
+    '''Calculate the total volume of a segment'''
+    return _aggregate_segments(section, mm.segment_volume)
+
+
+def area(section):
+    '''Calculate the surface area of a section'''
+    return _aggregate_segments(section, mm.segment_area)
+
+
+def radial_dist(pos, use_start_point=False):
```
juanchopanza: I wonder if we need a naming convention to express that this is a function that returns a function.

eleftherioszisis: A decorator would not work?

neurom/sections.py
mgeplf: not sure there's a way, you'd need a way to evaluate section[sec_idx] at call time, since it's __getitem__ could be anything, so you have to wrap it as far as I know.  operator.itemgetter doesn't help either, in this example.

eleftherioszisis:
Is it possible to interface:
```
rd = [d for d in seg.itr(n, seg.radial_dist(origin))
```
in a way that it will be like:
```
rd = [d for d in seg.radial_dist(n, origin)]
```
and thus hide the iterator and function argument from the user?

juanchopanza:
That would mean that every time a user wants to implement a new morphometrics segment function, they have to write an iterator, which leads us to the problem we have now. It isn't general enough. Which the scheme I propose, the iteration is less coupled to the function, making it easier to extend.

juanchopanza:
I think the best that can be hoped for is to do something like

    @segfun
    def foo(segment):
        return segment[-1] - segment[0]

then

    seg_foos = [f for f in neurom.magic_iter(n, foo)]

eleftherioszisis:
Maybe it is easier for us to use and extend but for the regular python user it will be hard to do both. And hard usually means "switch to another software that is user friendly".

I had to check the code and the tests in order to write the code correctly for simple functions as area and ones that need arguments as the radial_dist. I repeat that this software targets scientists that might not have programmed again in their life an probably need the analysis for their work, or maybe to implement some simple stuff and statistics.

This is not c or c++ where the users are usually experienced programmers. Sorry for insisting so much on this but I really want NeuroM to completely replace Pneumatk which is preferred by many because it is very simple to use and extend.

juanchopanza:
One of the reasons it is so complicated to write tests and such is because we were doing exactly the kind of thing you are suggesting. I cannot second-guess what this so called archetypal python end user might find confusing. And I am not going to repeat the mistakes we already made. If you have some simple, realistic "scientist" use-cases, go ahead and add them as an issues. Especially if it is something that would be "easily" done in pneumatk.

neurom/core/tree.py
```diff
@@ -248,6 +248,25 @@ def i_chain(trees, iterator_type, mapping=None, tree_filter=None):
...
+    Parameters:
+        trees: iterator or iterable of tree objects
+        iterator_type: type of the iteration (segment, section, triplet...)
+        mapping: optional function to apply to the iterator's target.
+        tree_filter: optional top level filter on properties of tree objects.
+    '''
+    nrt = (trees if tree_filter is None
+           else filter(tree_filter, trees))
+
+    chain_it = chain(*imap(iterator_type, nrt))
```
mgeplf: consider using chain.from_iterable:
https://docs.python.org/2/library/itertools.html#itertools.chain.from_iterable

neurom/sections.py
```diff
@@ -0,0 +1,151 @@
...
+
+#  TODO: If this proves useful, more it to neurom.core.tree
+def to_val(fun):
+    '''Decorate tree value function to accept trees'''
+    def _deep_map(data):
+        '''Recursive Tree -> Tree.data transformation function.
+
+        Maintains type of iterables
+        '''
+        return (type(data)(_deep_map(x) for x in data)
```
mgeplf: hmm, this will probably work for most 'data' types, but type(data)(...) might not work for more complicated types.  IE if data is an np.ndarray

neurom/core/tree.py
juanchopanza: Done (in this file only. I will do it elsewhere in a separate pull request.)

neurom/sections.py
juanchopanza: This hasn't been a problem yet. A similar thing is done for most "value" iterations. I think the function arguments are usually tuples of trees, because that is what the different types of iterations return. The problem I see is that once decorated, the function has specific requirements on its argument, which is why I was asking about getting an un-decorated version.

PR #201, opened 2016-01-13
morph_stats now can treat multiple directories as populations of neurons

eleftherioszisis:
Multiple directories can be provided as args in morph_stats. if --dir_as_pop is enabled then each directory is treated as a population and the stats are extracted population-wise.

Otherwise, a list with all the filenames in all directories is created and the stats that are extracted are neuron-wise.

### DIFF
--- a/apps/morph_stats
+++ b/apps/morph_stats
@@ -73,7 +73,7 @@ def parse_args():
 
     parser.add_argument('datapaths',
                         nargs='+',
-                        help='Paths to morphology data file or directory')
+                        help='Paths to morphology data files or directories')
 
     parser.add_argument('-v', '--verbose', action='count', dest='verbose', default=0,
                         help='-v for INFO, -vv for DEBUG')
@@ -82,8 +82,8 @@ def parse_args():
                         action='store_true',
                         default=False,
                         help='If more than one directories are specified in the \
-                              datapaths and this flag is enabled the cells in the \
-                              will be treated a population')
+                              datapaths and this flag is enabled the cells in each dir\
+                              will be treated as a population')
     return parser.parse_args()
 
 
--- a/apps/morph_stats
+++ b/apps/morph_stats
@@ -87,12 +87,20 @@ def extract_stats(files):
                 value = getattr(nrn, ns)(n)
                 L.debug('Stat: %s, Neurite: %s, Type: %s', ns, n, type(value))
                 if isinstance(value, np.ndarray):
-                    value = {
-                        'mean': np.mean(value),
-                        'std': np.std(value),
-                        'min': np.min(value),
-                        'max': np.max(value),
-                    }
+                    if len(value) > 0:
+                        value = {
+                            'mean': np.mean(value),
+                            'std': np.std(value),
+                            'min': np.min(value),
+                            'max': np.max(value),
+                        }
+                    else:
+                        value = {
+                            'mean': 0.,
+                            'std': 0.,
+                            'min': 0.,
+                            'max': 0.,
+                        }                        
                 stats[stat_name][n.name] = value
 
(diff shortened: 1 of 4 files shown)


juanchopanza:
What happens if we want to treat all the filenames in all the directories as a single population? I take it there is a difference between population and collection of neurons, or do these boil down to the same at the moment?

juanchopanza:
The stuff in apps is not tested, so I think we should have **less** code in there. For example, all the stats calculating stuff should go in a module and be tested.

eleftherioszisis:
Hmm, there is no option right now that all the files in all directories can be treated as a single population. I thought that the user can put a population inside a directory. For example, if they have a population of cells from a reconstruction, probably it will be inside a folder. If they have multiple reconstructions, i.e. multiple folders then they will be able either parse all their cells for analysis in a single cell basis or treat each directory as a single collection/population and by extension extract morphometrics with respect to each reconstruction.

I agree for the testing of the stats.

juanchopanza:
Maybe the population / collection of neurons - one directory / many directories problem can be split into more orthogonal options

* `as-population` (else treat as collection of neurons)
* `single-group` (all directories treated as one)

OTOH, if populations and single neurons are different, population vs. collection of neurons should not be mutually exclusive. The user should be able to pick one of them or both (and stats that are equivalent between the two should not be duplicated.)

juanchopanza:
Bear in mind that users might want to split an analysis into many jobs. So we could have many jobs running population stats on single directories, and then the data being aggregated over jobs. So it makes sense that a single directory is just a sub-set of many directories, and not something totally different.

eleftherioszisis:
I made the code more compact and now if the as-population flag is provided then the directories are treated as populations but if single files are provided as well then they are processes as single neurons. Thus in the resulting json populations and single neurons can coexist for comparison etc.

@juanchopanza can you please take a look on the layout of the code and give me some directions as to how can I achieve the data aggregation over multiple jobs? (In other words I am not sure that I understood your previous comment, thus I need more info)

juanchopanza:
I think if N files can be input into the application, whether they are treated as a population or as a collection of single neurons should depend on the flag, not on whether the files are in a directory. So I should be able to pass a list of files and have them treated as a population. Otherwise it is too confusing. Currently I think you can pass many files, set the `--as-population` flag, and the neurons don't get treated as a population.

As for data aggregation, it really depends on the type of stats being collected. There isn't a general way to aggregate, but let's assume for now that the users know how to deal with multiple output JSON files. In the future, we may have to have different output for parallelized jobs. But for now I think it is enough to let the user make a population out of the stuff in a single directory (and a list of input files, as mentioned in the previous paragraph.)

eleftherioszisis:
Ok now the app takes a single path as input and depending on the flag --as-population treats its contents (if a dir) separately or as population if it's false or true respectively.

This is the simplest form of the file and I think it's easy to run multiple jobs for different directories. One thing that we might need to think is how we separate in multiple jobs thousands of files that reside in one directory.

juanchopanza:
In that case, this comment has to be fixed:

    help='Paths to morphology data files or directories'

Also, better check with @lidakanari whether we want to lose the ability to deal with more than one population in a single job. It isn't clear to me that we want to stop doing that (assuming it worked before.)

eleftherioszisis:
Oops, yep you're right

lidakanari:
Is this ok, because I will need to work on the same files?

eleftherioszisis:
From my part it is :P

juanchopanza:
@lidakanari It is OK if you think it is OK that we can't analyse more then one population with this script.

lidakanari:
No problem as it is. We can easily generalize this for more directories if we need to (in a future commit).

PR #207, opened 2016-01-15
Move generic iteration function to top level module

juanchopanza:
Function is called neurom.iter_neurites. Code in neurom.core.
Functionality extended to segment functions.

### DIFF
--- a/neurom/core/__init__.py
+++ b/neurom/core/__init__.py
@@ -31,6 +31,7 @@
 from .tree import i_chain2 as _chain_neurites
 from .tree import Tree as _Tree
 
+
 def iter_neurites(obj, mapfun=None, filt=None):
     '''Iterator to a neurite, neuron or neuron population
 
--- a/neurom/core/__init__.py
+++ b/neurom/core/__init__.py
@@ -27,3 +27,25 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 ''' Core functionality and data types of NeuroM '''
+
+from .tree import i_chain2 as _chain_neurites
+from .tree import Tree as _Tree
+
+def iter_neurites(obj, mapfun=None, filt=None):
+    '''Iterator to a neurite, neuron or neuron population
+
+    Applies a neurite filter function and a mapping function.
+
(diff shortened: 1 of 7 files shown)


neurom/core/__init__.py
```diff
@@ -27,3 +27,26 @@
...
+        Get the lengths of sections in a neuron and a population
+
+        >>> from neurom import sections as sec
+        >>> neuron_lengths = [l for l in iter_neurites(nrn, sec.length)]
+        >>> population_lengths = [l for l in iter_neurites(pop, sec.length)]
+        >>> neurite = nrn.neurites[0]
+        >>> tree_lengths = [l for l in iter_neurites(neurite, sec.length)]
+
+    '''
+    #  TODO: optimize case of single neurite and move code to neurom.core.tree
```
mgeplf: this todo still relevant?

neurom/tests/test_segments.py
```diff
@@ -195,3 +204,8 @@ def test_count():
 def test_segment_radial_dists():
     _check_segment_radial_dists(SIMPLE_NEURON)
     _check_segment_radial_dists(SIMPLE_TREE)
+
+
+def test_segment_taper_rate():
```
mgeplf: taper wasn't added in this diff, right?

neurom/segments.py
```diff
@@ -75,11 +78,16 @@ def radial_dist(pos):
 
     Radial distance calculater WRT pos
     '''
-    return functools.partial(mm.segment_radial_dist, pos=pos)
+    @segment_function(as_tree=False)
```
mgeplf: I probably would use functools.partial to bind pos

mgeplf:
seems reasonable to me

neurom/segments.py
juanchopanza: I tried, and it clashed with my magic stuff. Maybe I was doing it wrong.

neurom/tests/test_segments.py
juanchopanza: I added `taper` to `neurom.segments`.

mgeplf: ah, right, sorry

PR #214, opened 2016-01-15
added units Fixes #209

### LINKED ISSUE #209: Units in dendrogram plots
The dendrogram functionality of NeuroM should generate plots with units on the axes (or scale bars)

eleftherioszisis:


neurom/view/view.py
```diff
@@ -722,8 +722,8 @@ def dendrogram(obj, show_diameters=True, new_fig=True, new_axes=True, subplot=Fa
     kwargs['xlim'] = [- dnd.dims[0][0] * 0.5, dnd.dims[-1][0] * 0.5 + displacement]
 
     kwargs['title'] = kwargs.get('title', 'Morphology Dendrogram')
-    kwargs['xlabel'] = kwargs.get('xlabel', '')
-    kwargs['ylabel'] = kwargs.get('ylabel', '')
+    kwargs['xlabel'] = kwargs.get('xlabel', 'micrometers (um)')
+    kwargs['ylabel'] = kwargs.get('ylabel', 'micrometers (um)')
```
liesbethvanherpe: I may be missing something, but how can you we put units if they are not part of the data format specification?

eleftherioszisis: The standard practice is that morphologies units are in micrometers. If this is not the case, the user can always change the labels.

juanchopanza: Can we just add "unit of length" instead? Then it is up to the caller to set these. At least that way they know if is a length and not something else.

lidakanari: I think a "unit of length" will be a bit confusing. We are assuming the units are um, unless the user selects a different unit. I think in this case, this should be an option. @wvangeit Could you please confirm what makes more sense for you concerning your issue https://github.com/BlueBrain/NeuroM/issues/209?

juanchopanza: I think we shouldn't make assumptions if the format doesn't specify micrometers. If we don't know for sure what the units are, then we don't say anything and leave it up to the user to set the units. But this process should be made as easy as possible, and it should be clear to the user that they are expected to do this. For example, setting up a default of "unknown units" in the interface might help.

wvangeit: Showing 'unknown units' is ok for me. Is there btw a format that doesn't have a unit specification ?

juanchopanza: Well, it seems `SWC` is one! If I were to do this, I would have a header section with units and also information on how the soma is represented.

wvangeit: Here the units are mentioned, no ? http://www.neuronland.org/NLMorphologyConverter/MorphologyFormats/SWC/Spec.html

lidakanari: Thanks for the reference :) I think we can go for the micrometer units for now. (Any idea about h5 specs about length units?)

juanchopanza: Actually, it is specified as micometers [here](http://www.sciencedirect.com/science/article/pii/S0165027098000910), so I think this is OK. We will have to write a spec of our own collecting all this information!

PR #216, opened 2016-01-16
Remove unused iterator functions from analysis.morphtree

juanchopanza:
Remove neurom.analysis.morphtree map iterator functions whose
functionality has been re-factored into generalized iterations.


juanchopanza:
@lidakanari @eleftherioszisis FYI I removed most of the `i_something` iterator mappers from `analysis.morphmath`. They are not needed anymore with the new generalized iteration scheme and some of them were causing confusion.

lidakanari:
Sweet... Much more clear!

eleftherioszisis:
It is indeed much better now.

Issue #210, opened 2016-01-15
Viewing dendrogram requires pylab.show() call


wvangeit:
After calling 'dendrogram()' the figure window closes immediately if I don't put a pylab.show(). Maybe you can add an option as argument that enables/disables pylab.show() ? (And a white background would also be nice, but I leave that up to you)

lidakanari:
OK, I will fix this in the next commit.

lidakanari:
About the background color, it can be chosen while you save the image. If white is what you need, just choose the option transparent=False (default is True and probably that's why it is not white :)
About the pylab.show, that is chosen by default now. To disable chose show_plot=False.

juanchopanza:
Fixed by #219.

PR #219, opened 2016-01-19
Fix show plot in python mode

lidakanari:
This should fix issue #210


juanchopanza:
I'll merge this, when @wvangeit can confirm if it fixed the issue.

wvangeit:
Ok, thanks, it works. (Although at the moment I tried, the new version was not pushed to pypi yet)

juanchopanza:
Fixes #210.

juanchopanza:
@wvangeit We're only putting releases on pypi. I will make a new release (v 0.0.13) soon.

PR #220, opened 2016-01-19
Basic statistical tests to compare two datasets 

lidakanari:
This is useful to built more generic comparison tests in examples. Please suggest other names for the function if the proposed one doesn't make sense.


neurom/stats.py
```diff
@@ -132,3 +132,32 @@ def scalar_stats(data, functions=('min', 'max', 'mean', 'std')):
...
+        dist: float\
+            High numbers define high dissimilarity between the two datasets
+        p-value: float\
+            Small numbers define high probability the data come from\
+            same dataset.
+    '''
+    results = getattr(_st, test)(data1, data2)
+
+    stats = {'dist': results[0]}
+    stats['p-value'] = results[1]
```
juanchopanza: stats = {'dist': results[0], 'p-value': results[1]}

But maybe you can use a named tuple instead.

    Stats = namedtuple('Stats', ['dist', 'pvalue'])
    return Stats(*results) # assumes results has len == 2

lidakanari: A named tuple is a good idea. Will change to that!

chalimou: Hi,

first: I would take all ttests of scipy.stats in neuroM (middleterm), depending on how much work it is to integrate them.

second: which default? Depends on which is the most frequent case the customers have. I would take either the  scipy.stats.ttest_ind(a, b, axis=0, equal_var=True)  if the the distributions are more or less unimodal and symmetric or the mannwhitneyu otherwise (if they are highly skewed and obviously non-normal, for example). All the above tests work for continuous distributions ONLY. Special attention should be paid to equal_var=True or False depending on what one has. It switches between two different tests.

The reason not to make the kstest the default is that it is sensitive to all kinds of shape differences of the distributions. The tests mentioned above focus on differences in location parameters like the means (ttest) or medians (mannwhittneyu).

Another caution point is the wether the two samples are correlated (“related" in the scipy.stats language) or not. The corresponding versions of the tests are: scipy.stats.ttest_rel(a, b, axis=0)  for the ttest_ind and for the mannwhitneyu is the wilcoxon.

ATTENTION!!! the following statement is wrong. It’s the other way around: small p-values point to significantly different distributions. More specific: if pval < 0.05 then we reject the null hypothesis that the two distributions are the same. If pval>=0.05 we cannot reject the null, which DOES NOT MEAN THAT THE TWO DATA SETS COME FROM THE SAME DISTRIBUTION. What I say in this case is that: there is no evidence for significant difference, thus I do not undertake any action (to modify the system, for example).

I hope that helps, have fun,

nancy



> +        p-value: float\
> +            Small numbers define high probability the data come from\

> + same dataset.
<http://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.ttest_ind.html#scipy.stats.ttest_ind>
On 19 Jan 2016, at 18:23, lidakanari <notifications@github.com<mailto:notifications@github.com>> wrote:


In neurom/stats.py<https://github.com/BlueBrain/NeuroM/pull/220#discussion_r50143496>:

> +        test: string\
> +            Defines the statistical test to be used, based\
> +            on the scipy available modules.\
> +            Accepted tests: ks_2samp(default), mannwhitneyu
> +
> +    Returns:
> +        dist: float\
> +            High numbers define high dissimilarity between the two datasets
> +        p-value: float\
> +            Small numbers define high probability the data come from\
> +            same dataset.
> +    '''
> +    results = getattr(_st, test)(data1, data2)
> +
> +    stats = {'dist': results[0]}
> +    stats['p-value'] = results[1]


A named tuple is a good idea. Will change to that!

juanchopanza: I would prefer to have no defaults, and let the client code specify what it wants. The scripts can be a magic black box, but the modules should not be.

chalimou: Yes! I agree, the lack of defaults reduces also the danger of accidental misuse. And there is a lot of it in the statistical tests just because of the misunderstandings. The client should think and make a conscious decision.

nancy

On 20 Jan 2016, at 10:10, Juan Palacios <notifications@github.com<mailto:notifications@github.com>> wrote:


In neurom/stats.py<https://github.com/BlueBrain/NeuroM/pull/220#discussion_r50229369>:

> +        test: string\
> +            Defines the statistical test to be used, based\
> +            on the scipy available modules.\
> +            Accepted tests: ks_2samp(default), mannwhitneyu
> +
> +    Returns:
> +        dist: float\
> +            High numbers define high dissimilarity between the two datasets
> +        p-value: float\
> +            Small numbers define high probability the data come from\
> +            same dataset.
> +    '''
> +    results = getattr(_st, test)(data1, data2)
> +
> +    stats = {'dist': results[0]}
> +    stats['p-value'] = results[1]


I would prefer to have no defaults, and let the client code specify what it wants. The scripts can be a magic black box, but the modules should not be.

lidakanari:
Ok, I added ks-test because all the other tests in scipy are either buggy or not working properly. I can remove that, but then the user will probably get wrong results if he uses a buggy module.

juanchopanza:
That is fine. If there really is a bug in scipy we can add a warning. But do you have three lines of code that can reproduce the scipy bug(s)?

lidakanari:
Ok, I removed the default and updated the test accordingly. 

Example of problematic test: 
```python
data = np.array([1., 1., 2., 2.])
data_same = np.array([1.0, 1.0, 2.0, 2.0])
stats.compare_two(data, data_same, test='ks_2samp')
Stats(dist=0.0, pvalue=1.0) # ok

stats.compare_two(data, data_same, test='mannwhitneyu')
Stats(dist=8.0, pvalue=0.43433053149553336) # wrong
```

juanchopanza:
I mean code that unambiguously shows a bug in `scipy`, not our own code.

lidakanari:
Updated example of problematic test: 

```python
from scipy import stats
data = np.array([1., 1., 2., 2.])
data_same = np.array([1.0, 1.0, 2.0, 2.0])

stats.mannwhitneyu(data, data_same)
(8.0, 0.43433053149553336) #wrong, same as before
```

juanchopanza:
Maybe it would make sense to be explicit about the one method we do support:

    stats.compare_ks(data, data_same) # two is implicit from the number of parameters

We could also have a `stats.compare_two(a, b, str)` where we restrict the values of the strings.

lidakanari:
But we want to support any type of test in principle. Compare two is not a good name, but compare_ks is too restricted and I would like to avoid assumptions unless we only allow ks-tests to be applied. 

What do you mean restrict the string values? Could I generate a list of accepted distributions and check if the string is in the list? Otherwise I would raise an error. Does that make sense?

juanchopanza:
If only KS works at the moment, then we should only support that. There is no ambiguity because the name expresses the intent clearly. So which tests "work" besides Ks?
About the string, yes, that makes sense. We could also use an enum I guess.

lidakanari:
I would prefer the second option (enum will do) in order to avoid rewriting the same function if new tests are needed. If scipy has bugs-errors we cannot usually predict this (only if we try the tests one by one :P).

We can start simple: [ks_2samp, wilcoxon] and build from this set of tests adding anything that will make sense and is available in scipy.

juanchopanza:
That sounds good. If you want to see an example of `Enum` use, see `neurom/core/types.py`.

neurom/stats.py
```diff
@@ -34,11 +34,28 @@
...
 from scipy import stats as _st
 import numpy as np
-
+from enum import Enum, unique
 
 FitResults = namedtuple('FitResults', ['params', 'errs', 'type'])
 
 
+@unique
+class Stat_tests(Enum):
```
juanchopanza: From [pep8 naming conventions](https://www.python.org/dev/peps/pep-0008/#class-names), maybe better to use CamelCase here.

PR #223, opened 2016-01-20
fix core/tests/test_population.py so it can be run from any path

mgeplf:


mgeplf:
Too much cruft got in, I will redo

PR #224, opened 2016-01-20
Add dedicated disconnected components test data file.

juanchopanza:
SWC file with disconnected components. Comments section has list
of disconnected component root node IDs. Use this in relevant tests
instead of (identical) missing IDs file.

Reduce the required minir version of PyYAML.


juanchopanza:
@mgeplf The change to `requirements.txt` fixes the new travis CI problem.

Issue #222, opened 2016-01-20
Write function to unpack all trees in faulty SWC file


juanchopanza:
Neurons require well formed component tree structures as input. Some experimental files have disconnected trees. Although these aren't SWC conforming, it is of interest to unpack all available trees into a collection so that morphometrics can be obtained from them.

```python
trees = load_trees('some_file.swc')
```

juanchopanza:
Implemented in #227.

PR #227, opened 2016-01-20
Add function to load trees

juanchopanza:
neurom.io.utils.load_trees loads array of trees in SWC file.
Disconnected components are loaded as separate trees.

### DIFF
--- a/neurom/io/tests/test_utils.py
+++ b/neurom/io/tests/test_utils.py
@@ -28,9 +28,12 @@
 
 '''Test neurom.io.utils'''
 import os
+from itertools import izip
 import numpy as np
 from neurom import io
 from neurom.io import utils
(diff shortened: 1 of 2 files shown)


juanchopanza:
@lidakanari @eleftherioszisis This could be of interest. The iterators don't work naturally with it yet, that is the next step. In the mean time, you can hack it:

```python
class FakeNeuron(object):
    def __init__(self, trees):
        self.neurites = trees

fake_nrn = FakeNeuron(load_trees('some_filename.swc'))
sec_vols = list(iter_neurites(fake_nrn, sections.volume))
```

lidakanari:
Seems good to my poor understanding!

juanchopanza:
Fixes #222.

PR #230, opened 2016-01-20
Extend neurom.iter_neurites to work with iterables of trees.

juanchopanza:
Accepts lists of trees as well as single tree, neuron and neuron
population.


juanchopanza:
@eleftherioszisis @lidakanari This change makes iterating over lists of trees transparent (no need for mock neurons)

lidakanari:
Ok! So, this will iterate over a list of neurites and a "whatever object" with neurites, right?

juanchopanza:
Yes, and also an individual Tree object.

PR #232, opened 2016-01-20
Separate errors for disconnected components and missing parents.

juanchopanza:
Missing parents make no sense and should be irrecoverable.
Disconnected components are simply multiple trees, i.e. more than one element with parent ID of -1.


juanchopanza:
@eleftherioszisis @lidakanari FYI, different categories of errors for "many trees" (i.e. more than one component with parent -1) and "missing parents", i.e. components whose parent IDs don't correspond to any existing components. I will merge unless there are any objections or suggestions.

PR #239, opened 2016-01-21
total score for list of paired feature distrs resolves #225

### LINKED ISSUE #225: Summary of kstest scores from multiple distributions
I need to get a summarizing score from a set of statistical tests between pairs of distributions that represent different features:

```python
score1 = score(f1_dataset1, f1_dataset2)
score2 = score(f2_dataset1, f2_dataset2)
...

total_score = norm(scores, p) # scores: all previous || p: p-norm
```

eleftherioszisis:


neurom/stats.py
```diff
@@ -176,3 +176,24 @@ def compare_two(data1, data2, test=StatTests.ks):
...
+
+    Options:
+        p : integer that defines the order of p-norm
+        test: Stat_tests\
+            Defines the statistical test to be used, based\
+            on the scipy available modules.\
+            Accepted tests: ks_2samp, wilcoxon
+
+    Returns:
+        A float corresponding to the p-norm of the scores that have
```
lidakanari: To the distances that have been calculated.

lidakanari: You should also make sure you explain that "0" is high similarity and "1" is low similarity.

lidakanari:
I think this is good! +2 from me.

neurom/stats.py
```diff
@@ -176,3 +176,26 @@ def compare_two(data1, data2, test=StatTests.ks):
     Stats = namedtuple('Stats', ['dist', 'pvalue'])
 
     return Stats(*results)
+
+
+def total_score(paired_dats, p=2, test=StatTests.ks):
```
juanchopanza: Does it make sense for `p` to be an input parameter? Can we get different, meaningful results by changing only p?

lidakanari: If we assume that the results can be meaningful in the first place, then p is important. The relation between the results for different p-norms should have a monotonic relation, but are not expect to be equal. How we choose an optimal p is another mystery for me...

juanchopanza: That doesn't really answer my question. But if `p` can be independently varied for the same input data, then this should be tested.

lidakanari: Yes, it is important for p to be an input parameter because we need to test how the scores are modified according to p (and we don't have an optimal p to propose before we try many different ones).

juanchopanza: So, we need some tests that use that. In fact, we need more tests for this.

lidakanari: True! We could add a test to check if p is correctly interpreted. (Example: results for p=inf, p=1 should be different when distributions are not the same).

eleftherioszisis: Do we really need tests for numpy's p-norm?

juanchopanza: We need to test that you aren't screwing up the call to p-norm :-)

eleftherioszisis: ok. I added a couple of tests for norm 1 and norm 2.

lidakanari:
If it doesn't need more changes could we merge this?

juanchopanza:
OK, I'll merge. Please try to keep unrelated issues in separate pull requests.

PR #243, opened 2016-01-22
Script to compare the morphometrics of two sets of trees

lidakanari:
This will be used in BigNeuron Hackathon for comparison between different algorithms of automatic reconstruction.

More features will be added gradually.

### DIFF
--- a/examples/comparison.py
+++ b/examples/comparison.py
@@ -117,7 +117,7 @@ def boxplots(collect_all, new_fig=True, subplot=False,
 
     ax.set_yticklabels(feature_titles)
 
-    fig, ax = common.plot_style(fig, ax, xlabel='Normalized units (dimensionless)',
+    fig, ax = common.plot_style(fig, ax, xlabel='Normalized units (dimensionless)', ylabel=''
                                 title='Summarizing features', tight=True, **kwargs)
 
     return fig, ax
--- a/examples/comparison.py
+++ b/examples/comparison.py
@@ -87,7 +87,7 @@ def boxplots(collect_all, new_fig=True, subplot=False,
                              'Section area',
                              'Section path length',
                              'Segment lengths',
-                             'Segment meander angles')):
+                             'Segment meander angles'), **kwargs):
     '''Plots a list of boxplots for each feature in feature_list for object 1.
     Then presents the value of object 2 for each feature as an colored objected
     in the same boxplot.
@@ -118,6 +118,6 @@ def boxplots(collect_all, new_fig=True, subplot=False,
     ax.set_yticklabels(feature_titles)
 
     fig, ax = common.plot_style(fig, ax, xlabel='Normalized units (dimensionless)',
-                                title='Summarizing features')
+                                title='Summarizing features', tight=True, **kwargs)
 
     return fig, ax
--- a/examples/comparison.py
+++ b/examples/comparison.py
@@ -31,13 +31,34 @@
 import numpy as np
 from neurom.core import iter_neurites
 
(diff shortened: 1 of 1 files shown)


examples/comparison.py
```diff
@@ -0,0 +1,123 @@
...
+        # Standardization of data: (data - mean(data))/ std(data)
+        m = np.mean(feature_pop)
+        st = np.std(feature_pop)
+
+        collect_all.append([(feature_pop - m) / st, (feature_neu - m) / st])
+
+    return collect_all
+
+
+def get_features(object1, object2, tri_list=('meander_angle',),
```
juanchopanza: I don't think it is a good idea to pass the features as strings and then try to pull them from a module. Better to make a map, for example,

```python
feature_map = {'segment_length': segments.length, 'section_length': sections.length, ...}
```

and so on. Less implicit stuff, and less scope for difficult to debug runtime errors.

examples/comparison.py
```diff
@@ -0,0 +1,123 @@
...
+
+import numpy as np
+from neurom.core import iter_neurites
+
+
+def get_mod_features(object1, object2, flist, collect_all, mod):
+    '''Computes features from module mod'''
+    for feat in flist:
+
+        feature_pop = np.array([l for l in iter_neurites(object1, getattr(mod, feat))])
```
juanchopanza: See note below.

lidakanari:
@juanchopanza  I've just send an update with an implementation of features as you proposed. It's much more clear this way :)

examples/comparison.py
```diff
@@ -0,0 +1,106 @@
...
+from neurom import sections as sec
+
+feature_map = {'Meander_angle': tri.meander_angle,
+               'Segment_length': seg.length,
+               'Bif_angles': bif.local_angle,
+               'Section_path_length': sec.end_point_path_length,
+               'Section_area': sec.area,
+               'Section_length': sec.length}
+
+features = ('Meander_angle',
```
juanchopanza: I think you can say

```python
features = feature_map.keys()
```

unless you want to have separate control over the default set of features, of course.

examples/comparison.py
```diff
@@ -0,0 +1,106 @@
...
+    for feat in flist:
+
+        feature_pop = np.array([l for l in iter_neurites(object1, feature_map[feat])])
+        feature_neu = np.array([l for l in iter_neurites(object2, feature_map[feat])])
+
+        # Standardization of data: (data - mean(data))/ std(data)
+        m = np.mean(feature_pop)
+        st = np.std(feature_pop)
+
+        collect_all.append([(feature_pop - m) / st, (feature_neu - m) / st])
```
juanchopanza: Does it make sense to use `feature_pop`'s mean and standard deviation with `feature_neu`?

examples/comparison.py
lidakanari: That's what I wanted: to have a 'library' of available features (feature_map) and then the user chooses what to use from this one. If it makes sense I would like to keep those two separated.

examples/comparison.py
lidakanari: Yes, because you normalize according to the data of the boxplot and you want to position another cell in this normalized range. It's like counting how many standard deviations apart in a datapoint from the population, and then you need to normalize the datapoint according to the population.

lidakanari: I will let @chalimou confirm the previous, if what I describe is understandable.

juanchopanza: OK, I will merge this. We can change it afterwards if needed.

PR #246, opened 2016-01-28
Add summary output file to morph_check.

juanchopanza:
Output file name controlled with option -o or --output.
JSON format file with pairs of testname : PASS/FAIL for each
tested file. Client code can unpack this into less verbose
representations (e.g CSV tables.)

### DIFF
--- a/apps/morph_check
+++ b/apps/morph_check
@@ -248,7 +248,7 @@ if __name__ == '__main__':
     L.info(SEPARATOR)
 
     with open(args.output_file, 'w') as json_output:
-        json.dump(summary, json_output)
+        json.dump(summary, json_output, indent=4)
 
     exit_status = 0 if res else 1
 
--- a/apps/morph_check
+++ b/apps/morph_check
@@ -38,6 +38,8 @@ import logging
 import yaml
 import os
 import sys
+import json
+
 
 DESCRIPTION = '''
 NeuroM Morphology Checker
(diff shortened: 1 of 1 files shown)


apps/morph_check
```diff
@@ -132,24 +137,29 @@ def log_msg(ok, msg, color=False):
...
+    if summary_dict is not None:
+        summary_dict[msg] = 'PASS' if ok else 'FAIL'
+
 
 def test_file(f, config):
     '''Run tests on a morphology file'''
 
     L.info('Check file %s...', f)
 
+    summary_ = {}
```
mgeplf: why the trailing _?

juanchopanza: pylint. `summary` is used in `__main__`.

apps/morph_check
```diff
@@ -231,6 +247,9 @@ if __name__ == '__main__':
 
     L.info(SEPARATOR)
 
+    with open(args.output_file, 'w') as json_output:
+        json.dump(summary, json_output)
```
mgeplf: I'd set `indent` so that it's more human readable.

mgeplf:
Looks fine to me.

apps/morph_check
mgeplf: Ah, I would make __main__ shorter and just call main().  Makes it easier to test, too

juanchopanza: I will do that in a future pull request. There's a lot here that needs cleaning up.

PR #248, opened 2016-02-01
Initial morph_check re-factoring

juanchopanza:
Move some logic out of __main__ into helper function.

### DIFF
--- a/apps/morph_check
+++ b/apps/morph_check
@@ -141,23 +141,23 @@ def log_msg(ok, msg, color=False, summary_dict=None):
         summary_dict[msg] = 'PASS' if ok else 'FAIL'
 
 
-def test_file(f, config):
+def check_file(f, config):
     '''Run tests on a morphology file'''
 
     L.info('Check file %s...', f)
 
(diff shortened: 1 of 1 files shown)


apps/morph_check
```diff
@@ -185,22 +186,57 @@ def test_file(f, config):
...
+                            'has_apical_dendrite',
+                            'nonzero_segment_lengths',
+                            'nonzero_section_lengths',
+                            'nonzero_neurite_radii'],
+          'options': {"nonzero_neurite_radii": 0.007,
+                      "nonzero_segment_lengths": 0.01,
+                      "nonzero_section_lengths": 0.01}}
 
 
 if __name__ == '__main__':
```
mgeplf: While you're here, I think I would try and pull more out of __main__, and put it in a main() function, personally; get as much as you can out of the global namespace

juanchopanza: I will leave that for later. It requires fixing the logging, because there is a re-definition of `log_msg` in the current `main`.

PR #250, opened 2016-02-01
morph_check refactoring

juanchopanza:
Move most code from main to helper function.
Main simply parses args, passes them to helper, and collects
exit status.


mgeplf:
Looks good

PR #252, opened 2016-02-04
features module

eleftherioszisis:
This is the module which stores the user level interface in order to extract features from objects. In the __init_.py there is the features_factory function along with the enumerations with the available features (not complete yet). It can be used int he following way:

```python
from neurom import features as fs

func = fs.features_factory(fs.NeuriteFeatures.section_lengths)

result = func(neuron)
result = func(neurite)
result = func(population)
```
where the enumeration maps to the respective neurite_features or neurite_neurons file and returns the respective feature function by matching its name with the function name.

I have moved the packaging function one level higher and thus it envelops the iterator function that are returned from the factory. In other words all the feature functions return iterators which are packaged in the feature_factory level with the _pkg decorator.

I am not sure that the enumerations are more useful than a dict, but I leave this up to you.

What do you think?


eleftherioszisis:
Tests don't fail in my machine. Peculiar! :/

juanchopanza:
I would prefer to pass the feature enumeration to a function:

```python
seclen = get_feature(neuron, NeuriteFeatures.section_lengths)
seclen = get_feature(pop, NeuriteFeatures.section_lengths)
seclen = get_feature(neurite, NeuriteFeatures.section_lengths)
```

and so on. I don't think the `features_factory ` is necessary.

neurom/features/__init__.py
```diff
@@ -0,0 +1,87 @@
...
+    section_radial_distances = 6
+    local_bifurcation_angles = 7
+    remote_bifurcation_angles = 8
+    neurite_number = 9
+
+
+class NeuronFeatures(_Enum):
+    '''Neuron Features
+    '''
+    soma_radius = 10
```
juanchopanza: Why start at 10?

eleftherioszisis:
Well, if we do not need to access the arguments of the respective function then I can do it like this. I thought it is limiting to prevent the user from accessing all the function's arguments.

neurom/features/neurite_features.py
```diff
@@ -0,0 +1,110 @@
...
+
+from neurom.core.types import TreeType
+from neurom.core.types import tree_type_checker as _ttc
+from neurom import segments as _seg
+from neurom import sections as _sec
+from neurom import bifurcations as _bifs
+from neurom import iter_neurites
+
+
+def section_lengths(neurites, neurite_type=TreeType.all):
```
juanchopanza: Isn't there already a simple way of getting this kind of information? It seems like we're re-writing every feature calculation over and over again.

neurom/features/__init__.py
eleftherioszisis: No reason. :P 
will change it in the next commit.

neurom/features/neurite_features.py
eleftherioszisis: What do you mean? The goal of this module is to eventually replace ezy and store all the possible features that can be extracted from the user and from the apps in order to avoid low level operations. Do you mean to create some kind of aliasing instead of creating a new function?

juanchopanza: RIght, otherwise you might as well just call this thing, and remove the whole enum thing. Which is another possibility.

eleftherioszisis: Hmm, I am still trying to follow your train of thought. If we had to call the iter_neurites directly then we would need a lambda which is essentially a different syntax for the same thing, isn't it?

juanchopanza: I am looking at the amount of code that has to be written for  a single feature, and it looks like too much. That leads me to think that the enum is a bad idea (sorry) and that we might as well have functions. Now, in most cases, these functions map to something without parameters other than a tree type filter (e.g. `get_segment_length`). So we have a most common use case where we have a functor (e.g. `sections.length`), `iter_neurites`, a tree type, and we want to return a `numpy` array. I think this can be easily expressed in terms of a decorator and the already existing functions. Then we only really need to write code for the functions that require extra parameters.

eleftherioszisis: Is this sufficient for the simple feature functions without extra arguments?
```python
def _wrapper(mapfun):
    def _wrapped(neurites, neurite_type=TreeType.all):
        '''Extracts feature from an object with neurites, i.e. either neurite, neuron, or population
        '''
        return iter_neurites(neurites, mapfun, _ttc(neurite_type))
    return _wrapped


section_lengths = _wrapper(_sec.length)
segment_lengths = _wrapper(_seg.length)
local_bifurcation_angles = _wrapper(_bifs.local_angle)
remote_bifurcation_angles = _wrapper(_bifs.remote_angle)
```

juanchopanza: I think for the `ezy` style behaviour in the first example we also need something that builds an iterable from the iterable (e.g. `numpy.array`, `list` or `tuple`.) It would be an optional parameter of the decorator, set to a reasonable default.

juanchopanza: Also I don't think we need these underscores for the decorators. They are useful outside of the module, if people want to wrap their own iterators.

eleftherioszisis:
How about this? I removed all the complicated dispatching with the enumerations and I replaced it with two dictionaries and the respective functions. Clean and clear! :P

neurom/features/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+''' Functionality for Feature Extraction'''
+
+
+from decorator import decorator as _dec
+import numpy as _np
+from neurom.features import neurite_features as _neuf
+from neurom.features import neuron_features as _nrnf
+
+
+@_dec
```
juanchopanza: `decorator` would have been fine.

neurom/features/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+def _pkg(func, *args, **kwargs):
```
juanchopanza: I'd probably call this `make_iterable`

neurom/features/neurite_features.py
```diff
@@ -0,0 +1,99 @@
...
+from neurom.core.types import TreeType
+from neurom.core.types import tree_type_checker as _ttc
+from neurom import segments as _seg
+from neurom import sections as _sec
+from neurom import bifurcations as _bifs
+from neurom import iter_neurites
+from functools import wraps
+
+
+def wrapper(mapfun):
```
juanchopanza: `feature_getter`?

neurom/features/neurite_features.py
```diff
@@ -0,0 +1,99 @@
...
+    ''' Wrapper around already existing feature functions
+    '''
+    def wrapped(neurites, neurite_type=TreeType.all):
+        '''Extracts feature from an object with neurites, i.e. either neurite, neuron, or population
+        '''
+        return iter_neurites(neurites, mapfun, _ttc(neurite_type))
+    return wrapped
+
+
+def number(f):
```
juanchopanza: Maybe `count`?

neurom/features/neuron_features.py
```diff
@@ -0,0 +1,60 @@
...
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+''' Neuron Related Features'''
+import math
+from decorator import decorator as _dec
+from neurom.core.neuron import Neuron
+
+
+@_dec
+def _make_iterable(func, *args, **kwargs):
```
juanchopanza: Does this really make an iterable? In pyhton, and iterable is a list, dict, set, array etc.

neurom/features/neuron_features.py
```diff
@@ -0,0 +1,60 @@
...
+    ''' Takes care of the neuron feature input. By using this decorator the neuron functions
+    can take as an input a single neuron, list of neurons or a population.
+    '''
+    obj = args[0]
+    neurons = [obj] if isinstance(obj, Neuron) else (obj.neurons if hasattr(obj, 'neurons')
+                                                     else obj)
+    args = tuple([neurons] + [a for i, a in enumerate(args) if i > 0])
```
juanchopanza: Wouldn't it be simpler to say `args[0] = neurons`? Or does this do something else?

neurom/features/neuron_features.py
eleftherioszisis: hmm, no. I think I randomly chose this name. I will change it.

eleftherioszisis: Also I think it would be better if we had an iter_neurons functions that would behave as iter_neurites and thus have similar functionality for the neuron/population specific features (e.g. somata features, probably features that treat the neuron as a whole (spatial?))

juanchopanza: Could you add an issue for that, with a precise description of what is needed? Then I can work on it.

neurom/features/neuron_features.py
eleftherioszisis: yep, I generalized it and it is not needed

neurom/features/neuron_features.py
eleftherioszisis: I think it is better if we put first some other features that are specific for the neurons in order to see if there is a pattern and then decide on using the ```iter_neurites``` example.

eleftherioszisis:
@juanchopanza if the new changes are ok, you can merge it. I will later add the rest of the features

neurom/features/neuron_features.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@iter_neurons
+def soma_surface_area(neurons):
+    '''Get the surface area of the soma.
+
+    Note:
+        The surface area is calculated by assuming the soma is spherical.
+    '''
+    return (4. * math.pi * nrn.get_soma_radius() ** 2 for nrn in neurons)
```
juanchopanza: Could you call `nrn.get_soma_surface_area()` here?

eleftherioszisis: sure!

eleftherioszisis:
Mergeminate!
Logo

sorry, just discovered the logo macro :P

neurom/features/neuron_features.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@iter_neurons
+def soma_surface_area(neurons):
+    '''Get the surface area of the soma.
+
+    Note:
+        The surface area is calculated by assuming the soma is spherical.
+    '''
+    return (nrn.get_soma_surface_area() for nrn in neurons)
```
juanchopanza: There's still the issue that this only works for `ezy.Neuron` objects. But you can use `nrn.soma.radius`, and maybe a sphere_area function. It was in an edit to my previous comment,maybe it got lost.

eleftherioszisis: oops yes you are right! will fix.

eleftherioszisis: I am wondering where should I put this sphere area function? Should I temporarily put it inside NeuronFeatures and later you can move it somewhere else (maybe a place dedicated for somata functions etc)?

juanchopanza: Maybe `morphmath`? I want to change the name of that module, but it seems to be for geometrical stuff that doesn't depend on trees etc.

eleftherioszisis: Yes, I was about to send the same. :P

PR #254, opened 2016-02-10
linking to more features in features module

eleftherioszisis:


neurom/features/neurite_features.py
```diff
@@ -97,3 +116,53 @@ def section_path_distances(neurites, use_start_point=False, neurite_type=TreeTyp
     magic_iter = (_sec.start_point_path_length if use_start_point
                   else _sec.end_point_path_length)
     return iter_neurites(neurites, magic_iter, _ttc(neurite_type))
+
+
+@ensure_neurite_list
+def section_radial_distances(neurites, origin=None, use_start_point=False,
```
eleftherioszisis: @juanchopanza 
Is there a way to avoid this hacky functor for the radial distances. I really don't like this :P

juanchopanza: I don't like it either, but I can't think of a better way to do it at the moment.

juanchopanza:
Looks OK to me.

eleftherioszisis:
Should I rename iter_neurons decorator to ensure_neuron_list in order to be consistent with the respective one in the neurite_features file?

juanchopanza:
Maybe `as_neuron_list` is better? You could also re-name the other `ensure_*` decorator similarly.

neurom/analysis/morphtree.py
```diff
@@ -222,8 +222,8 @@ def partition(tree):
     '''
     def partition_at_point(bif_point):
         '''Partition at each bif point.'''
-        n = n_sections(bif_point.children[0])
-        m = n_sections(bif_point.children[1])
+        n = float(n_sections(bif_point.children[0]))
```
juanchopanza: Is there a test for this issue? If not, you should add one. Anything "fixed" needs a test that would fail without the fix.

eleftherioszisis: The test was not failing because the division was with 1, giving the correct result. I added the partition in the bifurcations file, and added a general test that does fail in the case of integer division.

juanchopanza:
Looks OK to me. If you're done I can merge.

eleftherioszisis:
I am for today :P

Issue #256, opened 2016-02-11
Application for the validation of a neuron against a population


lidakanari:
We need an application similar to morph_check for the automatic validation of a morphology. An example use-case is:
```
morph_validate neuron1 population
```
which should result in:
```
Section lengths (test:ks-test, threshold:10%) : PASS
Section path distances (test:ks-test, threshold:10%) : FAIL
ALL: FAIL
```

eleftherioszisis:
Maybe having also as an optional input the yaml configuration file with the selected features?

lidakanari:
Of course! The idea is to reproduce morph_check but with a different type of tests that will represent actual validations.

PR #260, opened 2016-02-12
Initial application for validation of a neuron against a population

lidakanari:
Based on morph_check I created a morph_validation to add further checks on the quality of a cell.


lidakanari:
This implements #256. It is a very primitive state, but I think it will be useful for our validation task requirements.

juanchopanza:
There seems to be quite a bit of duplication with `morph_check`. Also, we need to move towards less untested code in apps and more tested code in modules. That is what I was hoping to do with `morph_check` whenever I get the time. Ideally, an app should only parse the commandline options, maybe open some configuration files, and delegate all the work to a function from a neurom module.

lidakanari:
Quite a bit of duplication indeed! I agree with you, but since this moving around is not urgent and this validation app is needed by Monday I would suggest that we leave the moving around for a subsequent pull request.

juanchopanza:
I don't think that is a good idea. If it really is that urgent, it can be put elsewhere, outside of NeuroM. But the best approach is to do this properly.

lidakanari:
I definitely agree that we should do this properly. However, I don't see the point of cleaning up this application, while the morph_check application is in a similar state without problems. What I could definitely do is move the check_feature function outside of the applications and add the corresponding tests.  But I think the logging features are not relevant for this specific request. Also, since I am not very familiar with logging this might take me too long to automate it properly. If you have some reasonable suggestions (concerning which functions can be moved in a different place without too many modifications) please let me know. But, I will need a more concrete suggestion than this needs to be cleaned up :P

juanchopanza:
Maybe I can re-work `morph_check` next week and you can use it as an example.

apps/morph_validate
```diff
@@ -0,0 +1,291 @@
...
+def parse_args():
+    '''Parse command line arguments'''
+    parser = argparse.ArgumentParser(description=DESCRIPTION,
+                                     formatter_class=argparse.RawTextHelpFormatter,
+                                     epilog=EPILOG)
+    parser.add_argument('datapath',
+                        help='Path to morphology data file or directory')
+
+    parser.add_argument('targetpath',
+                        help='Path to population morphology data file or directory')
```
juanchopanza: This is going to be tricky, unless somebody puts the reference morphologies on the document service. Shouldn't we be comparing against some summary statistics values?

lidakanari: I'm not sure what's the best practice for this but I assume we do need to provide the morphologies against which we will be testing. Alternatively we could provide the output distributions, as a different configuration file (?), but statistical values will not be sufficient for the tests that Nancy is interested in, as far as I am aware.

eleftherioszisis:
Is this going to be merged at some point? I would like to use it for Guy's synthesized glial cells.

juanchopanza:
No in its current form. I think this needs to go somewhere else.

eleftherioszisis:
Apps is not a good location for this? Or do you mean that it needs to be re-factored in a way to have minimal code in apps? If this is the case then I agree.

apps/morph_validate
berthelm: For M30 we are only providing a fixed set of m-types that the user can recreate, so we could consider putting reconstructed cells of these m-types somewhere and providing a direct path to this directory. But this does not scale, as we should not be dumping morphologies randomly in directories but retrieving them somehow from a persistent storage in the NIP. Maybe for M30, it is better to create summary statistics files. We should talk to Nancy.

juanchopanza:
Neither apps not neurom. In its current form this is not general enough. Some code could be factored out into modules, and if light and more general application can be build using them, then fine. Otherwise this should go in a BBP morphology validation package.

lidakanari:
So this comment: 

> juanchopanza commented 7 days ago
> 
> I don't think that is a good idea. If it really is that urgent, it can be put elsewhere, outside of NeuroM. But the best approach is to do this properly.

is outdated?

eleftherioszisis:
I think I cannot follow your reasoning. NeuroM extracts the distributions of the features and what the script is essentially doing is calculating a score between distributions that are extracted from NeuroM.

Do we really need a different package in order to have a function that loads from two different directories and one that applies a bunch of scipy tests on the distributions of NeuroM? I thought that was the point behind having apps, wasn't it?

I whole-heartedly agree with refactoring the code in order to have only trivial stuff in the apps, but I am not sure why is it is to have a function in stats that runs a test between two distributions.

lidakanari:
So this comment: 
 
> juanchopanza commented 7 days ago
> 
> Maybe I can re-work morph_check next week and you can use it as an example.

is outdated?

lidakanari:
I don't think that where the validation - comparison code will go should be **a personal** decision. 

I understand your point about cleaning up the code and I'm strongly in favor of cleaner + tested code, before we add it to an app. I don't see how comparing A with B (as long as the same features can be extracted from A and B) is not a general enough use case. FYI I have a least 3 use cases that need to use this type of comparisons, which makes it general enough. On top of that 2 of them are outside BBP, which makes the BBP repository suggestion a very weak idea. I though the point of NeuroM was to provide tools for a larger audience and therefore it was OS. If someone needs special BBP access to run basic scripts that weakens the applicability of this system.

juanchopanza:
Maybe I can clarify the two of the main issues:

1. Not general: as far as I can see, this is a very specific test on two populations. it is not a universal validation (there is no such thing), it is just a set of statistical tests chosen for a specific purpose. This could be made more general for example by allowing the user to select what kind of test they want to perform.

2. Code quality. I think there is agreement on this.

If we can make an application after fixing 1. and 2. then `neurom` is the right place for it. Otherwise, it should go elsewhere. But I really can't see why this is a problem at this stage. If we need something for M30, then we put whatever we have an internal repository used for morphology validations. If somebody outside of BBP needs it, then we can thing of providing it as an example. Bear in mind that applications used in the task service are tied to the release cycle of NeuroM and we won't be making a release every time some script gets modified. So it is good to have M30 scripts de-coupled from the NeuroM release cycle anyway.

PR #263, opened 2016-02-16
refactor: move raw data block parsing out of load_data

juanchopanza:
neurom.io.readers._unpack_raw_data(filename) parses input and produces a 2D
array with internal representation of neuron morphology. This
functionality is an implementation detail of
neurom.io.load_data, but it can be useful for debugging purposes.

Note: if and when the internal data structure is formalized and specified, this could be made public.

### DIFF
--- a/neurom/io/__init__.py
+++ b/neurom/io/__init__.py
@@ -28,5 +28,5 @@
 
 ''' IO operations module for NeuroM '''
 
-from .readers import load_data
+from .readers import load_data, unpack_raw_data
 from . import check
--- a/neurom/io/__init__.py
+++ b/neurom/io/__init__.py
@@ -28,5 +28,5 @@
 
 ''' IO operations module for NeuroM '''
 
-from .readers import load_data, unpack_raw_data
+from .readers import load_data
 from . import check
--- a/neurom/io/readers.py
+++ b/neurom/io/readers.py
@@ -51,6 +51,45 @@ from neurom.core.dataformat import ROOT_ID
 from neurom.utils import memoize
 
 
+def unpack_raw_data(filename):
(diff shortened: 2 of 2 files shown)


juanchopanza:
@jdcourcol This could be of interest for the morphology viewer.

juanchopanza:
@eleftherioszisis @mgeplf Any objections to this?

mgeplf:
If it's made accessible then it's basically something that can't be changed very easily.  I wonder a couple things:
* TYPE, ID, P are currently stored as floats, but they really should be ints
* is ID needed, as it usually is implicitly the row number (not sure if this is true in all cases, if not, could you point me to an example where it's not?)

juanchopanza:
@mgeplf Yes, that is a good point. I'll have to re-thing this. About the other issues,

* If there was an easy, cost free way to have mixed type rows, then I could use appropriate types
* ID is kind-of needed because SWC starts at an arbitrary positive value and there can be gaps in ID. I thought it would be easier to preserve this structure, especially because initially we spoke about this "round-trip" business. Of course, it can be changed, but it would require some work.

mgeplf:
Have you looked at Structured Arrays: http://docs.scipy.org/doc/numpy-1.10.1/user/basics.rec.html

juanchopanza:
I know of their existence. I'd have to evaluate any impact on performance.

juanchopanza:
Closing this for now. The internal data block shouldn't be public until we have a spec.

juanchopanza:
@mgeplf I made it "private" but kept it out of `load_data` because it might be useful for debugging. It would be nice to have a standardized internal representation at some point in the future.

PR #265, opened 2016-02-16
fixed bug where some labels had spaces

mgeplf:


neurom/io/neurolucida.py
```diff
@@ -88,7 +88,19 @@ def _get_tokens(morph_fd):
     for line in morph_fd.readlines():
         line = line.rstrip()   # remove \r\n
         line = line.split(';', 1)[0]  # strip comments
+        squash_token = []
         for token in line.replace('(', ' ( ').replace(')', ' ) ').split():
+            if squash_token:
+                print squash_token
```
juanchopanza: We probably don't want to be printing from here.

juanchopanza:
Running `make test` locally could save come CI time.

neurom/io/neurolucida.py
mgeplf: yes, the last few days have been bad for this sort of thing, le sigh

PR #267, opened 2016-02-18
A zigzagging check for morphologies

eleftherioszisis:
This check is implemented for the following issue:
https://bbpteam.epfl.ch/project/issues/browse/SCIMORPH-9

The algorithm checks if the endpoint of a segment lies within or close to another segment upstream.

Firstly, it checks that the distance of the projection of the vector from the center of the other segments to the endpoint of the running one is smaller than the half length of that segment plus a 5% tolerance to take into account exact overlaps.

Secondly, the orthogonal complement of the projection must be smaller or equal than the sum of the maximum radius of each segment pair.

Zero segments are ignored.

### DIFF
--- a/neurom/analysis/tests/test_morphmath.py
+++ b/neurom/analysis/tests/test_morphmath.py
@@ -47,21 +47,54 @@ def test_vector():
 
 
 def test_scalar_projection():
-
-    v1 = np.array([4., 1.])
-    v2 = np.array([2., 3.])
+    v1 = np.array([4., 1., 0.])
+    v2 = np.array([2., 3., 0.])
 
     res = mm.scalar_projection(v1, v2)
     nt.assert_true(np.isclose(res, 3.0508510792387602))
 
 
-def test_vector_projection():
+def test_scalar_projection_collinear():
+    v1 = np.array([1., 2., 0.])
+    v2 = np.array([4., 6., 0.])
 
-    v1 = np.array([4., 1.])
-    v2 = np.array([2., 3.])
+    res = mm.scalar_projection(v1, v2)
+
+    nt.assert_true(np.allclose(res, 16./np.linalg.norm(v2)))
+
+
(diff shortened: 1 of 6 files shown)


eleftherioszisis:
In the end, I did not use the distance of line to point that I added and tested in morphmath, because I calculated the distance from the complement vector of the projection, which was much faster. However, I kept the function in morphmath, because it might come handy. I can remove it if you think that it is unnecessary .

juanchopanza:
Is this really zigzagging? From the description in the link, it sounds like a section turns back on itself.

eleftherioszisis:
Well, to be honest I had no idea how to name it... :P 
You are right, a segment turns back to a previous one.

juanchopanza:
Maybe "back-track"?

neurom/analysis/tests/test_morphmath.py
```diff
@@ -57,21 +42,57 @@
 def test_vector():
     vec1 = (12.0, 3, 6.0)
     vec2 = (14.0, 1.0, 6.0)
-    vec = vector(vec1,vec2)
+    vec = mm.vector(vec1,vec2)
     nt.ok_(np.all(vec ==(-2.0, 2.0, 0.0)))
 
 
+def test_scalar_projection():
```
juanchopanza: Could you test some edge cases (collinear, perpendicular)

neurom/analysis/tests/test_morphmath.py
```diff
@@ -57,21 +42,57 @@
...
+
+    v1 = np.array([4., 1.])
+    v2 = np.array([2., 3.])
+
+    res = mm.scalar_projection(v1, v2)
+    nt.assert_true(np.isclose(res, 3.0508510792387602))
+
+
+def test_vector_projection():
```
juanchopanza: Same comment

PR #270, opened 2016-02-18
Update neuron_features.py

orena1:
Adding units to soma area ,and soma radius

### DIFF
--- a/neurom/features/neuron_features.py
+++ b/neurom/features/neuron_features.py
@@ -52,13 +52,13 @@ def as_neuron_list(func):
 
 @as_neuron_list
 def soma_radii(neurons):
-    '''Get the radius of the soma'''
+    '''Get the radius of the soma (\xce\xbcm).'''
     return (nrn.soma.radius for nrn in neurons)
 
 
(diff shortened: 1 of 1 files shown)


eleftherioszisis:
Hello,
the error that you get is the following from pylint (you can see it by clicking details):
```
Cannot decode using encoding "ascii", unexpected byte at position 50 (invalid-encoded-data)
```
I am not an encoding standards expert but the general idea is that the string is decoded assuming an ascii encoding, which is the default behaviour, while your special unicode characters are in UTF-8 encoding or something similar and thus the ascii decoder does not understand them.

Different encodings must be handled with care in order to make sure that you will not run into surprises! 
Check the following link if you want to see a simple explanation on the different encodings:
http://nedbatchelder.com/text/unipain/unipain.html#1

Now, as far as the change is concerned it is in general better to avoid special encoding, unless your application needs to process it. Thus, you can change it to um and um2 respectively.

Hope that this helps :)

juanchopanza:
I'm closing this pull request as it is very minor and is creating problems.

PR #275, opened 2016-02-24
Forgot to add the section number per neurite in the init

eleftherioszisis:


juanchopanza:
Can we make this `nsections` or `number_of_sections`? `section_number` means something else to me. Like the number of a section.

eleftherioszisis:
Sure.```number_of_sections``` is good and clear I think.

eleftherioszisis:
If this is ok, can you merge it? I have another pull request waiting on the line. :P

juanchopanza:
Done. You should probably use branches for separate features so you don't get blocked by pending merges. It is more involved than with gerrit, but it is workable.

PR #277, opened 2016-02-25
fraction coordinates on segment and section

eleftherioszisis:
These functions permit the extraction of the 3D coordinates that correspond to a fraction of the length of a segment or along the piecewise linear curve of the section. @dkeller9 asked me to implement these functions for they are useful for synapse location calculations.

### DIFF
--- a/neurom/analysis/morphmath.py
+++ b/neurom/analysis/morphmath.py
@@ -47,7 +47,7 @@ def vector(p1, p2):
     return np.subtract(p1[0:3], p2[0:3])
 
 
-def two_points_fraction(p1, p2, fraction):
+def linear_interpolate(p1, p2, fraction):
     '''Returns the point p where:
         |p1p| = fraction * |p1p2|
     '''
@@ -87,7 +87,7 @@ def path_fraction_point(points, fraction):
     frac_length = fraction * path_distance(points)
     p0, p1, cumm_length = path_until_threshold(points, frac_length)
     fraction = 1. - (cumm_length - frac_length) / point_dist(p0, p1)
-    return two_points_fraction(p0, p1, fraction)
+    return linear_interpolate(p0, p1, fraction)
 
 
 def scalar_projection(v1, v2):
--- a/neurom/analysis/morphmath.py
+++ b/neurom/analysis/morphmath.py
@@ -56,6 +56,32 @@ def linear_interpolate(p1, p2, fraction):
                      p1[2] + fraction * (p2[2] - p1[2])))
 
 
+def interpolate_radius(r1, r2, fraction):
+    '''Calculate the radius that corresponds to a point P that lies at a fraction of the length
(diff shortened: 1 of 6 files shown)


neurom/analysis/morphmath.py
```diff
@@ -47,6 +47,49 @@ def vector(p1, p2):
     return np.subtract(p1[0:3], p2[0:3])
 
 
+def two_points_fraction(p1, p2, fraction):
```
juanchopanza: How about "interpolate"?

neurom/sections.py
```diff
@@ -92,6 +92,20 @@ def area(section):
...
+    of the path length along the piecewise linear curve which
+    is constructed from the section points.
+
+    Args:
+        0, 1, 2 correspoding to 3D cartesian coordinates
+
+    Returns:
+        The 3D coordinates of the aforementioned point
+    '''
+    return mm.path_fraction_point(tuple(n.value[:3] for n in section), fraction)
```
juanchopanza: Do you really need to slide here? I think you can just pass the points as values.

neurom/segments.py
```diff
@@ -73,6 +73,21 @@ def identity(segment):
...
+    '''Computes the point p between two points p1, p2
+    for which |p1p| = fraction * |p1p2|
+
+    Args:
+        fraction: float between 0. and 1.
+
+    Returns:
+        The 3D coordinates of the aforementioned point
+    '''
+    return mm.two_points_fraction(segment[0].value[:3],
```
juanchopanza: Not sure you need to slice here.

neurom/analysis/morphmath.py
eleftherioszisis: Hmm, I think interpolate too general. Interpolation can be linear, polynomial,etc.

linear_interpolate maybe?

lidakanari:
What about the diameters - radii? Do we need an interpolation there as well?

neurom/segments.py
eleftherioszisis: You' re right. Lemme see if removing the slicing is ok for section above as well.

neurom/analysis/morphmath.py
juanchopanza: That's fine.

eleftherioszisis:
This is an interesting point. @dkeller9, would it be useful if the interpolated diameter is returned as well?

Also it needs a different name if it returns the radius as well. Any ideas?

eleftherioszisis:
Ok, I added the segment radius interpolation along with the 3D point. I renamed the segment function ```cross_section_at_fraction```

juanchopanza:
@eleftherioszisis  Looks OK to me. I can merge if you're ready to go.

eleftherioszisis:
I don't have anything else to add. Feel free to merge it!

PR #286, opened 2016-03-21
fix: local bifurcation angle calculation for zero length segments

### LINKED ISSUE #285: How to calculate a bifurcation angle when one or both children are zero distance from their parent
The bifurcation angle is defined as the angle between the lines connecting a parent point to its two child points in a morphology. We explicitly exclude multifurcations beyond two. Many biological cells have artifacts in which one or both children are at zero distance from their parent. We do not want to reject these points in calculating the bifurcation angles as they often constitute a large subset of any neuron's bifurcation points.

A simple solution is to make a check when calculating a bifurcation angle:

if( both children are greater than epsilon from their parent)
    calculate bifurcation angle using the two vectors from parent to the children
else if( one or both children are less than epsilon from their parent)
    replace the child of zero length with its own child if it is unique; if one child is itself a bifurcation ignore it.

    if( both these children exist)
       calculate the bifurcation angle using these children/grandchildren
    else
       // one child must be a termination
       do not count this point as a bifurcation

juanchopanza:
Algorithm now searches for first point in child section(s) such
that the bifurcation start - end points aren't the same.

Fixes #285.

### DIFF
--- a/neurom/analysis/morphtree.py
+++ b/neurom/analysis/morphtree.py
@@ -51,10 +51,22 @@ def path_length(tree):
 def local_bifurcation_angle(bifurcation_point):
     '''Return the opening angle between two out-going segments
     in a bifurcation point
+
+    The bifurcation angle is defined as the angle between the first non-zero
+    length segments of a bifurcation point.
     '''
+    def _skip_0_length(p, c):
+        '''Return the first child c with non-zero distance to parent p'''
+        while np.all(p.value[:COLS.R] == c.value[:COLS.R])\
+                and not tr.is_leaf(c) and not tr.is_forking_point(c):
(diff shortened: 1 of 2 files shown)


juanchopanza:
@berthelm @lidakanari this should fix the issue when points daughter points have exactly the same coordinates as the parent. Could someone check if the zeros disappear?

lidakanari:
OK, I will run it with bio cells and confirm the status of the output.

juanchopanza:
I will merge this. Then if there are still problems, I may add a point equality tolerance parameter.

lidakanari:
I think all is good now! Problem fixed:
figure_1

Issue #288, opened 2016-04-04
Feature to extract coordinates of all segments of a neurite


lidakanari:
This feature is needed for the density plots and to simplify the computation of more complex measurements. It should be something like:

NEURITEFEATURES["segment_x_coordinates"] should return a list of all the x coordinates of a structure as the mean (x_parent + x_child)/2 of each segment in this structure

eleftherioszisis:
In case you are interested, you can use something similar to the func below:

```python 
def segment_center(seg, projections):
    if projections in (0, 1, 2):
         return (seg[0][projections] + seg[1][projections]) / 2.
    else:
        s = slice(min(projections), max(projections) + 1)
        return (seg[0][s] + seg[1][s]) / 2
```

and then call it:

```
segment_center(seg, COLS.X)
segment_center(seg, (COLS.X, COLS.Y, COLS.Z))
etc
```

I just leave it here. :P

lidakanari:
Maybe this is optimal for the intermediate steps but I would like to avoid complications at the feature exposed to the user. So in the NEURITEFEATURES I would prefer to have the simplest implementation. In any case I'll let @juanchopanza decide which way is best.

PR #289, opened 2016-04-04
Add segment x, y, z coordinates and radii features

lidakanari:
This solves #288 and it is crucial for the simplification of plotting and constructing more complex measurements.


eleftherioszisis:
Seems fine to me!

juanchopanza:
I think it would be much simpler to calculate the "position" of a segment as the 3D mean of its two points, and *maybe* add helper functions to extract X, Y, and Z. Having three separate functions everywhere seems like too much. For example

    def segment_position(seg):
        return (seg[0][:COLS.R] + seg[1][1:COLS.R]) / 2.

Then, if you really want separate "features" for each coordinate, then it is three loops.

lidakanari:
For what I need those functions, the three coordinates need to be separated. I don't want to do all the calculations each time I only need the projection in x for example, that's why I disconnect the coordinates.

juanchopanza:
OK. It looks like a lot of code for something really simple. I will merge it and try to reduce it later.

lidakanari:
Yes, that's True. I just want to avoid expensive calculations in intermediate steps that will not be used in the final step. If you have a better alternative please go ahead and simplify it.

PR #290, opened 2016-04-04
Example to generate density plot

lidakanari:
This is the first attempt to migrate density plots from Pneumatk. Two main issues: 
- Functionality is not optimized so the density plot takes forever for VERY big datasets
- I need too many variables in order to control all the plotting options. As a result, I need to locally disable pylint errors. 

Please feel free to propose solutions if you have any easy fixes.


lidakanari:
@liesbethvanherpe This should be useful to automatically generate the plots you need :)

PR #291, opened 2016-04-10
Add total length measurements in neuritefeatures

lidakanari:


neurom/features/neurite_features.py
```diff
@@ -109,6 +110,19 @@ def number_of_sections_per_neurite(neurites, neurite_type=TreeType.all):
     return (_sec.count(n) for n in neurites if _ttc(neurite_type)(n))
 
 
+@as_neurite_list
+def total_length_per_neurite(neurites, neurite_type=TreeType.all):
+    '''Get an iterable with the total length of a neurite for a given neurite type'''
+    return (sum(_sec.length(ss) for ss in isection(n))
```
juanchopanza: I really don't think we need to be writing all these one-liners as functions. The point of the design is that you can write the one liner if and when you need it. You could just have this code in the feature map.

neurom/features/neurite_features.py
```diff
@@ -109,6 +110,19 @@ def number_of_sections_per_neurite(neurites, neurite_type=TreeType.all):
...
+            for n in neurites if _ttc(neurite_type)(n))
+
+
+@as_neurite_list
+def total_length(neurites, neurite_type=TreeType.all):
+    '''Get an iterable with the total length of a neuron for a given neurite type'''
+    yield sum(total_length_per_neurite(neurites, neurite_type=neurite_type))
```
juanchopanza: Same here.

neurom/features/neurite_features.py
juanchopanza: Although I realise it might be a bit awkward because python isn't very good with anonymous functions, and lambdas are quite limited. So if what I said above is problematic, let me know and I will try to see if it can be simplified.

eleftherioszisis: @juanchopanza I think you underestimate the complexity of the code needed in order to do these one liners.

For example a user must be familiar with the iterator design of NeuroM, the sections module, the type checker etc. I would put a bet that nobody who is not familiar with NeuroM, can write this one liner without spending at least one day to understand the structure of NeuroM.

Thus, I don't see  the harm in having these one liners accessible by users who have more important things to work on that spending their time understanding NeuroM code.

I should have put the total length functions myself when I created the features module, but I somehow forgot and thus asked Leda to do it because she needed this functionality anyway.

juanchopanza: I thought that was what the feature map was for.

eleftherioszisis: Oh, do you mean that instead of having a function for this code, to create an "alias" as with the functions in the beginning of the neurite_features file?

juanchopanza: Yes, something like that. Although, as I said, I realise that might be a bit awkward in python. But it might be an idea to try it that way first, and if it is too complicated or the one-liner becomes too complex, then we can use the functions (and I can see if I can simplify the design.)

lidakanari: Ok, that makes sense. I'll try to simplify it tomorrow.

juanchopanza: OK great! But please don't spend too much time on it. If it proves to be to complicated, just let me know.

lidakanari: Sure, no problem!

PR #293, opened 2016-04-12
Neurite feature tests test map of features.

juanchopanza:
* The map of features is the entry point for users of the feature
  functionality. Any helper functions it uses can be seen as an
  implementation detail, tested indirectly via the features map.
* Feature map accessed through helper function get_feature.
* Remove make_iterable functionality.


juanchopanza:
@eleftherioszisis @lidakanari I made further simplifications and added a helper `get_feature` function. I also removed some generality that isn't needed at the moment. I think it can be cleaned further.

lidakanari:
Looks good to me. We are currently only testing get_feature functionality, is that right?

eleftherioszisis:
Looks good to me as well!

PR #297, opened 2016-04-13
Cleanups and minor improvements.

juanchopanza:
* TreeType renamed to NeuriteType. Alias ezy.TreeType kept for
  backwards compatibility.
* neurom.geatures.get_feature added to neurom.ezy module.
* Documentation examples updated.
* Examples updated.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -143,17 +143,17 @@ Examples
 
 .. code-block:: python
 
-    >>> from neurom import ezy
-    >>> nrn = ezy.load_neuron('some/data/path/morph_file.swc')
-    >>> apical_seg_lengths = nrn.get_segment_lengths(ezy.TreeType.apical_dendrite)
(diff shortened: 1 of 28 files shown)


juanchopanza:
This affects many files, but most changes are trivial, i.e. changing `TreeType` for `NeuriteType`. Note that `ezy.TreeType` is kept for backwards compatibility, and is an alias for `NeuriteType`.

The next step will be to remove or deprecate `ezy.neuron` functions whose functionality is already available from `neurom.features.get_feature`.

eleftherioszisis:
Awesome! Is it possible that we have neurites instead of trees  for the neurons etc. and make a little more evident the underlying type injection that takes place in the making of a tree?

juanchopanza:
Yes, I think so. We can discuss tomorrow or Friday.

lidakanari:
Seems good to me!

PR #298, opened 2016-04-14
Test ezy.neuron.Neuron vs. features.get_feature compatibility.

juanchopanza:
Need to check get_features is equivalend to ezy Neuron member
functions functionality before deprecating and removing ezy
Neuron members.


juanchopanza:
@eleftherioszisis @lidakanari First step towards removing duplicated functionality: make sure it really is duplicated!

PR #300, opened 2016-04-14
Shorten neurom.features.get_feature

juanchopanza:
* Shortened to neurom.features.get
* Also available as neurom.ezy.get


juanchopanza:
@eleftherioszisis @lidakanari More fun with less typing.

PR #302, opened 2016-04-19
Iteration and path length optimizations.

juanchopanza:
* Convert points, segments, sections, bifurcations and triplets
  to value representation manually instead of calling as_elements.
* Use numpy.sqrt instead of numpy.linalg.norm to calculate distance
  between points.
* Add segment length squared function.
* Re-implement morphtree.path_length with a loop instead of using
  sum over complex iterator expression.

### DIFF
--- a/neurom/analysis/morphtree.py
+++ b/neurom/analysis/morphtree.py
@@ -122,8 +122,7 @@ def find_tree_type(tree):
 
     tree_types = tuple(NeuriteType)
 
-    types = [node[COLS.TYPE] for node in tr.val_iter(tr.ipreorder(tree))]
-    types = [node[COLS.TYPE] for node in tr.val_iter(tr.ipreorder(tree))]
+    types = np.array([node.value[COLS.TYPE] for node in tr.ipreorder(tree)])
 
     return tree_types[int(np.median(types))]
 
--- a/neurom/analysis/morphtree.py
+++ b/neurom/analysis/morphtree.py
@@ -44,8 +44,13 @@ import numpy as np
 
 def path_length(tree):
     '''Get the path length from a sub-tree to the root node'''
-    return np.sum(s for s in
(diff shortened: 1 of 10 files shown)


juanchopanza:
@eleftherioszisis This should improve the situation with the path lengths calculation. Could you check your profiling of `get_feature('section_path_distances', n)` using these changes?

neurom/analysis/morphtree.py
```diff
@@ -44,8 +44,13 @@
 
 def path_length(tree):
```
juanchopanza: I tried a recursive solution, which could greatly benefit from memoization, but the maximum recursion depth was reached. Maybe this thing can be expressed recursively in terms of sections instead of segments.

juanchopanza:
The timing of `get_feature('section_path_distances', n)` seems to drop from about 4m45s to 1m45s on my desktop. So, not orders of magnitude, but it is a start. Ideally this should take less than 1s. I think it is doable.

neurom/analysis/morphtree.py
eleftherioszisis: Recursion is a joke in python... Even with sections you will have to put a guard decorator or sth similar to check if you reached the extremely low recursion limit, in order to extend it, which sucks... -.- If you find however a nice way to do fast recursion, let me know. I would love to try it. :P

eleftherioszisis:
These are some awesome changes. Thanks for the time that you put on this!

neurom/analysis/morphtree.py
eleftherioszisis: Btw, we have to be careful with the handling of the memory as well because when I did the loading with the 1000 synthesized cells from Guy (I think that they had a relatively small number of segments), they took up 13 gb of ram. Maybe it would be a good approach to do memoization with limited capacity?

lidakanari:
Indeed! Great improvement :+1:

PR #303, opened 2016-04-19
Small optimizations to remove_duplicate_points.

juanchopanza:
* Simplify _find_last_point helper function.
* Use plain python array for to_be_reduced counters. Using numpy
  array seems to cause some conversion overheads when incrementing
  the contents and using in a zip iteration.

### DIFF
--- a/neurom/io/hdf5.py
+++ b/neurom/io/hdf5.py
@@ -184,17 +184,16 @@ class H5(object):
         '''
 
         group_initial_ids = groups[:, 0]
+        group_len = len(group_initial_ids)
 
         def _find_last_point(group_id):
             ''' Identifies and returns the id of the last point of a group'''
+            return group_initial_ids[group_id + 1] - 1
 
-            if group_id != len(group_initial_ids) - 1:
-                return group_initial_ids[np.where(group_initial_ids ==
-                                                  groups[group_id][0])[0][0] + 1] - 1
-
-        to_be_reduced = np.zeros(len(groups))
+        to_be_reduced = [0] * group_len
         to_be_removed = []
 
+        # This is the slow part
         for ig, g in enumerate(groups):
             if g[2] != -1 and np.allclose(points[g[0]],
                                           points[_find_last_point(g[2])]):
@@ -202,11 +201,12 @@ class H5(object):
                 to_be_removed.append(g[0])
                 # Reduce the id of the following sections
                 # in groups structure by one
-                for igg in xrange(ig + 1, len(groups)):
-                    to_be_reduced[igg] = to_be_reduced[igg] + 1
+                for igg in xrange(ig + 1, group_len):
+                    to_be_reduced[igg] += 1
 
         groups = np.array([np.subtract(i, [j, 0, 0])
                            for i, j in itertools.izip(groups, to_be_reduced)])
+
         points = np.delete(points, to_be_removed, axis=0)
 
         return points, groups
--- a/neurom/io/hdf5.py
+++ b/neurom/io/hdf5.py
@@ -144,6 +144,8 @@ class H5(object):
     def unpack_data(points, groups):
         '''Unpack data from h5 data groups into internal format'''
 
+        group_ids = groups[:, _H5STRUCT.GPFIRST]
+
         @memoize
         def find_group(point_id):
             '''Find the structure group a points id belongs to
@@ -151,7 +153,7 @@ class H5(object):
             Return: group or section point_id belongs to. Last group if
                     point_id out of bounds.
             '''
-            bs = np.searchsorted(groups[:, _H5STRUCT.GPFIRST], point_id, side='right')
+            bs = np.searchsorted(group_ids, point_id, side='right')
             bs = max(bs - 1, 0)
             return groups[bs]
 
@@ -166,7 +168,7 @@ class H5(object):
                 # parent is last point in parent group
                 parent_group_id = group[_H5STRUCT.GPID]
                 # get last point in parent group
-                return groups[parent_group_id + 1][_H5STRUCT.GPFIRST] - 1
+                return group_ids[parent_group_id + 1] - 1
 
         return np.array([(p[_H5STRUCT.PX], p[_H5STRUCT.PY], p[_H5STRUCT.PZ], p[_H5STRUCT.PD] / 2.,
                           find_group(i)[_H5STRUCT.GTYPE], i,
@@ -190,7 +192,7 @@ class H5(object):
             ''' Identifies and returns the id of the last point of a group'''
             return group_initial_ids[group_id + 1] - 1
 
-        to_be_reduced = [0] * group_len
+        to_be_reduced = np.zeros(group_len)
         to_be_removed = []
 
         # This is the slow part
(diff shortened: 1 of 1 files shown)


neurom/io/hdf5.py
```diff
@@ -184,29 +184,29 @@ def remove_duplicate_points(points, groups):
...
         def _find_last_point(group_id):
             ''' Identifies and returns the id of the last point of a group'''
+            return group_initial_ids[group_id + 1] - 1
 
-            if group_id != len(group_initial_ids) - 1:
-                return group_initial_ids[np.where(group_initial_ids ==
-                                                  groups[group_id][0])[0][0] + 1] - 1
-
-        to_be_reduced = np.zeros(len(groups))
+        to_be_reduced = [0] * group_len
```
juanchopanza: This was huge.

neurom/io/hdf5.py
```diff
@@ -184,29 +184,29 @@ def remove_duplicate_points(points, groups):
         '''
 
         group_initial_ids = groups[:, 0]
+        group_len = len(group_initial_ids)
 
         def _find_last_point(group_id):
             ''' Identifies and returns the id of the last point of a group'''
+            return group_initial_ids[group_id + 1] - 1
```
juanchopanza: Could somebody check whether this is OK? All the tests pass, and I couldn't figure out why the previous version was so complicated, so I am worried I missed something.

neurom/io/hdf5.py
lidakanari: Wow! I didn't know that... np.zeros is that slow?

neurom/io/hdf5.py
lidakanari: I think there is a problem here:

group_initial_ids[group_id + 1] implies that the ordering between sections is correctly increasing. This is indeed True for the example I created and most of the trivial trees we generate. I would like to generate some more complicated trees to add to the testing. 

However, we should check if the sections are expected to be ordered:
So section_of_branch_order1 is directly followed by section_of_branch_order2. In this case, I think the modification is accurate.

neurom/io/hdf5.py
juanchopanza: It is slow when used like this. Transforming between the numpy world and the pure python world has an overhead.

neurom/io/hdf5.py
juanchopanza: According to my reading of the spec, the group IDs are supposed to be ordered like this. This doesn't translate to the branch order relationship you suggest, because the parent ordering isn't fixed. So `group_initial_ids[group_id + 1]` is not a child of `group_initial_ids[group_id]`, but its index is one more than the last point in `group_initial_ids[group_id]`.

lidakanari: OK. Yes, I am aware of the ordering of IDs, but wasn't aware if they follow an ordering of sections. Here is an example of what would not work for the previous function:

sec_initial_segment_id, type, parent:
1, 2, 0 -> sec1
100, 2, 3 -> sec2
10, 2, 1 -> sec3

in this case the previous function will fail to compute the last segment id of sec1 correctly. It would return 99 instead of 9.

juanchopanza: I think that wouldn't satisfy the spec. It would need to be
```
1, 2, 0 -> sec1
10, 2, 1 -> sec3
100, 2, 3 -> sec2
```

lidakanari: Note that the first number is the groups initial segment, so not the section's ID in the groups!

juanchopanza: Maybe @mgeplf can comment on the h5 spec issue.

juanchopanza: @lidakanari it is actually the ID of the first *point* in the section. BTW, here is the relevant part of the v1 spec (File Format / `structure`):

> **start offset**: The implicit index in the `points` dataset of the first point of the section. _The points in this section go until the next Start offset defined by the next row in the `structure` dataset_, except for the last section in the file, then it runs until the end of the points dataset.

(emphasis mine)

I don't see anything in the v2 spec to deviate from this. But it would be great if more people could check this.

neurom/io/hdf5.py
```diff
@@ -184,29 +186,28 @@ def remove_duplicate_points(points, groups):
...
 
+        # This is the slow part
         for ig, g in enumerate(groups):
             if g[2] != -1 and np.allclose(points[g[0]],
                                           points[_find_last_point(g[2])]):
                 # Remove duplicate from list of points
                 to_be_removed.append(g[0])
                 # Reduce the id of the following sections
                 # in groups structure by one
-                for igg in xrange(ig + 1, len(groups)):
```
juanchopanza: Vectorializing this loop was did the trick.

neurom/io/hdf5.py
juanchopanza: By the way, this structure is assumed in the data unpacking, so if it isn't guaranteed we have bigger problems!

mgeplf: I'm not aware of any official 'spec' on ordering of branches.  I suspect that there are places where the expectations is that they match the original neurolucida ascii, and that if you write the file back, they are in the same order.

neurom/io/hdf5.py
mgeplf: I don't think it's np.zeros that is slow:
```
In [7]: %timeit  [0] * 100000
1000 loops, best of 3: 366 µs per loop

In [8]: %timeit np.zeros(100000)
10000 loops, best of 3: 28.8 µs per loop
```

neurom/io/hdf5.py
lidakanari: @juanchopanza Thanks for confirming this issue. I'm making the same assumption, so it's good to verify this!
@mgeplf If what Juan states is correct, I think we don't require more ordering assumptions. In Nascii format this ordering is important because we don't have ids of parents.

mgeplf: IIRC, neurolucida has a tree structure encoded in the format, so that's why they don't need parent ids.

eleftherioszisis:
I now realized that I had made changes in this function to make it faster but I forgot to do a pull request! @juanchopanza I  put here the code, in case you would like to use it. I use masks instead of resizing lists:

``` python
    @staticmethod
    def remove_duplicate_points(points, groups):
        ''' Removes the duplicate points from the beginning of a section,
        if they are present in points-groups representation.

        Returns:
            points, groups with unique points.

        '''

        def _find_last_point(group_id, groups):
            ''' Identifies and returns the id of the last point of a group'''
            group_initial_ids = np.sort(np.transpose(groups)[0])

            if group_id != len(group_initial_ids) - 1:
                return group_initial_ids[np.where(group_initial_ids ==
                                                  groups[group_id][0])[0][0] + 1] - 1

        to_be_reduced = np.zeros(len(groups))

        # mask of the points that will be returned. Starts as true, i.e. all points
        # will be returned and if a duplicate is found, the respective index will
        # become false
        p_mask = np.ones(len(points), dtype=np.bool)

        for ig, g in enumerate(groups):
            if g[2] != -1 and np.allclose(points[g[0]],
                                          points[_find_last_point(g[2], groups)]):
                # Set duplicate's index to False in order to remove
                # it from the resulting array
                p_mask[g[0]] = False
                # Reduce the id of the following sections
                # in groups structure by one
                to_be_reduced[ig + 1: len(groups)] += 1

        groups = np.array([np.subtract(i, [j, 0, 0])
                           for i, j in itertools.izip(groups, to_be_reduced)])

        return points[p_mask], groups
```

Is there a way to remove a commit? :P

juanchopanza:
@eleftherioszisis I tried it out and it didn't really change anything. Probably because it is a large array, because of the loop, and because most of the entries are true. There may be a better way of building the mask, or leveraging numpy to do the removal. But the "bottleneck" in unpacking is now `find_group`, responsible for about 1/4 of the time spent (so, not even a bottleneck, just one of the dominant components.)

juanchopanza:
For removing a commit, [see here](http://stackoverflow.com/questions/927358/how-do-you-undo-the-last-commit)

PR #305, opened 2016-04-20
Improve performance of H5.unpack_data.

juanchopanza:
* Pre-load point ID -> group ID map to avoid repeated binary
  searches.
* Remove memoization (ID map takes care of caching now.)


juanchopanza:
Unless there are any objections, I will merge this one.

eleftherioszisis:
I don't have any :P

PR #310, opened 2016-04-20
Speed-ups in raw data checkers.

juanchopanza:
More use of numpy operations and izip vs. zip.

### DIFF
--- a/neurom/io/check.py
+++ b/neurom/io/check.py
@@ -40,7 +40,7 @@ def has_sequential_ids(raw_data):
     with their predecessor)
     '''
     ids = raw_data.get_col(COLS.ID)
-    steps = [int(j) for (i, j) in zip(ids, ids[1:]) if int(j - i) != 1]
+    steps = ids[np.where(np.diff(ids) != 1)[0] + 1].astype(int)
     return len(steps) == 0, steps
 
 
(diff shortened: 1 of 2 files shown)


neurom/io/check.py
```diff
@@ -79,7 +77,7 @@ def has_increasing_ids(raw_data):
     with their predecessor)
     '''
     ids = raw_data.get_col(COLS.ID)
-    steps = [int(j) for (i, j) in zip(ids, ids[1:]) if i >= j]
+    steps = ids[np.where(np.diff(ids) <= 0)[0] + 1].astype(int)
```
eleftherioszisis: Is the dtype conversion in the end necessary? If the steps are used by numpy arrays, I think it does not matter if the are floats as long as they do not have decimals.

juanchopanza: Well, I was returning ints before, and they are supposed to be IDs. That is the only reason I did the conversion. I may look at using one of those mixed type numpy arrays, but I am worried it will slow things down.

lidakanari:
I think that they should be ints, as @juanchopanza has them. If they are indices they are expected to be ints, otherwise numpy throws an error. See similar issue here: 
http://stackoverflow.com/questions/18994787/python-svmlight-error-deprecationwarning-using-a-non-integer-number-instead-of

neurom/io/check.py
eleftherioszisis: Yes, my bad on this. They should be converted into ints or be ints from the beginning. I've never used the mixed arrays, thus I've no idea of the overhead.

juanchopanza:
OK to merge this?

lidakanari:
OK for me!

PR #311, opened 2016-04-20
Small optimizaitons in h5 unpack_data.

juanchopanza:
Minimize group ID look-ups and avoid simple
function call overhead.


juanchopanza:
Is this OK?

PR #312, opened 2016-04-21
Furhter optimizations to hdf5 unpack_data.

juanchopanza:
Vectorize more operations in data block construction.


juanchopanza:
After these changes, `io.hdf5.unpack_data` only takes 5% of time in `neurom.io.utils.load_data` and 1.5% of time in `neurom.io.utils.load_neuron. 

I think it is time to optimize other things. There isn't much more to gain here.

eleftherioszisis:
Awesome. This is great!

PR #313, opened 2016-04-25
Simplify h5 duplicate removal.

juanchopanza:
* Following tightening in HDF5 format specification, duplicate removal can be simplified to always remove first point of non-root, non soma sections.
* Small improvement in duplicate removal implementation by using vectorization.


juanchopanza:
@lidakanari @eleftherioszisis more speed, less code. I think this is as far as the H5 unpacking can go. The next thing will be to optimize the tree building, maybe changing to a tree of sections instead of points.

PR #317, opened 2016-04-28
Section tree module

juanchopanza:
Section tree has soma and has been moved to its own module.

### DIFF
--- a/examples/section_tree.py
+++ b/examples/section_tree.py
@@ -1,126 +1,21 @@
 '''Build a section tree'''
 
-from collections import defaultdict
-from collections import namedtuple
 import sys
-import numpy as np
-from neurom.io.hdf5 import H5
-from neurom.core.tree import Tree, ipreorder
-from neurom.core.dataformat import POINT_TYPE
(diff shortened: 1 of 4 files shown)


juanchopanza:
@eleftherioszisis @lidakanari I have implemented many "features" using the section tree, but the functions currently only work for a single neuron. If either of you have the time to try this out it woudl be great. I think I covered all the features used in the m30 synthesis validation.

eleftherioszisis:
Do you want to test something in particular or just load and run some analysis?

juanchopanza:
@eleftherioszisis If you have something you can run it would be good to see how much faster it is. BTW I _think_ the feature functions would work with populations too, at least for neurite features.

eleftherioszisis:
I can run it on the same population of 1000 cells that I posted before and see what happens :P

eleftherioszisis:
(Although I don't have a comparison for analysis apart from the big cell)

juanchopanza:
OK, sounds good. You may need to write a `load_population` function calling `section_neuron.load_neuron`. Maybe try a smaller sample first!

PR #331, opened 2016-05-10
Re-factor and extend section unpacking code.

juanchopanza:
* Enable section tree neurons from SWC data. This is
  easily extensible to NeuroLucida ASCII.
* Remove redundant section ID field.
* Use slice objects to represent section point ID ranges in h5 data.
* Use id lists to represent section point ID ranges in non-h5 data.


juanchopanza:
Close and re-open in an attempt to trigger travis

PR #337, opened 2016-05-12
Add principal_direction_extents to _fst features.

juanchopanza:


juanchopanza:
@eleftherioszisis Since you added the principal direction extent feature, do you have any comments on whether the duplicated points at section boundaries can make a difference or explain the discrepancies seen between this code and the point-tree one?

neurom/_fst/mm.py
```diff
@@ -273,3 +273,19 @@ def bifurcation_partition(bif_point):
     n = float(sum(1 for _ in ipreorder(bif_point.children[0])))
     m = float(sum(1 for _ in ipreorder(bif_point.children[1])))
     return max(n, m) / min(n, m)
+
+
+def principal_direction_extents(nrn, neurite_type=NeuriteType.all, direction=0):
+    '''Principal direction extent of neurites in neurons'''
+    def _pde(neurite):
```
eleftherioszisis: I spent one minute wondering why this is called "Partial Differential Equation" XD

eleftherioszisis:
Hmm, I am unsure on how the calculation of the pca will be affected from the duplication of the section boundaries. 

I think that it would stir the covariance ellipsoid towards the section points due to the fact that these points are encountered twice on the same position. In other words the spatial distribution of the point cloud is changed due to the duplicate points.

If duplicate points are artificially introduced in the test trees, do the results change a lot?

juanchopanza:
That makes sense. I guess the order of the points doesn't matter, right? I ask because I can also remove the duplicates by skipping the first point of each section, but then I would have to add back the initial neurite trunk points. If the order doesn't matter this should be quite straight forward.

eleftherioszisis:
If by order you mean the way you parse the tree, then yes it doesn't matter because the function uses a collection of points, not the tree itself.

Wouldn't it have been better if the duplicate points are removed in the loading of the neuron? I find it kind of inefficient having to remove them each time one needs to parse the tree and it may possibly lead to strange bugs due to the fact that the tree will be parsed regardless of the existence of duplicate points. I don't know exactly the mechanics of the section neurons, thus this is more of a question than a suggestion.

juanchopanza:
It is fixed by removing the duplicate points. Regarding removing them on loading, the sections need these points (I actually have to add them when parsing SWC data.)

PR #345, opened 2016-05-18
Section tree compatibility code for viewers

juanchopanza:


juanchopanza:
@lidakanari @eleftherioszisis FYI, with these changes the new-style neurons can be viewed.

PR #352, opened 2016-05-23
Use ubuntu 14.04 (trusty tahir) beta travis CI

juanchopanza:


juanchopanza:
@lidakanari FYI, this could help with #350.

PR #361, opened 2016-05-26
Update some examples to use fst module

juanchopanza:
Note: synthesis_json.py no longer produces segment taper rate stats.


examples/synthesis_json.py
```diff
@@ -116,12 +106,13 @@ def transform_package(mtype, files, components):
        features, feature_names, feature_components, feature_min, feature_max
     '''
     data_dict = transform_header(mtype)
+    neurons = load_neurons(files)
 
     for comp in components:
         params = PARAM_MAP[comp.name]
         for feature in comp.features:
             result = stats.fit_results_to_dict(
-                extract_data(files, feature.name, params),
```
juanchopanza: This was loading all the files for each feature of each component.

Issue #362, opened 2016-06-02
neurolucida: No handlers could be found for logger


juanchopanza:
`neurolucida` tries to write to a logger, and when there isn't one, this message appears:

> No handlers could be found for logger "neurom.io.neurolucida"

To reproduce:

``` python
from neurom import fst
nrn = fst.load_neuron('test_data/neurolucida/bio_neuron-001.asc')
```

lidakanari:
I don't think that this issue is fixed! Here's the last git pull I did:

```
commit fe2a5c8ac40dfb791e09731cc9589b8136dd3cba
Merge: 2603c7f 8b89302
Author: Juan Palacios <juan.palacios.puyana@gmail.com>
Date:   Thu Jun 2 23:06:26 2016 +0200

    Merge pull request #363 from juanchopanza/master
    
    Add neurom.log module with default stream logger
```

But the problem persists:
```python
from neurom import fst

nrn = fst.load_neuron('test_data/neurolucida/bio_neuron-001.asc')

neurom/io/neurolucida.py:254: UserWarning: This is an experimental reader. There are no guarantees regarding ability to parse Neurolucida .asc files or correctness of output.
  warnings.warn(msg)
No handlers could be found for logger "neurom.io.neurolucida"

```

juanchopanza:
You need to import the `neurom.log` module to get the handler.

```python
from neurom import log
from neurom import fst
nrn = fst.load_neuron('test_data/neurolucida/bio_neuron-001.asc')
```

But if you don't do that, at least now you get a warning telling you not to trust the reader :-)

eleftherioszisis:
I have a question. If the ```log``` is not imported, the file will be loaded correctly?

I am asking because a user will not know that he has to load the log, which he is not going to use, in order for the ```load_neuron``` function to properly work.

juanchopanza:
What file? `load_neuron` behaves as "correctly" as it can for `.asc` files, that is to say, kind of maybe a little bit. The issue is that the user wasn't being warned before, and now they are.

eleftherioszisis:
Ok, I was wondering if there was a difference in loading process or possible error that could be generated by not loading the logger.

wvangeit:
I can still reproduce this on the latest version of NeuroM on PyPi.

wizmer:
What can you reproduce exactly ? Simply opening a neuron does not trigger any message for me.

wvangeit:
[...]

PR #363, opened 2016-06-02
Add neurom.log module with default stream logger

juanchopanza:


juanchopanza:
Fixes #362.

neurom/log.py
```diff
@@ -0,0 +1,50 @@
...
+
+* Sets up a default stream handler with shor tformat string and level DEBUG.
+
+* Defines short and long format strings FMT_LONG and FMT_LONG respectively.
+'''
+import logging
+import logging.handlers
+
+# root level logger with stream handler.
+LOG = logging.getLogger()
```
mgeplf: I would find this behavior quite surprising.  Playing w/ the root logger means that when someone else changes it, importing neurom could change the behavior, right?

juanchopanza: Yes, good point. I should get a named logger, e.g.

    LOGGER_NAME = 'neurom'
    LOG = logging.getLogger(LOGGER_NAME)

mgeplf: Yeah, that helps.  I would probably also have a helper to remove it.

Issue #365, opened 2016-06-03
Name of cells is not saved in fst module


lidakanari:
Here is an example. This problem has a significant impact on the analysis, since we cannot track back the cells that are analyzed.

```
from neurom import ezy
from neurom import fst

n = ezy.load_neuron('./test_data/neurolucida/sample.asc')
n1 = fst.load_neuron('./test_data/neurolucida/sample.asc')

n.name
Out: 'sample'

n1.name
AttributeError: 'Neuron' object has no attribute 'name'
```

The same applies for neurons in a population. We should be able to identify which neuron is analyzed, even when we study a whole population of them.

juanchopanza:
Thanks for reporting this. I will fix it next week.

juanchopanza:
Implemented in #367.

PR #367, opened 2016-06-06
Set neuron name when loading with fst module.

juanchopanza:
Follows same logic of neurom.io point neuron loader.
Implements #365.


lidakanari:
That's great. Thanks!

Issue #364, opened 2016-06-02
Change definition of "number_of_sections"  of populations in get (of both ezy and fst modules)


lidakanari:
The "number_of_sections" should return a list of the total number of 
sections per *neuron*. Currently it adds up all the sections of the population. 
Here is an example:

```python
from neurom import fst
pop = fst.load_population('./test_data/valid_set/')
fst.get('number_of_sections', pop)

Out: array([336])
```

The number of neurons in the population is 5:
```python
len(pop.neurons)
Out: 5

```
So the correct result would be:

```python
Out: array([42],  [42],  [84],  [84], [84])
```

juanchopanza:
That probably makes more sense. Whatever it does, it should be documented clearly. i.e. the behaviour here would be different for populations and single neurons. So we'd need to document what it does in either case.

juanchopanza:
Maybe it would be better to specify clearly what each of the "feature" functions should return, both for single neurons and for populations.

lidakanari:
Yes, sorry.. I though it was common sense!

The feature of "number_of_sections" should return:
- one number which is the number of all the sections of the neuron, if the input is a single neuron
- a list which contains one number **per** neuron, which is the number of all the sections of the neuron, if the input is a population

The same goes for the "total_length" or other total measurements. 
- one number which is the sum of all the section lengths of the neuron, if the input is a single neuron
- a list which contains one number **per** neuron, which is the sum of all the section lengths of the neuron, if the input is a population

Is this a little bit more clear?

juanchopanza:
@lidakanari That is fine, thanks. @eleftherioszisis I think you implemented the "feture getters", right? If so, did you have a reason to implement the current functionality, or is it OK to change it?

lidakanari:
@eleftherioszisis Also, I am not sure if it should be a list or a np.array (please confirm so that we make all the features consistent )!

eleftherioszisis:
@juanchopanza Yes, indeed I implemented them. I didn't have any particular reasons for the total numbers due to not knowing what was needed. I believe that the per neuron numbers are much more useful for analysis, and in fact I needed per neuron total numbers as well for astrocytes, which I did manually. :P

All the features I implemented in the features module returned iterators that were converted into numpy arrays.

juanchopanza:
OK, thanks for the information. I can start working on this next week.

juanchopanza:
BTW, shouldn't the output be

```python
array([42,  42,  84,  84, 84])
[...]

PR #368, opened 2016-06-06
Change definition of some neurom.fst.get features.

juanchopanza:
For some scalar quantities, return list of values per neuron instead
of aggregation over all neurons in a population. Features affected:

* total_length
* number_of_sections
* number_of_segments
* number_of_neurites

Implements part of #364.

### DIFF
--- a/neurom/fst/tests/test_get_features.py
+++ b/neurom/fst/tests/test_get_features.py
@@ -1,0 +1,204 @@
+# Copyright (c) 2015, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
+# All rights reserved.
+#
+# This file is part of NeuroM <https://github.com/BlueBrain/NeuroM>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#     1. Redistributions of source code must retain the above copyright
+#        notice, this list of conditions and the following disclaimer.
+#     2. Redistributions in binary form must reproduce the above copyright
+#        notice, this list of conditions and the following disclaimer in the
+#        documentation and/or other materials provided with the distribution.
+#     3. Neither the name of the copyright holder nor the names of
+#        its contributors may be used to endorse or promote products
+#        derived from this software without specific prior written permission.
+#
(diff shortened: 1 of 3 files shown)


juanchopanza:
@lidakanari @eleftherioszisis Does this do "the right thing"?

neurom/fst/tests/test_get_features.py
```diff
@@ -0,0 +1,204 @@
...
+    nt.assert_items_equal(fst.get(feat, POP), [84, 42, 202])
+
+    nt.assert_items_equal(fst.get(feat, POP, neurite_type=NeuriteType.all),
+                          [84, 42, 202])
+
+    nt.assert_items_equal(fst.get(feat, POP, neurite_type=NeuriteType.axon),
+                          [21, 21, 179])
+
+    nt.assert_items_equal(fst.get(feat, POP, neurite_type=NeuriteType.apical_dendrite),
+                          [21, 0, 0])
```
lidakanari: Here is an interesting point... Do we want to return zeros for the types of the tree that are missing? In my opinion it makes sense, but I think it is important to clearly document this. 
Why is this crucial? Because if we have a population of 100 cells where 1 of them has an apical and we compare it to a population of 10 cells among which 1 has an apical we might get very weird results! Also, in the past (NeuroM previous versions and Pneumatk) that was not the case, so this might alter results because of a different definition.

lidakanari:
This seems to resolve the #364 issue as far as fst is concerned.

neurom/fst/tests/test_get_features.py
juanchopanza: I think if we're returning numbers *per neuron*, it makes sense to return `0` where applicable, because that is the number of things for a given neuron for a given neurite type. I think not to do this would be quite confusing.

eleftherioszisis: I agree that for the per neuron results, 0 should be put when applicable. This way it is also possible to get the index of the neuron that has the respective value as the number of elements in the array and the number of neurons are the same.

lidakanari: OK, I agree with both of you. Then we should clearly document that so that the user knows what to expect!

PR #369, opened 2016-06-06
Add segment_radii feature to neurom.fst.get.

juanchopanza:
Added regression and neurom.ezy.get compatibility tests.


juanchopanza:
@lidakanari This might be of interest. It isn't the most efficient implementation, but it is usable.

lidakanari:
Thanks!

PR #372, opened 2016-06-08
Iterable population

juanchopanza:
`Population` class extended to provide list-like iteration over neurons, `len` and `[]` access.
The `load_population` function has been merged into `load_neurons` and removed.
A deprecated alias is kept in the `neurom.ezy` module for backwards compatibility.


lidakanari:
Why did we decide to deprecate load_population and not load_neurons. As far as I am concerned I have only been using load_population and not load_neurons. @chalimou that might also concern you.

neurom/core/population.py
```diff
@@ -33,15 +33,33 @@
...
 
         Arguments:
-            neurons: iterable of neuron objects (core or ezy) .
+            neurons: iterable of neuron objects.
             name: Optional name for this Population.
         '''
-        self.neurons = neurons
-        self.somata = [neu.soma for neu in neurons]
-        self.neurites = list(chain.from_iterable((neu.neurites for neu in neurons)))
+        self.neurons = tuple(neurons)
```
mgeplf: why tuple?

mgeplf:
seems fine to me

neurom/core/population.py
juanchopanza: I want it to be defined at initialization, as far as python allows that. So now at least people can't add elements to any of the fields.

juanchopanza:
@lidakanari It is a somewhat arbitrary choice. To me `load_population`requires explaining what "population" means in this context. Since it is a just glorified collection of neurons, `load_neurons` seemed like the simpler choice. If this is a problem I can leave a deprecated version in the `fst` module too.

neurom/core/population.py
mgeplf: Ok, thought so, just wanted to make sure that was the intent.

lidakanari:
No, I don't think it is very important. It is just biologically counter - intuitive. Neurons seems like a list of cells, while we actually want  (and for my understanding that's what we have) a collection of neurons on which we can perform operations as they were one entity.

juanchopanza:
I can swap it around, and keep `load_population`, and deprecate `load_neurons`, if you think that is clearer.

lidakanari:
To me it is a bit more clear to have a population, but if there is a strong preference for neurons, I wouldn't mind using either. In any case, I think that for clarity it would be useful to document exactly what we mean by either "neurons" or "population".

juanchopanza:
OK, I kept `load_neurons`, but added a `load_population` symbol to the `fst` module with a deprecation warning.

lidakanari:
OK, we can check with other users which one is better and eventually keep the one that most of the ppl are using. Thanks!

lidakanari:
@romani79 This might be interesting for you.

PR #373, opened 2016-06-09
Add neurom.viewer module with helper draw function.

juanchopanza:
Draws all kinds of morphologies in 2d, 3d, and dendrogram modes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,5 +154,8 @@ Examples
 .. code-block:: python
 
     >>> # Initialize nrn as above
-    >>> fig, ax = ezy.view(nrn)
+    >>> from neurom import viewer
+    >>> fig, ax = viewer.draw(nrn)
+    >>> fig.show()
+    >>> fig, ax = viewer.draw(nrn, mode='3d') # valid modes '2d', '3d', 'dendrogram'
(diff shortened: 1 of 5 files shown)


juanchopanza:
I am wondering whether it is confusing to have `neurom.viewer` and `neurom.view`. Maybe everything can be put under one single module, maintaining the top level functionality shown here. The question would then be, which name? `viewer` or `view`? @lidakanari @eleftherioszisis any thoughts?

lidakanari:
I agree that it is confusing to have both. I guess we should be consistent with the general code: do we prefer verbs or nouns? Also, we should consider the view.draw and viewer.draw which will be available to the user. I guess the second one sounds a little bit better, but this is just a matter of personal taste...

juanchopanza:
I was thinking the same. `view.draw` sounds like two verbs (even if both words can also be nouns.) How about adding this as it is right now, and then moving the stuff from `neurom.view` into  `neurom.viewer` module at a later stage?

lidakanari:
No problem for me.

juanchopanza:
OK, maybe we can discuss naming tomorrow.

PR #382, opened 2016-06-16
Add section tortuosity feature to fst module.

### LINKED ISSUE #375: New feature: tortuosity
It would be useful for validations to add the section_tortuosity feature at the "get" level for the next release. 

Definition of tortuosity can be found here (http://www.ncbi.nlm.nih.gov/pmc/articles/PMC3928584/):

"Also, the distance metric (DM) tortuosity (a dimensionless number) provides a ratio between the actual path length `(correction: this should be section length)` of a meandering curve (i.e. along segments) and the linear distance between endpoints (radial distance)"

Neuroanatomical Tract-Tracing: Molecules, Neurons, and Systems, by Laszlo Zaborszky, Floris G. Wouterlood, José Luis Lanciego

> The dendritic tortuosity gives a ratio between the actual length of a dendritic segment (our section) and the distance between its endpoints

juanchopanza:
Resolves #375.

### DIFF
--- a/neurom/fst/_mm.py
+++ b/neurom/fst/_mm.py
@@ -101,11 +101,6 @@ def map_sections(fun, nrn, neurite_type=NeuriteType.all):
                 for s in i_chain2(nrn.neurites, tree_filter=is_type(neurite_type)))
 
 
-def section_lengths(nrn, neurite_type=NeuriteType.all):
-    '''section lengths'''
-    return map_sections(mm.path_distance, nrn, neurite_type)
-
-
 def section_branch_orders(nrn, neurite_type=NeuriteType.all):
     '''section lengths'''
     return [branch_order(s)
@@ -123,7 +118,7 @@ def section_path_lengths(nrn, neurite_type=NeuriteType.all):
     dist = {}
 
     for s in i_chain2(nrn.neurites, tree_filter=is_type(neurite_type)):
-        dist[s] = mm.path_distance(s.value)
+        dist[s] = mm.section_length(s.value)
 
     def pl2(sec):
         '''Calculate the path length using cahced section lengths'''
@@ -211,7 +206,7 @@ def section_radial_distances(nrn, neurite_type=NeuriteType.all, origin=None):
 def trunk_section_lengths(nrn, neurite_type=NeuriteType.all):
     '''Get a list of the lengths of trunk sections of neurites in a neuron'''
     tree_filter = is_type(neurite_type)
-    return [mm.path_distance(s.value) for s in nrn.neurites if tree_filter(s)]
+    return [mm.section_length(s.value) for s in nrn.neurites if tree_filter(s)]
 
 
 def trunk_origin_radii(nrn, neurite_type=NeuriteType.all):
@@ -281,7 +276,7 @@ def total_length_per_neurite(nrn, neurite_type=NeuriteType.all):
 
 def section_path_length(section):
     '''Path length from section to root'''
-    return sum(mm.path_distance(s.value) for s in iupstream(section))
+    return sum(mm.section_length(s.value) for s in iupstream(section))
 
 
 def map_sum_segments(fun, section):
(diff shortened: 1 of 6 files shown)


neurom/fst/_mm.py
```diff
@@ -101,11 +101,6 @@ def map_sections(fun, nrn, neurite_type=NeuriteType.all):
                 for s in i_chain2(nrn.neurites, tree_filter=is_type(neurite_type)))
 
 
-def section_lengths(nrn, neurite_type=NeuriteType.all):
-    '''section lengths'''
-    return map_sections(mm.path_distance, nrn, neurite_type)
```
lidakanari: This is a bit confusing, or maybe I don't understand it correctly. The path_distance is not suppose to return the same as the path_distance. The first is the sum(length of all segments in a section), the second is sum(length of all segments from current section to the root of the tree).

juanchopanza:
@lidakanari Does this look OK?

lidakanari:
See my previous comment:
I got a bit confused between section_length and path_distance definitions.

neurom/fst/_mm.py
juanchopanza: mm.path_distance is the sum of distances between the points in a section, i.e. the sum of lengths of the segments. It wasn't my choice of naming!

lidakanari: OK, that is a better name for the section_length (=sum(length of segments in a section) indeed. However, that is not what path distance should be. The path distance is supposed to be the sum of all the segment lengths from a section to the root (so the section-length plus all its ancestors section-lengths).

lidakanari: Give me a while to confirm some definitions, because this could be a misunderstanding of the term path_distance in general...

juanchopanza: Are you saying the tortuosity calculation is wrong? I went by the definition you gave in #375. Regarding the path distance to the root, there is another function for that. The definition of `mm.path_length` hasn't changed since the beginning of `NeuroM`.

lidakanari: Sorry, I think that you are right and this should be done at each section's level (in which case it should not be called path_distance but section_length as you correctly indicated). However, I fail to find any reasonable reference of this definition. We can assume that it is the same as here: http://cng.gmu.edu:8080/Lm/help/index.htm and I will try to find another reference for confirmation.

juanchopanza: The function that calculates the cummulative segment length to the root is `morphtree.path_length`.

lidakanari: OK, you are definitely right. Here is a more concrete definition. 

*Neuroanatomical Tract-Tracing: Molecules, Neurons, and Systems*,  by Laszlo Zaborszky, Floris G. Wouterlood, José Luis Lanciego 

> The dendritic tortuosity gives a ratio between the actual length of a dendritic segment (our section) and the distance between its endpoints

lidakanari: Yes, this is irrelevant to tortuosity. I think the function you implemented is doing exactly what is expected. The discussion about the ```path_length``` / ```path_distance``` / ```section_length``` can be moved somewhere else (because this is still confusing, but has nothing to do with the actual functionality).

neurom/fst/_mm.py
```diff
@@ -299,6 +294,17 @@ def section_area(section):
     return map_sum_segments(mm.segment_area, section)
 
 
+def section_tortuosity(section):
+    '''Tortuosity of a section
+
+    The tortuosity is defined as the ratio of the path length of a section
+    and the euclidian distnce between its end points.
+
```
lidakanari: So, maybe we should update the doc-string: 

The tortuosity is defined as the ratio between the actual length of a  section and the distance between its end points.

neurom/fst/_mm.py
juanchopanza: Yes, we should discuss these names at some point because I also get confused. For fst, I want to have separate modules for functions that apply to an individual section, so that should help clarify things.

PR #396, opened 2016-06-24
Add most basic fst functionality to top level module

juanchopanza:
Symbols imported from neurom.fst into neurom:
* load_neuron
* load_neurons
* get
* NeuriteType
* NEURITE_TYPES

Added aliases for NeuriteType values:

* APICAL_DENDRITE
* BASAL_DENDRITE
* AXON
* SOMA
* ANY_NEURITE

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -123,10 +123,10 @@ Examples
 
 .. code-block:: python
 
-    >>> from neurom import fst
-    >>> nrn = fst.load_neuron('some/data/path/morph_file.swc')
-    >>> apical_seg_lengths = fst.get('segment_lengths', nrn, neurite_type=fst.NeuriteType.apical_dendrite)
-    >>> axon_sec_lengths = fst.get('section_lengths', nrn, neurite_type=fst.NeuriteType.axon)
+    >>> import neurom as nm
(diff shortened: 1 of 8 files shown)


juanchopanza:
Note: these changes are backwards compatible. Code that worked before will continue to work unchanged.

eleftherioszisis:
Is the```get``` function going to be called from ```neurom``` and not from ```neurom.features```?

juanchopanza:
@eleftherioszisis Yes, the `get` function is available directly as `neurom.get`. BTW it isn't the same one as `neurom.features.get`, which uses the point tree representation. It is pulled from `neurom.fst`. At some point, there will be a clean-up of the point tree stuff.

eleftherioszisis:
@juanchopanza Yep, I know that it's the ```fst``` one. I see, so when the point tree is cleaned, will the ```fst.get``` be called from ```neurom.features```, or it will stay in the top level?

I am wondering if when everything is cleaned that we keep the features, etc. file organization (with the ```fst``` functionality). This is a question out of curiosity for the future organization of NeuroM.

juanchopanza:
@eleftherioszisis I'm not sure about the future sum-module structure, but it can certainly do with some simplification. We can discuss this at a future stage.

Issue #338, opened 2016-05-13
Minor Issue with Soma Type


eleftherioszisis:
In the ```SingleSomaContour``` in ```neurom/core/neuron.py``` there is a small inconsistency where the center is returned as a list, while in the other types as an array. Armando discovered this while trying to load cells of different soma types.

``` python
return list(np.mean(points[:, :COLS.R], axis=0))
```
@juanchopanza Can you change this into a numpy array in a next commit? Thanks!

juanchopanza:
Specifically what code had an issue with this? Could you post a self contained, minimal example that reproduces this "issue" (that should be standard when reporting this type of issues.)

eleftherioszisis:
I think it is evident that in one case the center will be returned as a list and in the other as a numpy array.

You can try loading the hippocampal neurons that have a contour-like soma and one from test_data.

juanchopanza:
No, you need to provide a working example showing this is an issue.

eleftherioszisis:
I don't understand. You have the same member variable (center) returned as a different type of data structure depending on the type of the neuron soma during the loading.

If you want for example to take the mean of the somata positions and some of the neurons represent them as lists while others as numpy arrays, then this will fail.

juanchopanza:
You're reporting an issue, so you need to provide the example that shows there is a issue. I can't assume that returning different types is an issue without proof, and there is nothing in the documentation that specifies that the center is represented as a numpy array anyway.

eleftherioszisis:
The example is the code itself. You have the ```BaseSoma``` that returns the center as a numpy array, the ```SomaSinglePoint``` and the ```SomaThreePoint``` that inherit from it and thus return numpy arrays as well.

Then you have the ```SomaSimpleContour``` that inherits from ```BaseSoma```, overrides the center method and returns a list instead (which can just be removed as it converts the numpy array that is returned from the np.mean to a list).

juanchopanza:
Yes, but you have failed to show why that is a problem. If you don't have a test that fails because of this, then there is no issue.

eleftherioszisis:
Nevermind. Apparently it's not as trivial as I thought.

I will fix this and provide with the respective "proof" when I have the time. There is no point behind further arguing about this.

juanchopanza:
[...]

PR #398, opened 2016-06-28
Add points properties to Soma and fst.Neuron types.

juanchopanza:
* returns 2D array with (x, y, z, r) points of a Neuron and a Some
  respectively.
* Add neurom.geom.bounding_box, which returns the (x, y, z)
  bounding box of an object with a points property as described above.

Note: the bounding_box functions makes no assumptions about the
      meaning of a collection of points. It just calculates the
      minimum and maximum (x, y, z) coordinates.

### DIFF
--- a/neurom/_compat.py
+++ b/neurom/_compat.py
@@ -51,6 +51,8 @@ def is_new_style(obj):
 def bounding_box(neurite):
     '''Get a neurite's X,Y,Z bounding box'''
     if is_new_style(neurite):
+        if isinstance(neurite, Tree):
+            neurite = fst.Neurite(neurite)
         return geom.bounding_box(neurite)
     else:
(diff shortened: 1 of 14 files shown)


juanchopanza:
This also resolves #338, by arbitrarily choosing to consistantly return numpy arrays for the soma centres instead of lists. There are no tests for this because that issue is not well specified.

juanchopanza:
@lidakanari, @eleftherioszisis This change should be largely transparent to the end-user, but there is one difference: neuron and population `neurites` fields now hold `Neurite` objects, not `Trees`. So user code that explicitly treats `nrn.neurites` as `Tree`s may need to be modified. All `fst` and `viewer` code has been modified to cope with this transparently.

PR #400, opened 2016-06-30
Use fst neurons in morphology checking code.

juanchopanza:
All checkers in neurom/check have been migrated to check
fst neurons and neurites.


juanchopanza:
Users of `morph_check` may see different behaviour in certain types of badly formed SWC files: errors that could be pin-pointed before now result in a failure to read the file. The outcome is the same, but there may be less information regarding the specific error. It may be possible to recover this functionality if needed via extensions to the raw data checking module and application.

Issue #279, opened 2016-03-07
Axes scaling when plotting morphology


wvangeit:
Based on feedback from @jdcourcol on the NeuroM figure generated here:
https://github.com/BlueBrain/BluePyOpt/blob/master/examples/l5pc/L5PC.ipynb
we were wondering if it wouldn't make more sense if NeuroM would keep the aspect ratio when plotting morphologies. The morphology looks a bit unnatural with the current default plotting.
(maybe this could be an option ?)

lidakanari:
Could you please define "keep the aspect ratio"? Do you need this to be X-range/Y-range in morphology to be equal to the range of the axes on the plot? 

Currently, you can define this ratio by choosing the xlim - ylim. However, we could change the default values in order to be consistent with what is expected. Due to lack of specs for this the decision for the axes ratio was arbitrary, so I wouldn't mind making it consistent with a spec if that is well defined :)

jdcourcol:
We would like to have the scale to be the same in both axes. So that the shape of the neuron is not transformed. The resulting image does not need to be a square.

wvangeit:
And regarding 'specs'. Don't think we need this for this function, would rather make default what most people are expecting. Personally I think that most people would want to see a morphology in the same shape has if looking at a slice, but of course I'm not sure. That's why I was suggesting you could add it as an option (one way or the other).

lidakanari:
I agree with you Werner. There is no standard. So in the lack of a standard I just left the figure's axes to be defined a bit arbitrary. The option to choose them is to define xlim and ylim. So if you need something specific select the appropriate xlim, ylim. 
In order to define a standard, I would like a intuition of what would be better:  do we need to have xlim == ylim? That will make the plot square. Do we prefer to have the cell as it would look in a slice (that is usually not square), so xlim=horizontal range of cell, ylim= vertical range of cell.

wvangeit:
There is no standard, and I really don't think we need one either: https://xkcd.com/927/
The only thing we're proposing is that the most used case is probably that 'it would look like in a slice', so it could be the default. Maybe it shouldn't, we don't know for sure, it's ok to close the ticket. 
People can change the xlim, ylim every time they plot a morphology, but an option (e.g. keep_aspect_ratio, stretch etc ?) would be much more handy.

jdcourcol:
thank you Lida. I will give it a try as soon as a package is available.

PR #408, opened 2016-07-04
Cleaning view code and adding aspect ration option, as requested in  #279

lidakanari:
Choose aspect_ratio = 'equal' to set the aspect ratio of the figure.

### DIFF
--- a/neurom/view/common.py
+++ b/neurom/view/common.py
@@ -173,7 +173,7 @@ PLOT_STYLE_PARAMS = '''        pretitle (Optional[str]): String to include befor
             according to matplotlib default options. \
             For equal aspect ratio select 'equal'. \
             For normal aspect ration select 'auto'. \
-            Default value is set to 'auto'.
+            Default value is set to 'equal'.
 
     Returns:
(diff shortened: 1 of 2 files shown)


lidakanari:
Check this @wvangeit, @chalimou, @romani79, @eleftherioszisis :

Please let me know which value I should set to default if you have any preference. Examples of 'normal' versus 'equal' aspect_ratio:

normal

equal

lidakanari:
This fixes  #279

chalimou:
I’m not an expert but I think equal ratio as default makes more sense,

nancy

juanchopanza:
@lidakanari Looks OK to me. Let me know whenever this is ready to be merged.

lidakanari:
OK, I will change the default value and I'll let you merge it.

lidakanari:
It should be good now, if there are no errors in CI.

wvangeit:
Yep, equal aspect ratio made most sense indeed

romani79:
I absolutely agree with Werner

lidakanari:
OK, done :) Let me know if it works for you!

PR #409, opened 2016-07-04
Fix unclosed figures in viewer.draw tests

lidakanari:
The tests of viewer.draw were generating too many figures, because they were not closed during the testing process. I made a quick fix for this to avoid the warnings of "too many opened figures" during the tests.

### DIFF
--- a/neurom/tests/test_viewer.py
+++ b/neurom/tests/test_viewer.py
@@ -32,7 +32,7 @@ from neurom.fst import load_neuron as load_fst_neuron
 from neurom import viewer
 from neurom.analysis.morphtree import set_tree_type
 import os
-from matplotlib import pylab as plt
+from matplotlib import pyplot as plt
 
 
 class Dummy(object):
--- a/neurom/tests/test_viewer.py
+++ b/neurom/tests/test_viewer.py
@@ -32,7 +32,7 @@ from neurom.fst import load_neuron as load_fst_neuron
 from neurom import viewer
 from neurom.analysis.morphtree import set_tree_type
 import os
-import pylab as plt
+from matplotlib import pylab as plt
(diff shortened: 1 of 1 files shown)


neurom/tests/test_viewer.py
```diff
@@ -32,6 +32,7 @@
 from neurom import viewer
 from neurom.analysis.morphtree import set_tree_type
 import os
+import pylab as plt
```
juanchopanza: Does it work if you import pyplot from matplotlib instead?

lidakanari: ```from matplotlib import pylab as plt``` ?

Sure, should I replace it with this instead?

juanchopanza: No, I mean `from matplotlib import pyplot as plt`.

lidakanari:
@juanchopanza Changed it to this. Let me know if that's OK.

neurom/tests/test_viewer.py
lidakanari: Sorry, yes it works for this as well.

neurom/tests/test_viewer.py
```diff
@@ -49,26 +50,31 @@ class Dummy(object):
...
+    plt.close('all')
 
 
 def test_draw_soma():
     viewer.draw(nrn.soma)
     viewer.draw(fst_nrn.soma)
+    plt.close('all')
 
 
 def test_draw_soma3d():
```
juanchopanza: Just wondering, does this function also need a `plt.close('all')`?

lidakanari: Yes, of course. I missed it. Thanks!

PR #415, opened 2016-07-05
View code was failing when specific limits were selected. 

lidakanari:
This fixes the limits error.


lidakanari:
@chalimou That should fix the xlim, ylim error.

Issue #417, opened 2016-07-06
Cannot translate neuron in viewer.draw - produces error


chalimou:
I used to translate synthesized neurons, so that the soma has (0,0,0) coordinates. I include an example synthesized neuron. After updating to "viewer" following the deprecating message the  function "translate" produces an error. I used the following script:

```
from neurom import ezy
from neurom.geom.transform import translate
import numpy
if __name__ == '__main__':
    n = ezy.load_neuron('../test_data/L23_PC_ID50.h5')
    m = translate(n, - numpy.asarray(n.soma.center))
    f1,a1=ezy.view(m, diameter=False)
```

Afterwards, I updated to fst and viewer like this:

```
from neurom import fst
from neurom import viewer
from neurom.geom.transform import translate
import numpy
if __name__ == '__main__':
    n = fst.load_neuron('../test_data/L23_PC_ID50.h5')
    m = translate(n, - numpy.asarray(n.soma.center))
    f1,a1=viewer.draw(m, diameter=False)
```

It produces the following error:

```
      7     n = fst.load_neuron('../test_data/L23_PC_ID50.h5')
      8     m = translate(n, - numpy.asarray(n.soma.center))
----> 9     f1,a1=viewer.draw(m, diameter=False)

/Users/chalimou/Projects/NEUROM/NeuroM/neurom/viewer.pyc in draw(obj, mode, **kwargs)
    117         tag = 'soma'
    118     else:
--> 119         raise NotDrawableError('draw not implemented for %s', obj.__class__)
    120
    121     viewer = '%s_%s' % (tag, mode)
```

[L23_PC_ID50.h5.zip](https://github.com/BlueBrain/NeuroM/files/350219/L23_PC_ID50.h5.zip)

juanchopanza:
`neurom.geom.translate` is not implemented for new-style neurons, and unfortunately the implementation does not tell you this, letting the error propagate to the viewer. I am working on an implementation of translate at the moment. I hope to have an up-date soon.

juanchopanza:
Fixed in #418.

PR #418, opened 2016-07-07
Generic 3D transformations

juanchopanza:
Implemented generic 3D transformation types and added transform methods
to fst.Neuron and fst.Neurite.

Transformations implemented in neurom.geom.transform:

* pivot rotation
* rotation
* translation

fst.Neuron and fst.Neurite have transform methods that accept a
generic transformation object and return a transformed copy of
themselves. For example

```
nrn = load_neuron(....)
m = Translation([1,2,3])
nrn2 = m(nrn)
```

The neurom.geom.transform entry points use these transformations behind the scenes, but the interface remains the same:

```
from neurom.geom import transform as tr
nrn = load_neuron(...)
nrn2 = tr.translate(nrn, [1, 2, 3])
nrn3 = tr.rotate(nrn, [0, 0, 1], math.pi / 4.0)
```

Note: these transformations are not implemented for the deprecated
point neurite representation of morphologies. These have been removed
from neurom.geom.transform.

Re-factoring and clean-up related to efficient cloning of neurons and
neurites has been performed.

### DIFF
--- a/neurom/check/morphology.py
+++ b/neurom/check/morphology.py
@@ -37,7 +37,6 @@ from neurom.analysis.morphmath import section_length
 from neurom.analysis.morphmath import segment_length
 from neurom.check.morphtree import is_flat, is_monotonic, is_back_tracking
 from neurom.exceptions import SomaError
-from neurom.io import check as io_check
 from neurom.fst import _mm as fst_mm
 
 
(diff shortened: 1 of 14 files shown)


juanchopanza:
This resolves #417.

PR #420, opened 2016-07-11
Split fst._mm into submodules.

juanchopanza:
Create separate sum-modules for functions that can be called on

* neuriites and collections of neurites
* sections

### DIFF
--- a/examples/section_tree.py
+++ b/examples/section_tree.py
@@ -2,7 +2,8 @@
 # pylint: disable=protected-access
 
 import sys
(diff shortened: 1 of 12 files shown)


juanchopanza:
@lidakanari @eleftherioszisis @chalimou Just a heads up. This may cause problems if you are using the `fst._mm` module directly. If so, let me know because the fix would be trivial. But note that modules starting with `_` are "internal", so once the migration to section-trees is complete, user code should not be relying on these modules.

PR #450, opened 2016-07-28
Re-vamp morph_stat to make more configurable.

### LINKED ISSUE #449: Make morph_stats configurable
The `morph_stats` script should be extended to accept a configuration script. If we restrict the morphometrics to those accessible via `neurom.get`, is would be straight forward to configure based on required feature and neurite type where applicable.

Reasonable default feature set, with possibility of splitting by neurite type:

* neurite length
* maximum section length
* maximum branch order
* neurite volume
* soma diameter

juanchopanza:
* Uses features from neurom.get and summary options such as 'max', 'min',
  'raw', 'total'.
* Option for writing summary json file. See for example
  apps/config/morph_stats.yaml.
* Moved apps/config.yaml to apps/config/morph_check.yaml.

Resolves #449.


juanchopanza:
FYI @jdcourcol @mgeplf

Issue #272, opened 2016-02-19
z-jump check


eleftherioszisis:
Another check that is important for morphologies, is the z-jump check.

In neurolucida, the reconstruction is of course in 3 dimensions but the user can only see in his/her screen a 2D slice of the 3D space. In order to navigate in the z-axis he/she has to roll the mouse scroll button. The z-jump takes place when the reconstructor thinks that he/she is at the correct height while tracking the neurite, but in reality he/she has gone deeper. Here's an example:
![z_jump](https://cloud.githubusercontent.com/assets/15011025/13175580/9a320190-d70b-11e5-95bd-a0b837b742d9.png)
This usually manifests on bifurcations where the spatial transition from parent to children might be tricky, but it can also happen on other places as it can be seen in the image. 

Currently, there is an algorithm in the repair tool to detect the z-jumps, thus we can either check this one first and either modify it to our needs or create another one which will be suitable for our data structures.

juanchopanza:
@eleftherioszisis Does #461 cover this?

juanchopanza:
I will assume it is covered.

PR #461, opened 2016-08-10
added z-jump check

mgeplf:
* Follows BBP policy of what a z-jump is:
    Large movements in the Z axis between two points


mgeplf:
need

neurom/check/neuron_checks.py
```diff
@@ -171,3 +171,25 @@ def has_nonzero_soma_radius(neuron, threshold=0.0):
...
+        neuron: Neuron object whose neurites will be tested
+        max_z_distance: value above which consecutive z-values are
+        considered a jump
+
+    Return:
+        status and list of ids bad sections
+    '''
+    bad_ids = []
+    for s in _nf.iter_sections(neuron):
+        it = izip(s.points, s.points[1:])
```
lidakanari: Thanks Mike! Do we need to detect jumps between any consecutive points of beginning and end of a segment? I think it should be the second, which from my understanding is not what is done here, but my interpretation of this error is a bit blurry.

mgeplf: Isn't a segment defined by two points?

lidakanari: Now I see the difference: those are two points within a section, right? It's OK then, I'm saying the same thing.

apps/morph_check
```diff
@@ -84,13 +84,15 @@ Default errors checked for
...
 * Zero length sections
+* Jumps in the Z-direction
 
 Default values for options
 --------------
 * has_nonzero_soma_radius 0.0
 * has_all_nonzero_neurite_radii: 0.007
 * has_all_nonzero_segment_lengths: 0.01
 * has_all_nonzero_section_lengths: 0.01
+* has_no_zjumps: 30 micrometer - distance considered a jump
```
juanchopanza: Do we really want this switched on by default? Especially given that it arbitrarily picks one axis, so is not very general. We could let users enable this in their config file.

mgeplf: As we discussed, my plan is to:
1) Turn this off by default
2) allow one to pick the 'axis' - not everyone does reconstructions with slice in z axis, however, it should always (?) be one of x, y, z

juanchopanza: That sounds fine to me!

neurom/check/runner.py
```diff
@@ -126,13 +126,14 @@ def _check_file(self, f):
...
         try:
             data = load_data(f)
-            result &= self._check_loop(data, 'structural_checks')
-            nrn = fst_core.FstNeuron(data)
-            result &= self._check_loop(nrn, 'neuron_checks')
         except StandardError as e:
             L.error('Failed to load data... skipping tests for this file')
             L.error(e.message)
-            result = False
+            return False, {f: OrderedDict(('ALL', False))}
```
mgeplf: I wrote my config.yaml wrong, but it was error out saying it was b/c the file couldn't load, and that was confusing.

neurom/check/runner.py
```diff
@@ -126,12 +126,17 @@ def _check_file(self, f):
...
+            return False, {f: OrderedDict([('ALL', False)])}
+
+        try:
             result &= self._check_loop(data, 'structural_checks')
             nrn = fst_core.FstNeuron(data)
             result &= self._check_loop(nrn, 'neuron_checks')
         except StandardError as e:
-            L.error('Failed to load data... skipping tests for this file')
-            L.error(e.message)
+            L.error('Check failed: %s: %s', type(e), e.message)
```
mgeplf: not sure happy w/ this, but it's the previous behavior.

mgeplf:
does this satisfy #272?

PR #473, opened 2016-08-17
Add tests for contour soma.

### LINKED ISSUE #472: Add tests for SWC contour soma reading
Add simple test and test data to ensure that a "contour soma" SWC is parsed correctly.

juanchopanza:
A contour soma is one where the neurites branch out of some points
other than its end-points.

Resolves #472.

### DIFF
--- a/neurom/io/tests/test_swc_reader.py
+++ b/neurom/io/tests/test_swc_reader.py
@@ -75,6 +75,24 @@ def test_read_split_soma():
         nt.eq_(s.ids, r)
 
 
+def test_read_contour_soma_neuron():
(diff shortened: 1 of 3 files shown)


lidakanari:
Ok, that's exactly what we need! Should we also have an example that fails to load: i.e., when an actual bifurcation is present in the soma? Something simple as the following will do:

+1 1 0 0 0 0.0 -1
+2 1 0 0 2 0.0 1
+3 1 0 0 3 0.0 2
+4 1 0 0 4 0.0 2
+5 1 0 0 5 0.0 4
+6 1 0 0 6 0.0 5

juanchopanza:
Yes, we could do that, as a separate task. Could you add that as an issue?

lidakanari:
OK, I added issue #475

PR #479, opened 2016-08-29
loading SWC file with bifurcating soma raises.

### LINKED ISSUE #475: Add test to ensure no bifurcations occur in the soma
The following example of a morphology file (swc) should fail to load since two soma points have the same soma parent:

+1 1 0 0 0 0.0 -1
+2 1 0 0 2 0.0 1
+3 1 0 0 3 0.0 2
+4 1 0 0 4 0.0 2
+5 1 0 0 5 0.0 4
+6 1 0 0 6 0.0 5

juanchopanza:
Somas that bifurcate are considered an error. A check
is applied when a neuron is constructed from a raw data block,
and is invoked implicitly when loading neurons via neurom.load_neuron or
neurom.load_neurons.

Note: make_soma does not apply this check, because it works on a
collection of points and has no knowledge of whether they originate
from an SWC file.

Note: Unfortunately this error occurs in most of the SWC test files.
This required updating many SWC reader regression tests.

Note: Some h5 test files, converted from SWC ones, had issues with soma
sections. These have been re-generated from the corresponding new SWC
files. As a result many regression tests had to be modified.

Resolves #475.


juanchopanza:
@lidakanari This implements the test for issue #475. Most of the changes are due to the original test files being wrong!

lidakanari:
True! I checked the file changes and they seem OK to me. I only have one question about the changes in the features.

neurom/fst/tests/test_get_features.py
```diff
@@ -99,26 +99,26 @@ def test_section_tortuosity_pop():
     nt.ok_(np.allclose(_stats(fst.get(feat, POP)),
                        (1.0,
                         4.6571118550276704,
-                        440.38275799536723,
-                        1.3426303597419733)))
+                        440.40884450374455,
+                        1.3427098917797089)))
```
lidakanari: Why are features like tortuosity modified?

neurom/fst/tests/test_get_features.py
```diff
@@ -321,39 +321,39 @@ def test_total_length_pop():
     feat = 'total_length'
 
     nt.ok_(np.allclose(fst.get(feat, POP),
-                       [840.47744821, 418.83424432, 13250.82577394]))
+                       [840.68522362, 418.83424432, 13250.82577394]))
```
lidakanari: Or total lengths? Do they depend on the soma points or there was a mistake in previous versions?

juanchopanza: I see a discrepancy in the Z coordinate of the second point in each section between the old and new h5 files. This could be an artefact of the conversion. I will look into it.

lidakanari: Yes, that would explain it. So this is related to the conversion to h5. Yes, it is outside the scope of this pull request. However, it is nice to document why this changed and to keep it in mind for future reference. Thanks!

juanchopanza: @lidakanari After further checks, I see that the defect is in h5 files converted from the original, invalid SWC file. The new ones look OK. I am adding some regression tests, see #480.

lidakanari: OK, that makes sense. We cannot do much about the converter. Great idea! It would be very helpful to have those tests so that we can detects errors as soon as possible.

PR #485, opened 2016-09-01
Iter neurites fix

juanchopanza:
Update neurom.core.iter_neurites.
    
The existing implementation was only useful for the deprecated
neurom.point_neurite code. It has been moved to
neurom.point_neurite.core. A new, simplified version has been
added back to neuro.core and is imported directly in neurom.

### DIFF
--- a/examples/comparison.py
+++ b/examples/comparison.py
@@ -29,7 +29,7 @@
 '''Module for the comparison of the morphometrics between two sets of trees.'''
 
 import numpy as np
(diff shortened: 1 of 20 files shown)


juanchopanza:
@mgeplf Here it is. The relevant files are `neurom/core/__init__.py` and `neurom/core/tests/test_core.py`.

mgeplf:
I diggit, looks good to me

PR #486, opened 2016-09-02
Fix inconsistency between soma definitions in docs and code.

juanchopanza:


juanchopanza:
@lidakanari as per previous discussion.

PR #487, opened 2016-09-02
Add file format documentation

juanchopanza:
Minimal, list of formats with some links and comments.


juanchopanza:
@lidakanari FYI I made a file format doc page. It is quite brief but we can refine in the future.

lidakanari:
That is super useful! Sure, I will update it once I have more information.

PR #488, opened 2016-09-02
Simplify fst._neuritefunc

juanchopanza:
Use re-worked neurom.core.iter_neurites to make simplifications.

Notes:
* can be further simplified. Many functions are one-liners which
  could be replaced by functools.partial.
* Further re-factoring could be done. Need to make sure clarity isn't
  lost.


juanchopanza:
@mgeplf Some simplifications. Can be improved further.

mgeplf:
Late to the party, but it looks good

Issue #438, opened 2016-07-14
section_path_distances gives an error for the hippocampus cell


chalimou:
The extraction of the feature section_path_distances for the hippocampus cell (included below) crashes with an error when computed with the following script: (this was not the case with ezy)

```
import neurom as nm

if __name__ == '__main__':
    ref_population = nm.load_neurons('./hippo')
    ref_data = nm.get('section_path_distances', ref_population, neurite_type=nm.AXON)
```
[hippo.zip](https://github.com/BlueBrain/NeuroM/files/363706/hippo.zip)

juanchopanza:
I will look into it, but note that this neuron has two zero-length sections. This could be the root of the problem.

```
morph_check hippo/ca3b-N2.CNG.swc -d
```

Relevant part of output:

```
DEBUG: Nonzero section lengths: 2 failing ids detected: [1, 366]
ERROR:             Nonzero section lengths FAIL
```

Note I removed unnecessary stuff from your example.

juanchopanza:
I think I found a potential issue: this file has two of its three soma points somewhere in the middle (SWC IDs 597 and 598). The SWC reader must be making some assumption that the somas are all at the beginning. I will create a small data sample to reproduce this and see about fixing it.

juanchopanza:
OK, the morphology looks broken. Maybe @lida can comment on whether NeuroM should try to read a file with this structure or not. Here is the issue:

There is a starting soma point, that has two children: an apical dendrite branch, and another soma branch!

Here's the first soma point, with the first couple of basal dendrite points:
```
1 1 436.0 -1566.0 533.2 1.5 -1
2 3 435.1 -1583.0 500.0 1.5 1
3 3 438.2 -1586.0 500.0 1.5 2
```

And here is the second soma branch, which has two soma points and is followed by two apical dendrite points:
```
597 1 435.0 -1567.0 531.2 1.5 1
598 1 444.7 -1568.0 566.8 1.5 597
599 4 444.4 -1568.0 600.0 1.5 598
600 4 443.2 -1566.0 600.0 1.5 599
```

I don't think we can make sense out of this data. At the very least, it the soma points to not satisfy [any of out supported definitions](http://neurom.readthedocs.io/en/latest/definitions.html#soma), but even if we extend the number of definitions supported, I am not sure how we'd interpret these points.

juanchopanza:
[...]

PR #496, opened 2016-09-08
Add detailed information about somata in SWC format

lidakanari:


doc/source/file_formats.rst
```diff
@@ -46,9 +46,48 @@ experimental support for NeuroLucida .asc files.
 SWC
 ---
 
-The SWC format represents a neuron as a tree of 3D points with additional information.
+The SWC format represents a neuron as a set of trees that are connected to a soma.
+The neuronal morphology is encoded as a rooted tree of 3D points and the corresponding radii.
 More information can be found `here <http://research.mssm.edu/cnic/swc.html>`_.
 
+The points of type "1" that represent the soma, have to be in the beginning of the file.
```
juanchopanza: We decided earlier that this restriction doesn't apply. Out code doesn't require that any more. See #438 and #478.

doc/source/file_formats.rst
```diff
@@ -46,9 +46,48 @@ experimental support for NeuroLucida .asc files.
...
+The soma can be represented in one of the following formats: 
+
+TypeA: One point soma
+^^^^^^^^^^^^^^^^^^^^^^
+
+The soma is represented by its center and a radius that corresponds to the sphere
+that preserves the surface area of the soma. The center of the soma has as parent ID -1 
+and all the initial points of the neuronal trees are connected to the soma. 
+
+TypeB: Three point soma
```
juanchopanza: It is probably worth noting that NeuroM currently rejects this soma type.

doc/source/file_formats.rst
```diff
@@ -46,9 +46,48 @@ experimental support for NeuroLucida .asc files.
...
+^^^^^^^^^^^^^^^^^^^^^^
+
+The soma is represented by three points that correspond to the center of the soma (x, y, z), 
+with parent ID -1, and two diametrically opposite points (x, y-r, z) and (x, y+r, z), where r 
+is the radius of the sphere that  preserves the surface area of the soma. The radii of the soma 
+points are not necessarily meaningful, so they can be set to zero. As such, they should not be 
+taken into account in further calculations. The rest of the soma points are connected to the 
+center of the soma, as well as the first points of all the neuronal trees. 
+
+TypeC: N - points soma
```
juanchopanza: Also worth noting NeuroM doesn't do this currently.
Just to be clear, this would allow 2-point somas, right?

doc/source/file_formats.rst
```diff
@@ -46,9 +46,48 @@ experimental support for NeuroLucida .asc files.
...
+
+TypeB: Three point soma
+^^^^^^^^^^^^^^^^^^^^^^
+
+The soma is represented by three points that correspond to the center of the soma (x, y, z), 
+with parent ID -1, and two diametrically opposite points (x, y-r, z) and (x, y+r, z), where r 
+is the radius of the sphere that  preserves the surface area of the soma. The radii of the soma 
+points are not necessarily meaningful, so they can be set to zero. As such, they should not be 
+taken into account in further calculations. The rest of the soma points are connected to the 
+center of the soma, as well as the first points of all the neuronal trees.
```
juanchopanza: And this sounds like yet more points :-)

doc/source/file_formats.rst
juanchopanza: Also note that the contour soma described in http://neuromorpho.org/SomaFormat.html does not treat the first point as the centre, but treats all the points equally. Is Type C supposed to be something else?

doc/source/file_formats.rst
lidakanari: The restriction concerns the fact bifurcation of tree from a soma point. I was not aware if the points had to be in the beginning of the file which is something different. After the meeting about reconstructions (with Julian, Rembrandt and Giorgio) I became aware of this restriction.

doc/source/file_formats.rst
lidakanari: It shouldn't reject them though.

doc/source/file_formats.rst
lidakanari: That answer was given by the person who wrote this document. Apparently, this is not clear in the document!

lidakanari:
This documentation was updated to conform to the general standards of the SWC format and not what NeuroM is currently doing. I guess we have to either fix NeuroM accordingly, or change the claim that we support SWC (standard) format. If some points are not clear I will be happy to rewrite them or start a follow-up discussion to clarify those issues, with the people that are more familiar with the expected representation in this file format.

doc/source/file_formats.rst
lidakanari: Yes, I will change it to the two other soma points for clarity :)

doc/source/file_formats.rst
juanchopanza: I mean the restriction that all the soma points have to be at the beginning of the file. We lifted that, and it is not really necessary to impose it. If we do, then cells such as the one used by Armando and Nancy in #438 won't be valid. I don't mind, but I'm not sure we gain anything from that.

doc/source/file_formats.rst
juanchopanza: But it does, since the changes for issue #475. I think it has to be clear in this section what is implemented in NeuroM and what is SWC "standard". And if there is a standard, it would be good to have a reference. Otherwise it could change arbitrarily in the future.

doc/source/file_formats.rst
juanchopanza: Which "answer" are you talking about?

juanchopanza:
If there is an SWC standard from which this comes from, we need a reference. Otherwise it is a set of arbitrary restrictions, which I guess we will implement in NeuroM in the future. But readers of this section may think NeuroM already implements this, so it would be better to clarify that it doesn't (yet.)

lidakanari:
There is no SWC standard reference. I think we have to create one, in the absence of a better solution. Yes, NeuroM doesn't do that yet. Should I add this in the documentation?

doc/source/file_formats.rst
lidakanari: I don't say that it is what we should impose. However, I discussed this point with the "experts" and they agreed that this should be the case! In the future I guess reconstructors should have to fix this in their data!

doc/source/file_formats.rst
lidakanari: Yes, all the data from neuromorpho.org are in this format though. This means we reject 90% of the neuron data. This should not be the case for three point somata and it was very clear in the discussion that this is the standard. For this point the documentation in http://neuromorpho.org/SomaFormat.html is clear at least.

doc/source/file_formats.rst
lidakanari: OK, I will contact the "experts" again and get a written confirmation of what the contour is supposed to be. Yes, we are indeed getting contradicting information, that's why I had to contact people instead of refer to a document. Let's not change this one in the code before I clarify it!

doc/source/file_formats.rst
juanchopanza: OK, we can refine some of the changes in 479, so that a bifurcating soma is accepter only if the bifircation is from the 1st point and the soma has 3 points.

juanchopanza:
Yes, I think adding a short note would be very helpful. You can use the `.. note::` syntax to create a note box.

doc/source/file_formats.rst
lidakanari: No, a bifurcation is still not allowed in any type of some (crazy I know). It depends what you mean by bifurcation. So, we do not accept something like the following:
```
o---o---o---o
    |
    o
    |
    o
```

But the three point some is represented as: 
```
o (Parent_ID = 0)
|
o (Parent_ID = -1)
|
o (Parent_ID = 0)

```
because all the points are connected to the center of the soma.  See the picture from http://neuromorpho.org/SomaFormat.html :

juanchopanza: Here, the point with parent ID -1 is a soma point and has two children. So, it is a soma point with a bifurcation into soma points. That is not allowed anymore in NeuroM but it is one of the SWC "standard" configurations. So, this kind of bifurcation **should** be allowed, but isn't, since #475.

lidakanari: OK, this type of soma that you call bifurcation should be allowed according to the latest information. Any other bifurcation should not.

juanchopanza: Agreed. As per my earlier comment, restricted to only three soma points, and only the first one (with parent ID -1) can bifurcate.

lidakanari: OK, let's wait to see what happens for the contour of the SWC somata.

doc/source/file_formats.rst
juanchopanza: One more thing about the 3-point soma. Should we check that the points are aligned in Y by +/- R? Note that this will make rotated morphologies "incorrect".

doc/source/file_formats.rst
lidakanari: Hi @rbakker, I'd like to confirm the following points with you:

1) Is the three point soma in SWC format expected to have the following three points:
- center of the soma (x, y, z)
- translation in the y-coordinate (x, y + R, z), connected to the center
- translation in the y-coordinate (x, y - R, z), connected to the center

2) Is the multiple point soma in SWC expected to have:
- the center of the soma as the first point
- are the rest of the points connected to the center? 

Thanks a lot,
Lida

rbakker: Hi @lida, I can confirm your first point. From the discussion with Giorgio I understood that for your case 2), if a Neurolucida file is converted that has multiple contours, then each contour is converted to a center+radius.
The first contour that appears in the file becomes point 1, the next point 2 etc. The neurites are attached to point 1. So, it is NOT checked whether this point 1 is from a contour that represents the middle of the soma.

lidakanari: Thanks a lot! OK, that's very helpful! Could you also let me know if all the soma points are expected to be connected to the initial point, even if this is not the center?

juanchopanza: @rbakker About the first point, do you know the reason for having the 2nd and 3rd points, instead of a single point soma with the appropriate radius set? It seems to me that the extra points don't add any information.

doc/source/file_formats.rst
```diff
@@ -46,9 +46,54 @@ experimental support for NeuroLucida .asc files.
...
+that preserves the surface area of the soma. The center of the soma has as parent ID -1 
+and all the initial points of the neuronal trees are connected to the soma. 
+
+TypeB: Three point soma
+^^^^^^^^^^^^^^^^^^^^^^
+
+The soma is represented by three points that correspond to the center of the soma (x, y, z), 
+with parent ID -1, and two diametrically opposite points (x, y-r, z) and (x, y+r, z), where r 
+is the radius of the sphere that  preserves the surface area of the soma. The radii of the soma 
+points are not necessarily meaningful, so they can be set to zero. As such, they should not be
```
juanchopanza: How should the soma be interpreted it any of the points has a non-zero radius? Leaving this possibility open without specifying the semantics is problematic.

doc/source/file_formats.rst
```diff
@@ -46,9 +46,54 @@ experimental support for NeuroLucida .asc files.
...
 ---
 
-The SWC format represents a neuron as a tree of 3D points with additional information.
+The SWC format represents a neuron as a set of trees that are connected to a soma.
+The neuronal morphology is encoded as a rooted tree of 3D points and the corresponding radii.
 More information can be found `here <http://research.mssm.edu/cnic/swc.html>`_.
 
+.. note::
+    The following information about SWC format concern the standard representation as 
+    it has been established in the neuromorpho.org database. We would like to clarify
```
juanchopanza: Is there a URL for this specification? If so, a link would be good.

doc/source/file_formats.rst
```diff
@@ -46,9 +46,54 @@ experimental support for NeuroLucida .asc files.
...
+with parent ID -1, and two diametrically opposite points (x, y-r, z) and (x, y+r, z), where r 
+is the radius of the sphere that  preserves the surface area of the soma. The radii of the soma 
+points are not necessarily meaningful, so they can be set to zero. As such, they should not be 
+taken into account in further calculations. The other two of the soma points are connected to the 
+center of the soma, as well as the first points of all the neuronal trees. 
+
+TypeC: N - points soma
+^^^^^^^^^^^^^^^^^^^^^^
+
+The soma is represented by a set of points. The first point corresponds to the center of the soma
```
juanchopanza: We still need to clarify this first point issue. I don't think the first point should be special.

lidakanari: Probably it should not be, but this is the case for most of the data as I understand it. This has no effect on the computation of the actual center but has a serious effect on the computation of the soma radius. Ideally, we should have a way of knowing if the first point is the center or not.

doc/source/file_formats.rst
lidakanari: Will add the _current_ URL.

doc/source/file_formats.rst
lidakanari: If it is not a one-point soma or a multiple contour some, then we should ignore the values of the radius of the points and compute the correct equivalent radius.

juanchopanza: Right, but it is a bad standard to allow the possibility of radius values being there without saying how they will be treated. So either non-zero radii are an error, or we specify that they are always ignored. That is besides the point that it makes no sense to have these three points instead of one!

doc/source/file_formats.rst
juanchopanza: We need confirmation of this. Because if it is the case, then we don't need to calculate the center. And it will affect the calculation, unless all the mean position of all the non-center points coincides with the center. And it will require special treatment for SWC wrt. H5.

juanchopanza: Maybe if you remove the part about the first point, we can merge this and then add it back again (or not) once we get some clarification.

lidakanari: OK, removed.

PR #498, opened 2016-09-12
Re-enable bifurcating soma for 3 point case.

### LINKED ISSUE #475: Add test to ensure no bifurcations occur in the soma

juanchopanza:
The neuromorpho 3 point soma has the first soma point bifurcating
into the 2nd and 3rd soma points. It applies to 3 point somata only.

Note: the code does not impose the topology described above, but
accepts any 3-point soma topology that satisfies the SWC format
(technically this means 1st point bifurcating, or linear sequence of 3
points.) The first point is always taken to represent the center.

Re-structure SWC soma test samples and add one with neuromorpho 3pt
soma.

Resolves #475.


juanchopanza:
@lidakanari Related to the soma discussions. Allowing bifurcation in a 3pt soma, + other SWC requirements, means the bifurcation happens at first point.

lidakanari:
Sure, looks good!

PR #508, opened 2016-09-28
Attempt to use travis python functionality.

juanchopanza:
Unclear if this has been fixed in the trusty beta.


juanchopanza:
@jdcourcol @mgeplf FYI, I am using the travis python functionality again. This wasn't working in the trusty beta a while back, so I had to implement a manual solution.

PR #514, opened 2016-09-29
add CSV output to morph_stats

mgeplf:


lidakanari:
Looks good to me. Thanks a lot Mike!

apps/morph_stats
```diff
@@ -93,12 +93,35 @@ def _get_config(config):
         return _CONFIG
 
 
-if __name__ == '__main__':
-    args = parse_args()
-    logging.basicConfig(level=(logging.WARNING,
-                               logging.INFO,
-                               logging.DEBUG)[min(args.verbose, 2)])
-
+def _get_header(results):
```
juanchopanza: I would prefer it if this code was tested. It can go in `neurom.apps`. This script should stay minimal.

apps/morph_stats
```diff
@@ -93,12 +93,35 @@ def _get_config(config):
...
+    for k, v in results.values()[0].items():
+        if isinstance(v, dict):
+            for metric in v.keys():
+                ret.append('%s:%s' % (k, metric))
+        else:
+            ret.append(k)
+    return ret
+
+
+def _generate_flattened_dict(headers, results):
```
juanchopanza: Same here.

apps/morph_stats
mgeplf: Totally agree.

I'll move it there when I get some time.  Thanks for pointing out where it should go.

PR #516, opened 2016-09-30
Fixes for readthedocs installation

juanchopanza:
* Avoid installing dependencies when in readthedocs environment.


setup.py
```diff
@@ -60,7 +60,12 @@ def parse_reqs(reqs_file):
 
 
 BASEDIR = os.path.dirname(os.path.abspath(__file__))
-REQS = parse_reqs(os.path.join(BASEDIR, 'requirements.txt'))
+
+# Hack to avoid installation of modules with C extensions
```
mgeplf: I'd mention that the hack is for RTD (eventhough it says read the doc on the next line

Issue #504, opened 2016-09-28
Neuronal density


lidakanari:
According to the following paper: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4092368/ the neuronal density is defined as the total volume of a tree divided by the volume of the convex hull that encloses the tree.

This feature should be computed per tree and therefore is TreeType specific.

juanchopanza:
I won't read the paper, but I have a question: the feature should be computed by tree type, but does it have to be aggregated by tree type, or should it produce a value per neurite? The latter gives the caller more freedom.

lidakanari:
This is a personal opinion so it is completely subjective: for me it makes sense to be computed as a value per neurite. So you will have a value of density per tree and as you suggest the user can then sort them by tree-type.

juanchopanza:
OK, good. That would also be my preference.

juanchopanza:
I haven't fully put this in features, but with the latest changes it is easier to extract the information. For example,

```
from neurom.geom import convex_hull
def density(neurite):
    return neurite.volume / convex_hull(neurite).volume
```

and then

```
import neurom as nm
nrn = nm.load_neuron('....')
d = [density(n) for n in nrn.neurites]
```

I think this should be enough to get started with evaluating this quantity in the classification.

lidakanari:
Thanks! That's very helpful...

juanchopanza:
Resolved by #517.

PR #517, opened 2016-09-30
Add neurite_volume_density feature to neurom.get.

juanchopanza:
Returns array with the volume density of each neurite. The
volume density is the ratio of neurite volume to neurite convex hull
volume.


juanchopanza:
@lidakanari Does this resolve #504? It gives you per-neurite densities.

lidakanari:
Sure, looks good to me!

juanchopanza:
OK good. Note that this means you can get the values with `morph_stats` now.

PR #518, opened 2016-10-03
Update jupyter tutorial notebook.

juanchopanza:
Update installation instructions.
Use morph_stats instead of example feature extraction script.
Add links to morph_check and morph_stats documentation.


juanchopanza:
@lidakanari FYI, some updates that could be relevant for tutorials.

PR #526, opened 2016-10-11
Simplify installation instructions.

juanchopanza:
* No longer suggest virtualenv --system-site-packages as default,
  but let NeuroM install all dependencies into clean virtualenv.
  Mention --system-site-packages as an alternative, allowing to use
  system-wide packages.

* Increase version to 1.1.1.


juanchopanza:
@liesbethvanherpe FYI.

PR #528, opened 2016-10-14
Update examples.

juanchopanza:
Use neurom instead of neurom.fst where possible.


juanchopanza:
@lidakanari FYI, I found places where fst was still used unnecessarily in the examples. I thought I had already taken care of these but it seems not to be the case.

PR #529, opened 2016-10-14
Remove unnecessary helper functions from fst module.

juanchopanza:
Functions and objects that ara available from the piblic API, directly in
the neurom module, have been removed:

* load_neuron
* load_neurons
* NEURITE_TYPES

Deprecated function fst.load_population has been removed.

### DIFF
--- a/doc/source/definitions.rst
+++ b/doc/source/definitions.rst
@@ -94,7 +94,7 @@ whose first and last element are any of the following combinations:
 The first point of a section is a duplicate of the last point of its parent section,
 unless the latter is a soma section.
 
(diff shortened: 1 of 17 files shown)


lidakanari:
Could you possibly not change this until after the summit? :) Some users are already accustomed to the use of this module and it would be useful to leave it as it is for the demonstration of the current functionality...

juanchopanza:
It is not necessary to demonstrate the current functionality. That has been available directly under `neurom` for a long time now, and announced.

lidakanari:
Don't worry, we didn't use fst, not even once :P Thanks for waiting!

PR #539, opened 2016-10-26
Added checks to make sure app configs are correct

mgeplf:
* when they are loaded, so that things fail faster if there
  is a config error

### DIFF
--- a/apps/morph_stats
+++ b/apps/morph_stats
@@ -39,8 +39,8 @@ from neurom.apps import get_config
 from neurom.apps.morph_stats import (extract_stats,
                                      sanitize_config,
                                      get_header,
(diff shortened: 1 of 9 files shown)


apps/morph_stats
```diff
@@ -82,49 +84,44 @@ def parse_args():
...
-    else:
-        return _CONFIG
-
-
 def main(args):
     '''main function'''
-    _f = args.datapath
-    _results = {}
-    _config = _get_config(args.config)
+    path = args.datapath
```
mgeplf: renaming:
_f -> path
_results -> results
_config -> config

juanchopanza:
Looks good now.

mgeplf:
fixes #539

PR #540, opened 2016-10-26
add some doc for DataWrapper

mgeplf:


neurom/io/datawrapper.py
```diff
@@ -36,7 +36,25 @@ class DataWrapper(object):
...
+            X(float): x-coordinate
+            Y(float): y-coordinate
+            Z(float): z-coordinate
+            R(float): radius
+            TYPE(integer): one of the types described by POINT_TYPE
+            ID(integer): unique integer given to each point, the `ROOD_ID` is -1
+            P(integer): the ID of the parent
+
+        Notes:
+            - there is no ordering constraint: a child can reference a parent ID that comes
```
mgeplf: @juanchopanza: please edit this, add to it if there are other invariants, or if I got something wrong

mgeplf: Nothing needs to be added?

juanchopanza: I think it is OK like this.

PR #541, opened 2016-11-01
added terminal_length_per_neurite feature extraction

mgeplf:
* allow for getting the path lengths to all terminal
  point per neurite


neurom/fst/_neuritefunc.py
```diff
@@ -182,6 +182,13 @@ def total_length_per_neurite(neurites, neurite_type=NeuriteType.all):
                 for n in iter_neurites(neurites, filt=is_type(neurite_type)))
 
 
+def terminal_length_per_neurite(neurites, neurite_type=NeuriteType.all):
```
lidakanari: I would modify the name to "terminal_path_length" to avoid possible confusion between length and path_length.

lidakanari:
Seems good to me.

neurom/fst/tests/test_neuritefunc.py
```diff
@@ -186,3 +186,14 @@ def test_volume_density_per_neurite():
...
     nt.ok_(np.allclose(vol_density, ref_density))
+
+
+def test_terminal_length_per_neurite():
+    nrn = nm.load_neuron(os.path.join(SWC_PATH, 'simple.swc'))
+    terminal_distances = np.array(_nf.terminal_path_lengths_per_neurite(nrn))
+    np.testing.assert_allclose(terminal_distances,
+                               np.array([4 + 5., 4 + 6., 4. + 6., 4. + 5]))
+    terminal_distances = np.array(_nf.terminal_path_lengths_per_neurite(
+        nrn, neurite_type=nm.AXON))
```
jdcourcol: From what I see in the swc, the result would be the same with nm.DENDRITE right?
can we have a difference somewhere between the 2 neurites ?

jdcourcol: only one of the neurite is sufficient. What I say is that you get the same result with whatever neurite you select.

PR #543, opened 2016-11-02
Updated documentation for neurom.get() function

mgeplf:
* Added documentation of features to morph_stats documentation
* Add printing of docstring, when available, for features, so
  their descriptions are available in the .get() description,
  and elsewhere
* Added '-l' option to morph_stats so the features can be listed

Note: Need to have a way of documenting functions that are using
functools.partial.

### DIFF
--- a/apps/morph_stats
+++ b/apps/morph_stats
@@ -60,15 +60,18 @@ _CONFIG = {
 }
 
 
-def parse_args():
+def get_parser():
     '''Parse command line arguments'''
     parser = argparse.ArgumentParser(description='Morphology statistics extractor')
 
-    parser.add_argument('datapath',
+    parser.add_argument('datapath', nargs='?',
                         help='Path to a morphology data file or a directory')
 
     parser.add_argument('-v', '--verbose', action='count', dest='verbose', default=0,
                         help='-v for INFO, -vv for DEBUG')
+
+    parser.add_argument('-l', '--list', action='store_true', default=False,
+                        help='List the available features')
 
     parser.add_argument('--as-population',
                         action='store_true',
@@ -81,7 +84,7 @@ def parse_args():
                         help=('Summary output file name, if it ends in .json, '
                               'a json file is created, if .csv, then a csv file'))
 
-    return parser.parse_args()
+    return parser
 
 
 def main(args):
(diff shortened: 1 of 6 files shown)


apps/morph_stats
```diff
@@ -60,16 +60,19 @@ _CONFIG = {
 }
 
 
-def parse_args():
+def get_parser():
     '''Parse command line arguments'''
     parser = argparse.ArgumentParser(description='Morphology statistics extractor')
 
-    parser.add_argument('datapath',
+    parser.add_argument('datapath', nargs='?',
```
mgeplf: I have to do this so that the '-l' option doesn't require a datapah

doc/source/morph_stats.rst
```diff
@@ -112,8 +111,13 @@ are to be split. This is a sample output using the above configuration:
...
 .. code-block:: bash
 
     morph_stats --help
 
+Features
+********
+
+.. runblock:: console
+
+    $ morph_stats -l
```
mgeplf: sadly the autorun plugin doesn't allow me to hid the commands executed when using the pycon driver, so I have to use the console one.

neurom/fst/__init__.py
```diff
@@ -137,24 +137,47 @@ def get(feature, obj, **kwargs):
...
+    ret = indent + string.replace('\n', '\n' + indent)
+    return ret.rstrip()
+
+
+def _get_doc():
+    '''Get a description of all the known available features'''
+    def get_docstring(func):
+        '''extract doctstring, if possible'''
+        docstring = ':\n'
+        if not isinstance(func, partial) and func.__doc__:
```
mgeplf: I believe I'm going to have to create a FeatureFunc structure so that I can override the docstrings for lambda's and partials, since many of the features use this, but it means their description is blank.

I will probably do that in another commit, as this one is big enough as is, I think

apps/morph_stats
```diff
@@ -123,8 +126,16 @@ def main(args):
...
 
 if __name__ == '__main__':
-    _args = parse_args()
+    _args = get_parser().parse_args()
     logging.basicConfig(level=(logging.WARNING,
                                logging.INFO,
                                logging.DEBUG)[min(_args.verbose, 2)])
+
+    if _args.list:
+        print nm.fst._get_doc()  # pylint: disable=W0212
```
juanchopanza: At some point the `get()` function should move to a public API module (together with other feature-getter helper code). This is a remnant of the section tree implementation work, which I didn't get to clean up completely.

mgeplf: Ok, good to know.

PR #546, opened 2016-11-15
fix places where h5py.File might be leaked due to exceptions

mgeplf:


neurom/point_neurite/io/tests/test_h5_reader.py
```diff
@@ -48,19 +48,18 @@
 
 
 def test_get_version():
-    v1 = h5py.File(os.path.join(H5V1_PATH, 'Neuron.h5'), mode='r')
-    v2 = h5py.File(os.path.join(H5V2_PATH, 'Neuron.h5'), mode='r')
-    nt.assert_equal(hdf5.get_version(v1), 'H5V1')
-    nt.assert_equal(hdf5.get_version(v2), 'H5V2')
-    v1.close()
-    v2.close()
+    with h5py.File(os.path.join(H5V1_PATH, 'Neuron.h5'), mode='r') as v1:
```
mgeplf: seems similar to neurom/io/tests/test_h5_reader.py, need to look into that later

PR #548, opened 2016-11-15
Add BlockNeuronBuilder to ease creating DataWrapper blocks

mgeplf:
* When loading morphologies, there are cases where all
  the segments are in one packed 'block' of points & radii.
  If one knows the parents for each of the sections w/ blocks
  of segments, this helper makes it easier to build the
  DataWrapper block
* Similar to the functionality in h5vX loader
* h5v1 and h5v2 ported over

### DIFF
--- a/neurom/io/hdf5.py
+++ b/neurom/io/hdf5.py
@@ -41,11 +41,10 @@ There is one such row per measured point.
 
 '''
 
-from collections import namedtuple
 import h5py
 import numpy as np
-from ..core.dataformat import COLS
-from .datawrapper import DataWrapper
+
+from .datawrapper import DataWrapper, BlockNeuronBuilder
 from neurom._compat import zip_longest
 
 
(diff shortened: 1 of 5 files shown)


neurom/io/hdf5.py
```diff
@@ -59,21 +58,20 @@ def get_version(h5file):
...
-    '''Read a file and return a tuple of data, format.
+    '''Read a file and return a `data_wrapper'd` data
 
     * Tries to guess the format and the H5 version.
     * Unpacks the first block it finds out of ('repaired', 'unraveled', 'raw')
 
     Parameters:
-        remove_duplicates: boolean, \
-        If True removes duplicate points \
+        remove_duplicates: boolean, If True removes duplicate points
```
juanchopanza: I think this functionality can be removed as soon as the point neurite stuff gets killed.

mgeplf: Ok, good to know.  Did you have plans on when that would be happening?

mgeplf: Good to know, thanks for the input.

Let's take this discussion to #553, so I know exactly what I need to do.

PR #550, opened 2016-11-16
Lsqshr python3

mgeplf:


mgeplf:
@lsqshr Could you test that this works for everything you need?  Thanks for the changes.

lsqshr:
@mgeplf Thanks for the changes. Most of the things work like charm. 

When I run morph_stats, it complained that 'map' cannot be dumped in json. It also would not work to change everything to list in eval_stats since the "mean_soma_radius" is only a float. So I made a small type check to only convert maps to list. It works now for me with python3.5 on multiple swc files now. Pls see #554

PR #551, opened 2016-11-25
add python 2.6 to travis

mgeplf:
- attempted move to container based build system if possible, but the whitelist of allowed packages doesn't seem to work with scipy & hdf5 requirements.
- Build time, once cached wheels exist (ie: after the first build) are under 2 minutes, which seems reasonable
- I tested that coverage and failing tests will fail the build for both 2.6 and 2.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,29 +2,29 @@ sudo: required
 dist: trusty
 language: python
 python:
+  - 2.6
   - 2.7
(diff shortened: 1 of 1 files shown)


juanchopanza:
+1, Looks OK to me. Note that each python version will take up a slot in the queue system, which is shared with all the BlueBrain things that are on travis.

PR #552, opened 2016-11-28
cleanup/readability

mgeplf:
* removed extra parens in places
* remove unused ok() from neurom/check/__init__.py

### DIFF
--- a/neurom/fst/_neuronfunc.py
+++ b/neurom/fst/_neuronfunc.py
@@ -33,7 +33,12 @@ import numpy as np
 from neurom.core.types import NeuriteType
 from neurom.core.types import tree_type_checker as is_type
 from neurom.core.dataformat import COLS
-from neurom import morphmath as mm
+from neurom import morphmath
+
+
+def neuron_population(nrns):
+    '''Makes sure `nrns` behaves like a neuron population'''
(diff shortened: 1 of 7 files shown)


neurom/fst/_neuronfunc.py
```diff
@@ -33,7 +33,12 @@
 from neurom.core.types import NeuriteType
 from neurom.core.types import tree_type_checker as is_type
 from neurom.core.dataformat import COLS
-from neurom import morphmath as mm
+from neurom import morphmath
+
+
+def neurons_pop(nrns):
```
jdcourcol: I don't feel 'pop' is a correct name here. I sounds to me that the result should be removed from nrns.
neurons_as_iterable ? - not super pretty though -

mgeplf: Yeah, I wasn't a fan either.  Neurons_as_iterable is too generic, as it could mean any of the many iters that neurons have, in my mind.

How about `neuron_population`?

jdcourcol: ahhh. pop was refering to population. I believed that it was pop like list.pop()
ok for neuron_population. I don't have a better name.

PR #554, opened 2016-11-28
fix morph_stats.eval_stats by checking the returning type

lsqshr:
When I run morph_stats, it complained that 'map' cannot be dumped in json. It also would not work to change everything to list in eval_stats since the "mean_soma_radius" is only a float. So I made a small type check to only convert maps to list. It works now for me with python3.5

### DIFF
--- a/neurom/apps/morph_stats.py
+++ b/neurom/apps/morph_stats.py
@@ -52,7 +52,11 @@ def eval_stats(values, mode):
         mode = 'sum'
 
     try:
-        return getattr(np, mode)(values)
+        tval = getattr(np, mode)(values)
+        if type(tval) is map:  # map cannot be dumped in json
+            return list(tval)
+        else:  # For "mean_soma_radius"
+            return tval
     except ValueError:
         pass
 


juanchopanza:
That's interesting. It may be more productive to submit a bug report with enough information and data to reproduce the problem. See [reporting issues](http://neurom.readthedocs.io/en/latest/issue_reporting.html). It would be very useful for neurom developers to know, if `morph_stats` stopped working at some point, why this was and how it wasn't caught by any of the tests.

As for submitting pull requests, the process will be much faster if you first check whether the tests pass locally. In your case, they haven't because they no longer have the required 100% coverage. Every change should come with a test :-)

mgeplf:
Hmm, I don't seem to be having a problem w/ 'mean_soma_radius'.

Ex:
```
$ morph_stats test_data/swc/simple.swc
{
  "test_data/swc/simple.swc":{
    "axon":{
      "total_section_length":15.0,
      "max_section_length":6.0,
      "max_section_branch_order":1,
      "total_section_volume":24.085543677521748
    },
    "all":{
      "total_section_length":31.0,
      "max_section_length":6.0,
      "max_section_branch_order":1,
      "total_section_volume":51.312680008633286
    },
    "mean_soma_radius":1.0,
    "apical_dendrite":{
      "total_section_length":0.0,
      "max_section_length":null,
      "max_section_branch_order":null,
      "total_section_volume":0.0
    },
    "basal_dendrite":{
      "total_section_length":16.0,
      "max_section_length":6.0,
      "max_section_branch_order":1,
      "total_section_volume":27.227136331111538
    }
  }
}
```

Could you send the command you're running, with the config file, if applicable.  Thanks.

lsqshr:
Thanks Juan. I do agree with your point. Will raise an issue next time! will also look into how the tests should be amended.

juanchopanza:
@lsqshr I would prioritize raising an issue with a reproducible test case, because if this can't be reproduced, there is no reason to change anything.

lsqshr:
@juanchopanza @mgeplf Cool! Let's track it in #555.

PR #556, opened 2016-12-01
Remove point_neurite

mgeplf:
* we want to live in a fst world


doc/source/examples.rst
```diff
@@ -57,20 +57,6 @@ Here we visualize a neuronal morphology:
     >>> fig, ax = viewer.draw(nrn, mode='3d') # valid modes '2d', '3d', 'dendrogram'
     >>> fig.show()
 
-Basic feature extraction example
```
juanchopanza: Why is this example removed?

neurom/fst/tests/test_feature_compat.py
```diff
@@ -28,6 +28,7 @@
 
 '''Test neurom._point_neurite.features and neurom.fst features compatibility'''
```
juanchopanza: This is no longer applicable, so the name and docstring should be changed to reflect whatever this test does now.

doc/source/examples.rst
mgeplf: It's in the spirit of getting people to use morph_stats instead of rolling their own.

lidakanari:
I have updated versions for most of the examples (simple ones) as they have been presented in the HBP summit. If it is useful I can add them when the clean-up is completed.

PR #561, opened 2016-12-05
move iteration out of the core/__init__.py to core/_neuron.py

### LINKED ISSUE #558: iterator to get segments from a section.
If I have a section, I would like to be able to iterate on its segments.I didn't find an obvious way to do it.
I can get the 'list' of segment from a section that way:
segs = zip(sec.points[:-1], sec.points[1:])

but I don't find it straightforward for the end user.

mgeplf:
* Allows for functions in _neuron to refer to iterators
* taught iter_segments to take a Section() (fixes #558)

### DIFF
--- a/neurom/core/_neuron.py
+++ b/neurom/core/_neuron.py
@@ -134,8 +134,7 @@ class Section(Tree):
         The area is calculated from the segments, as defined by this
         section's points
         '''
-        return sum(segment_area(s) for s in zip(self.points[:-1],
-                                                self.points[1:]))
+        return sum(segment_area(s) for s in iter_segments(self))
 
     @property
     @memoize
@@ -145,8 +144,7 @@ class Section(Tree):
         The volume is calculated from the segments, as defined by this
         section's points
         '''
-        return sum(segment_volume(s) for s in zip(self.points[:-1],
-                                                  self.points[1:]))
+        return sum(segment_volume(s) for s in iter_segments(self))
 
     def __str__(self):
         return 'Section(id = %s, points=%s) <parent: %s, nchildren: %d>' % \
--- a/neurom/core/_neuron.py
+++ b/neurom/core/_neuron.py
@@ -28,13 +28,88 @@
 
 '''Neuron classes and functions'''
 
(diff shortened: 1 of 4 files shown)


neurom/core/_neuron.py
```diff
@@ -28,13 +28,88 @@
...
+    '''
+    return chain.from_iterable(iterator_type(neurite.root_node)
+                               for neurite in iter_neurites(neurites, filt=neurite_filter))
+
+
+def iter_segments(obj, neurite_filter=None):
+    '''Return an iterator to the segments in a collection of neurites
+
+    Parameters:
+        obj: neuron, population, neurite, section, or iterable containing neurite objects
```
mgeplf: this is the only API change, and it's a widening, so should be ok, correct?

mgeplf: this is the only API change, and it's a widening, so should be ok, correct?

neurom/core/_neuron.py
```diff
@@ -131,7 +206,7 @@ def transform(self, trans):
 
     def iter_sections(self, order=Tree.ipreorder):
         '''iteration over section nodes'''
-        return order(self.root_node)
+        return iter_sections(self, iterator_type=order)
```
mgeplf: I'm tempted to add iter_sections, iter_segments, iter_neurites to Neuron() and Population(), and iter_segments to Section(), thus making the iterators tab-compleable.

Thoughts?

jdcourcol: I am more than happy to have the iter_segments on the section since I opened #558

mgeplf: Ok, that is a big change in how the API works: moving away from bare functions for iteration to object style iteration.  I'm not married to either concept, and we can concurrently support them, but there are benefits to bare functions, and hinting towards using them (by not having object accessible iterators) has benefits.

Are we comfortable making that switch?

Also, that is independent of this PR, so if you're happy with it, let me know.
Finally, #558 iter_segments now takes a Section(), so that should be closed when this is closed.

PR #566, opened 2016-12-07
remove unused log.py

mgeplf:


juanchopanza:
Note that this had been added to make it easy for users to set up a logger by importing this module. Logging messages were being lost.

mgeplf:
Hmm, I don't see references to it in the docs, and it seems that no tests were covering it, so I didn't think it was being used.

juanchopanza:
Yes, it seems I never got round to documenting this. I think there used to be an example at some point. Do you know of a simpler or more idiomatic way of achieving the same?

mgeplf:
My personal preference is to put this as an example in the documentation, and then point users to the official python documentation for more logging information.

PR #570, opened 2016-12-13
add Sholl analysis

mgeplf:
Initial version

### DIFF
--- a/neurom/fst/_neuronfunc.py
+++ b/neurom/fst/_neuronfunc.py
@@ -30,9 +30,11 @@
 
 import math
 import numpy as np
+from neurom.geom import bounding_box
 from neurom.core.types import NeuriteType
 from neurom.core.types import tree_type_checker as is_type
 from neurom.core.dataformat import COLS
+from neurom.core._neuron import iter_neurites, iter_segments
 from neurom import morphmath
 
 
@@ -136,3 +138,71 @@ def trunk_origin_elevations(nrn, neurite_type=NeuriteType.all):
     return [_elevation(s.root_node.points, n.soma)
             for n in nrns
             for s in n.neurites if neurite_filter(s)]
+
+
+def sholl_crossings(neurites, center, radii):
+    '''calculate crossings of neurites
+
+    Args:
+        nrn(morph): morphology on which to perform Sholl analysis
+        radii(iterable of floats): radii for which crossings will be counted
+
+    Returns:
+        Array of same length as radii, with a count of the number of crossings
+        for the respective radius
+    '''
+    def _count_crossings(neurite, radius):
+        '''count_crossings of segments in neurite with radius'''
+        r2 = radius ** 2
+        count = 0
+        for start, end in iter_segments(neurite):
+            start_dist2, end_dist2 = (morphmath.point_dist2(center, start),
+                                      morphmath.point_dist2(center, end))
+
+            count += int(start_dist2 <= r2 <= end_dist2 or
+                         end_dist2 <= r2 <= start_dist2)
+
+        return count
+
+    return np.array([sum(_count_crossings(neurite, r)
+                         for neurite in iter_neurites(neurites))
+                     for r in radii])
+
+
+def sholl_frequency(nrn, neurite_type=NeuriteType.all, step_size=10):
+    '''perform Sholl frequency calculations on a population of neurites
+
+    Args:
+        nrn(morph): nrn or population
+        neurite_type(NeuriteType): which neurites to operate on
+        step_size(float): step size between Sholl radii
+
+    Note:
+        Given a neuron, the soma center is used for the concentric circles,
+        which range from the soma radii, and the maximum radial distance
+        in steps of `step_size`.  When a population is given, the concentric
+        circles range from the smallest soma radius to the largest radial neurite
+        distance.  Finally, each segment of the neuron is tested, so a neurite that
+        bends back on itself, and crosses the same Sholl radius will get counted as
+        having crossed multiple times.
+    '''
+    nrns = neuron_population(nrn)
+    neurite_filter = is_type(neurite_type)
+
+    min_soma_edge = float('Inf')
+    max_radii = 0
+    neurites_list = []
+    for neuron in nrns:
+        neurites_list.extend(((neurites, neuron.soma.center)
+                              for neurites in neuron.neurites
+                              if neurite_filter(neurites)))
+
+        min_soma_edge = min(min_soma_edge, neuron.soma.radius)
+        max_radii = max(max_radii, np.max(np.abs(bounding_box(neuron))))
+
+    radii = np.arange(min_soma_edge, max_radii + step_size, step_size)
+    ret = np.zeros_like(radii)
+    for neurites, center in neurites_list:
+        ret += sholl_crossings(neurites, center, radii)
+
+    return ret
--- a/benchmarks/benchmarks.py
+++ b/benchmarks/benchmarks.py
@@ -107,6 +107,9 @@ class TimeFeatures(object):
     def time_principal_direction_extents(self):
         nm.get('principal_direction_extents', self.neuron)
 
+    def time_sholl_frequency(self):
+        nm.get('sholl_frequency', self.neuron)
+
 
 class TimeChecks:
     def setup(self):
--- a/neurom/fst/__init__.py
+++ b/neurom/fst/__init__.py
@@ -96,6 +96,7 @@ NEURONFEATURES = {
     'trunk_origin_azimuths': _nrn.trunk_origin_azimuths,
     'trunk_origin_elevations': _nrn.trunk_origin_elevations,
     'trunk_section_lengths': _nrn.trunk_section_lengths,
+    'sholl_frequency': _nrn.sholl_frequency,
 }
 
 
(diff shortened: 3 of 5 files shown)


mgeplf:
Could you please try out this diff, and see if it makes sense for you?

This is available through morph_stats by using the following config:
```
neuron:
    sholl_frequency: [max, min, mean, median, std, total]
```

It can also be used directly if you want to plot things, for instance (note, this is using the dev API, which is subject to change):
```
import matplotlib.pylab as plt
import numpy as np
import neurom as nm
from neurom.fst._neuronfunc import sholl_crossings

morph = nm.load_neuron('C010306C.h5')
radii = np.arange(1, np.max(np.abs(nm.geom.bounding_box(morph))), 10)
count = sholl_crossings(morph, morph.soma.center, radii)

fig, ax = plt.subplots()
ax.bar(radii, count, color='r')
fig.savefig('sholl.png')
```

a1eko:
Can't say about your h5 test file but it gives reasonable numbers for my swc morphologies. So, yes, it works!

mgeplf:
Great! Appreciate the feedback.

neurom/fst/_neuronfunc.py
```diff
@@ -133,3 +135,68 @@ def _elevation(section, soma):
...
+    '''
+    def _count_crossings(neurite, radius):
+        '''count_crossings of segments in neurite with radius'''
+        r2 = radius ** 2
+        count = 0
+        for start, end in iter_segments(neurite):
+            start_dist2, end_dist2 = (morphmath.point_dist2(center, start),
+                                      morphmath.point_dist2(center, end))
+            count += int(start_dist2 <= r2 <= end_dist2 or
+                         end_dist2 <= r2 <= start_dist2)
```
lidakanari: My impression is that this counts the crossings of each segment with each circle. The only problem here is what happens if the section would bend back. In this case we would have two crossings (for two segments) for the same section. This might be rare and even unimportant but I think it will worth to note it in the documentation?

mgeplf: That is counted as two crossings.  From the links you provided before, my impression was that any crossings counted, not just the first one.  Do you have examples where only the first ones are counted?

mgeplf: Hmm, I think if they were treated different, it would be even more complicated to understand what it meant.

Either way, I can add a note to the docstring, no problem.

lidakanari:
I tested the Sholl analysis of this Neuromorpho.org cell from LNMC lab:
http://neuromorpho.org/neuron_info.jsp?neuron_name=OG061123A1-4-IDA
against the Neurolucida output for the same cell, as produced by Ying Shi.

Please find attached the results and the exact morphology used.
There are a few inconsistencies which seem to persist for multiple cells, 
but they are minor and this might also be related to loading inconsistencies.

Here's the code used:
```python
import neurom as nm
from neurom.fst._neuronfunc import sholl_crossings

n1 = nm.load_neuron('./OG061123A1-4-IDA.ASC')
tr = n1.neurites[0]

radii = np.arange(0, np.max(np.abs(nm.geom.bounding_box(tr))), 20)[1:]
count = sholl_crossings(tr, n1.soma.center, radii)
```
[Check_Sholl_analysis.zip](https://github.com/BlueBrain/NeuroM/files/655407/Check_Sholl_analysis.zip)

mgeplf:
Thanks.  These are the differences I'm seeing:
```
R     NL NM
300   8  10
1240  4   2
```
Out of 87, that's a pretty good start.

WRT your comments about segments vs. sections, from what I understand, you're saying that only the *first* crossing of a segment is counted.  Is that correct? IE, in the following picture, only the red points of intersection are counted, and not the blue ones?
sholl

lidakanari:
Yes, indeed! Almost there :)

Sure, your description is right. However, I am not sure that this change will resolve the issue.

mgeplf:
Could you find out what Neurolucida is using as its origin point for the sholl spheres for this morphology?  That would help to track down the problem.  For instance, the difference at the R=1240 probably comes from the little branch near the intersection in this diagram (green dot in the lower right):
sholl_spheres

lidakanari:
Sorry but I don't have access to NL (not even to use it) and they are quite cryptic with their descriptions. I will refer the question to the corresponding people next week...

lidakanari:
Here are the comments from the Neurolucida expert (Ying):
* The segment intersections are indeed what is counted (so ignore my previous comment about sections)
* The analysis starts from the center of the soma (it is not clear how the center of the soma is computed)
* The Sholl analysis probably refers to 2D circles instead of spheres. This is not clear from the documentation of neurolucida, but that is how it was represented in the visual environment of NL. If someone is more familiar with NL it would be useful to check this...

mgeplf:
From the neurolucida documentation I have access to, it sounds like it's using spheres `This means that the smallest radius row is a sphere of the given radius`:
http://www.mbfbioscience.com/help/nx11/Content/Analyses/Sholl_Analysis.htm

I was hoping that we could get the centroid that is calculated by hovering over the results, or asking Neurolucida to generate the centroid for the cell body, and noting it.

mgeplf:
To get this merged, I would really like to know what the differences at Sholl Radii 300, and 1240 are, as described in https://github.com/BlueBrain/NeuroM/pull/570#issuecomment-267542011

If neurolucida could highlight the cross at these two radii, and an image saved, I could use that to confirm my thoughts in https://github.com/BlueBrain/NeuroM/pull/570#issuecomment-267580746

Maybe @chalimou can see with Ying?
 
Alternatively, if someone could try another tool that does Sholl analysis, we could compare those results.

chalimou:
Yes, I'll see with Ying as soon as she comes back (in a week) and check for the following:
1. Do NL and NM use the same starting point for the spheres?
2. As @mgeplf requested some visual input at radii 300 and 1240.
@lidakanari just as a reminder: If you have one or two other cells with inconsistencies of the kind would be helpful.
@mgeplf and @lidakanari if you have further questions for Ying, just let me know.

chalimou:
Ying and I reproduced the results of the Sholl analysis of the axon of the cell OG061123A1-4-IDA.ASC in Check_Sholl_analysis.zip done by @lidakanari end of last year and tried to find the reason for the differences between NL (Neurolucida) and NM (NeuroM).

The visualisation of the Sholl analysis results given by Neurolucida (NL) could not help because the figures NL gives for the Sholl analysis are very imprecise. 
- First, they are 2d projections, although what is computed is the number of intersections with spheres.
- Second, the resolution of the neurites is low. The neurites are much more tortuous than in the Sholl analysis figures (see last Figure, also from Neurolucida).
- Third, Neurolucida does not give explicitly the coordinates of the spheres center (the so-called reference point). Even if the coordinates are computed the same way as in NM, there can be small numerical differences, which, in combination with the tortuosity of the neurites may lead to the observed differences between the two applications.

My conclusion is that both applications (NL, NM) may be correct and consistent in themselves. The observed differences may result from differences in the computation of the reference point (the spheres center) in addition with the tortuosity of the neurite. This cannot be verified though, since the reference point coordinates in NL are not reported.

I include a small report with the details and the figures. The Sholl analysis figures start with radius 260 um and 1200 um respectively and proceed in 20 um steps, exactly as the analysis.


[ShollAnalysis.zip](https://github.com/BlueBrain/NeuroM/files/1041472/ShollAnalysis.zip)

chalimou:
Ying and I reproduced the results of the Sholl analysis of the axon of the cell OG061123A1-4-IDA.ASC in Check_Sholl_analysis.zip done by @lidakanari at the end of last year. We try to pinpoint the reported differences between Neurolucida (NL) and NeuroM (NM).

The visualization of the Sholl analysis results as given by NL does not help since the figures are very imprecise. 
- They are 2d projections, although what is reported are the numbers of intersections with spheres.
- The resolution of the neurites is very low on these figures. The neurites are much more tortuous than shown (see last figure, also from Neurolucida).
- NL does not give explicitly the coordinates of the center of the spheres (the so-called reference point). Even if its coordinates are computed in the same way like in NM, there can be small numerical differences. The latter in combination with the tortuosity of the neurites can lead to the observed differences between the two applications.

My impression is, that both, NL and NM can be correct and consistent in themselves. The observed differences in the intersections numbers may result from differences in the center coordinates together with the toruosity of the neurites. This cannot be verified though, since the reference (center) point coordinates are not given in NL.

I include a small report and the figures. The Sholl analysis figures start with radius 260 um and 1200 um respectively and proceed in 20 um steps, exactly as the analysis does.
[ShollAnalysis.zip](https://github.com/BlueBrain/NeuroM/files/1041594/ShollAnalysis.zip)

mgeplf:
@chalimou Thanks for looking into that, it's really appreciated.

Sorry about the delay on merging this, but it seems we've come to the conclusion that it is creating results that make sense for those who understand morphometrics (@lidakanari, @a1eko).  If that is indeed the case, let me know and I can merge this today.

chalimou:
Thanks @mgeplf, I think you can safely merge your work on Sholl analysis.

lidakanari:
Thanks @chalimou and @mgeplf for looking into this. In my opinion, and since we cannot have more information on this issue, the code is doing what it is expected to do. I would be happy to have this code merged with a note that communicates the possible inconsistency with Neurolucida to the user. If someone comes with more information we can reconsider the implementation, but for the moment I can't see anything wrong with it.

elisabettai:
I comment here because I'm also interested in using Sholl analysis, as discussed with @lidakanari and @chalimou. I'm not very expert in morphometrics yet, so I don't have any relevant comment right now on the inconsistencies with Neurolucida. Thanks @mgeplf for merging it in master.

PR #571, opened 2016-12-13
tests failing due to imprecision, make them less exact

mgeplf:
* On some systems, the results aren't exactly the same, have those tests use 'close', when appropriate.


neurom/fst/tests/test_feature_compat.py
```diff
@@ -132,13 +132,6 @@ def test_get_section_path_distances(self):
         pl = [_sec.section_path_length(s) for s in i_chain2(self.sec_nrn_trees)]
         _close(pl, get('section_path_distances', self.ref_nrn))
 
-    @nt.nottest
```
mgeplf: this test isn't needed any more

Issue #567, opened 2016-12-07
Principal direction extents for a neuron as a whole


eleftherioszisis:
It would be useful to be able to calculate the principal direction extents for a whole neuron.

Right now the neurites are treated separately, calculating the principal components and the projection of points upon the eigenvectors of the covariance matrix of the points that constitute the neurite. 

However, sometimes one would like to calculate the extent of a neuron as a whole.

Thus, the feature could calculate the pcds for a neuron if a neuron is passed and for separete neurites if a set of neurites is passed.

mgeplf:
At the developer API level, you can do this:
```
import neurom as nm
import neurom.morphmath as mm
from neurom.core.dataformat import COLS

foo = nm.load_neuron('foo.swc')
mm.principal_direction_extent(foo.points[:, XYZ])
```
Is that what you expect?  This includes the soma points, but I guess you may want this for the 'neuron as a whole'

eleftherioszisis:
Yes, this is what I would expect. I just think that it would be useful to be able to differentiate between neurons and neurites from the feature getter level of this feature.

asanin-epfl:
@eleftherioszisis Do you still need this? If yes then am I correct that you want to have a feature `principal_direction_extents` at neuron level? Currently there is `principal_direction_extents` in `neuritefunc` for neurites.

arnaudon:
take a look around here: https://github.com/BlueBrain/morph-tool/blob/master/morph_tool/transform.py#L179

PR #572, opened 2016-12-13
Use a decorator to register features

mgeplf:
* This makes it easier to add features, and have aliases for them
* Will make it easier to add descriptions of the return types, in
  the future

### DIFF
--- a/neurom/core/features.py
+++ b/neurom/core/features.py
@@ -1,0 +1,136 @@
+# Copyright (c) 2016, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
+# All rights reserved.
+#
+# This file is part of NeuroM <https://github.com/BlueBrain/NeuroM>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#     1. Redistributions of source code must retain the above copyright
+#        notice, this list of conditions and the following disclaimer.
+#     2. Redistributions in binary form must reproduce the above copyright
+#        notice, this list of conditions and the following disclaimer in the
+#        documentation and/or other materials provided with the distribution.
+#     3. Neither the name of the copyright holder nor the names of
+#        its contributors may be used to endorse or promote products
+#        derived from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+# ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+'''Feature registrationg and retrieval'''
+
+import numpy as np
+
+from functools import partial
+from ..exceptions import NeuroMError
+from .types import tree_type_checker
+from . import iter_neurites
+from . import NeuriteType
+
+
+FEATURES = {'NEURITEFEATURES': {},
+            'NEURONFEATURES': {}, }
+
+
+def register_neurite_feature(name, func):
+    '''Register a feature to be applied to neurites
+
+    Parameters:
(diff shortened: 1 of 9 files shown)


neurom/core/features.py
```diff
@@ -0,0 +1,136 @@
...
+from .types import tree_type_checker
+from . import iter_neurites
+from . import NeuriteType
+
+
+FEATURES = {'NEURITEFEATURES': {},
+            'NEURONFEATURES': {}, }
+
+
+def register_neurite_feature(name, func):
```
arsenius7: Can't see where this one is used.

neurom/core/features.py
```diff
@@ -0,0 +1,136 @@
...
+        raise NeuroMError('Attempt to hide registered feature %s' % name)
+
+    FEATURES[namespace][name] = func
+
+
+def feature(func=None, namespace=None, names=''):
+    '''feature decorator to automatically register the feature in the appropriate namespace'''
+
+    if func is None:
+        return partial(feature, namespace=namespace, names=names)
```
arsenius7: what is that?

neurom/core/features.py
```diff
@@ -0,0 +1,136 @@
...
+        name(string): name of the feature, used for access via get() function.
+        func(callable): single parameter function of a neurite.
+    '''
+    if name in FEATURES[namespace]:
+        raise NeuroMError('Attempt to hide registered feature %s' % name)
+
+    FEATURES[namespace][name] = func
+
+
+def feature(func=None, namespace=None, names=''):
```
arsenius7: parameter name 'aliases' would convey the meaning better.
And why is it a space-separated string rather than list?

neurom/core/features.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def _indent(string, count):
+    '''indent `string` by `count` * INDENT'''
+    indent = _INDENT * count
+    ret = indent + string.replace('\n', '\n' + indent)
+    return ret.rstrip()
+
+
+def get_doc():
```
arsenius7: That's a lot of manual formatting going on.
I would look for a way to dump it to some hierarhy, and let pretty print yaml or whatever deal with indents.

neurom/core/features.py
```diff
@@ -0,0 +1,136 @@
...
+    Returns:
+        features as a 1D or 2D numpy array.
+    '''
+
+    for name in resolution_order:
+        if feature_name in FEATURES[name]:
+            feat = FEATURES[name][feature_name]
+            break
+    else:
+        raise NeuroMError('Unable to find feature: %s' % feature_name)
```
eleftherioszisis: If a feature exists in both resolution orders, then this would always return the ```NEURITEFEATURES``` one, right?
Why not checking the type of the object and looking up for the corresponding resolution order?

mgeplf: @eleftherioszisis this is so the behavior hasn't changed from the old version

neurom/core/features.py
mgeplf: it's user feature that was added here: #537

neurom/core/features.py
```diff
@@ -0,0 +1,136 @@
...
+import numpy as np
+
+from functools import partial
+from ..exceptions import NeuroMError
+from .types import tree_type_checker
+from . import iter_neurites
+from . import NeuriteType
+
+
+FEATURES = {'NEURITEFEATURES': {},
```
mgeplf: yeah, good point

neurom/core/features.py
```diff
@@ -0,0 +1,136 @@
...
+    '''
+    def _fun(neurites, neurite_type=NeuriteType.all):
+        '''Wrap neurite function from outer scope and map into list'''
+        return list(func(n)
+                    for n in iter_neurites(neurites, filt=tree_type_checker(neurite_type)))
+
+    register_feature('NEURITEFEATURES', name, _fun)
+
+
+def register_feature(namespace, name, func):
```
mgeplf: correct, I could keep a list of namespaces, and assert against that

neurom/core/features.py
mgeplf: good points, will change to that

neurom/core/features.py
mgeplf: trick to deal w/ the @feature vs. @feature() syntax

neurom/core/features.py
mgeplf: hmm, I like it due to not needing `feature_found` variable

neurom/core/features.py
mgeplf: I think that would be good for another commit

neurom/core/features.py
```diff
@@ -0,0 +1,136 @@
...
+    '''
+
+    for name in resolution_order:
+        if feature_name in FEATURES[name]:
+            feat = FEATURES[name][feature_name]
+            break
+    else:
+        raise NeuroMError('Unable to find feature: %s' % feature_name)
+
+    return np.array(list(feat(obj, **kwargs)))
```
eleftherioszisis: Well, provided that all the generators yield elements that can map to 1d arrays (i.e. no single numbers or nested elements), then I think it should be fine.

neurom/core/features.py
eleftherioszisis: My concern is that this will be wrong for features that apply to both resolution orders, such as the ```principal_direction_extents``` that I suggested in #567.

Unless the user passes ```NEURONFEATURES``` as the ```resolution_order``` or swap the elements in the tuple, the feature will be assumed as a neurite feature and treat the cell as a collection of separate neurites, which gets the specific feature for a neuron object.

My rationale when I created the two feature categories was to have the ability to apply some features in both resolution orders, if it is required.

wizmer:
Is this PR still relevant ? Should we merge it ?

mgeplf:
Good question, I don't really know.  I think some of the changes are useful, but the decorator to register features is probably not a good idea.  I also think we could drop the 'register your own features', b/c I doubt anyone has used that.

This was one of the main reasons for doing it:
https://github.com/BlueBrain/NeuroM/pull/572/files#r93192751

asanin-epfl:
closing in favor of #860

PR #577, opened 2016-12-15
create and use COLS.XYZ, COLS.XYZR as much as possible

mgeplf:
* less likely to make mistakes when referencing XYZ

### DIFF
--- a/neurom/core/_neuron.py
+++ b/neurom/core/_neuron.py
@@ -33,7 +33,8 @@ from copy import deepcopy
 import numpy as np
 
 from . import NeuriteType, Tree
-from ..morphmath import segment_area, segment_volume, section_length
+from neurom import morphmath
+from neurom.core.dataformat import COLS
 from neurom.utils import memoize
 from neurom._compat import filter, map, zip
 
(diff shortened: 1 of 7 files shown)


neurom/core/_soma.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, points):
     @property
     def center(self):
         '''Obtain the radius from the first stored point'''
```
liesbethvanherpe: This should probably be "obtain the center".

neurom/core/_neuron.py
```diff
@@ -162,9 +163,10 @@ def __init__(self, root_node):
     def points(self):
         '''Return unordered array with all the points in this neurite'''
         # add all points in a section except the first one, which is a duplicate
-        _pts = [v for s in self.root_node.ipreorder() for v in s.points[1:, :4]]
+        _pts = [v for s in self.root_node.ipreorder()
+                for v in s.points[1:, COLS.XYZR]]
```
mgeplf: Yeah, the language isn't as tight as I would like, but referencing points often includes the radius, so it's included.

neurom/core/_soma.py
mgeplf: yes

PR #578, opened 2016-12-15
Fix default aspect ratio for dendrogram

lidakanari:
This fixes #557

PR #579, opened 2016-12-15
Dendrogram plot corrections

lidakanari:
Further improvement as a follow-up to #578


neurom/view/view.py
```diff
@@ -574,7 +574,7 @@ def dendrogram(obj, show_diameters=True, new_fig=True, new_axes=True, subplot=Fa
     kwargs['xlabel'] = kwargs.get('xlabel', 'micrometers (um)')
     kwargs['ylabel'] = kwargs.get('ylabel', 'micrometers (um)')
     kwargs['no_legend'] = False
-    kwargs['aspect_ratio'] = "auto"
+    kwargs['aspect_ratio'] = kwargs.get('aspect_ratio', 'auto')
```
lidakanari: This ensures that if the user selected a different value it will overwrite the default.

PR #581, opened 2016-12-20
adding DOI badge

jdcourcol:
Adding DOI badge created by zenodo.
see https://guides.github.com/activities/citable-code/


jdcourcol:
You can look into https://zenodo.org/record/209498#.WFjf7XUrLmE to see the outcome.
You may want to change the citation authors name and affiliation since it is by default your github account.
Let me know if you have trouble doing it.

juanchopanza:
Minor comment: the author list should be consistent with [AUTHORS.md](https://github.com/BlueBrain/NeuroM/blob/master/AUTHORS.md). I'm not sure which one is the "correct" one, but it should be result of a process, not automatically generated.

jdcourcol:
@juanchopanza : yes, This one is automatically generated so I think it needs to be reviewed and edited in Zenodo to be compliant with more academic standards.

lidakanari:
I agree with @juanchopanza with one more suggestion from me: it would make sense to modify the Maintainer to MikeG and probably the corresponding contact email as well?

lidakanari:
I agree with @juanchopanza with one more suggestion from me: it would make sense to modify the Maintainer to MikeG and probably the corresponding contact email as well?

jdcourcol:
@lidakanari do you have access to zenodo to make modifications ? for instance I am not sure you want you name to be 'lidakanari' in the DOI .

lidakanari:
@jdcourcol I don't know how to modify this. Yes, it would be better to have my name instead.

PR #585, opened 2017-01-05
Partition asymmetry as defined in https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2868386/

lidakanari:
I also added a higher level test for both of the partition functions.

### DIFF
--- a/neurom/fst/_bifurcationfunc.py
+++ b/neurom/fst/_bifurcationfunc.py
@@ -37,6 +37,8 @@ def local_bifurcation_angle(bif_point):
     '''Return the opening angle between two out-going sections
     in a bifurcation point
 
+    We first ensure that the input point has only two children.
+
     The bifurcation angle is defined as the angle between the first non-zero
     length segments of a bifurcation point.
     '''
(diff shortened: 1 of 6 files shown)


neurom/fst/_bifurcationfunc.py
```diff
@@ -77,3 +77,19 @@ def bifurcation_partition(bif_point):
...
+
+def partition_asymmetry(bif_point):
+    '''Calculate the partition asymmetry at a bifurcation point
+    as defined in https://www.ncbi.nlm.nih.gov/pubmed/18568015
+
+    The number of nodes in each child tree is counted. The partition
+    is defined as the ratio of the absolute difference and the sum
+    of the number of bifurcations in the two daughter subtrees
+    at each branch point.'''
+    n = float(sum(1 for _ in bif_point.children[0].ipreorder()))
```
mgeplf: might want to assert that there are exactly 2 children, in case it's a multifurcation

neurom/fst/_bifurcationfunc.py
```diff
@@ -77,3 +77,19 @@ def bifurcation_partition(bif_point):
...
+    m = float(sum(1 for _ in bif_point.children[1].ipreorder()))
+    if n != m:
```
mgeplf: don't need this test since if n==m, then 0/(n+m) == 0

neurom/fst/tests/test_neuritefunc.py
```diff
@@ -207,6 +207,11 @@ def test_partition():
     assert_allclose(partition,
                     (1.0, 1.0))
 
+def test_partition_asymmetry():
```
mgeplf: we should probably create a separate 'SIMPLE' morphology that has more interesting properties than the symmetry we have w/ this one.  Don't need to do it now thought

neurom/fst/_bifurcationfunc.py
lidakanari: Sure. I guess that should be the case for all the partition functions though, right?

neurom/fst/_bifurcationfunc.py
lidakanari: You're right, thanks!

neurom/fst/tests/test_neuritefunc.py
lidakanari: I test the higher level functionality in "test_get_features.py". Maybe this should be moved here in a future commit.

neurom/fst/_bifurcationfunc.py
mgeplf: yeah, it would be worth adding, I would say, so the user can know if there are multifurcations that are being ignored.

lidakanari: Should I return anything else in case the point is a multifurcation or just exclude this point?

neurom/fst/_bifurcationfunc.py
```diff
@@ -51,45 +53,52 @@ def skip_0_length(sec):
 
         return cur
 
-    ch0, ch1 = (skip_0_length(bif_point.children[0].points),
-                skip_0_length(bif_point.children[1].points))
+    if len(bif_point.children) == 2:
```
mgeplf: This means that if it's not a bifurcation, it's silently ignored, which I would find surprising, especially since there isn't (currently) a guaranteed ordering: If the children list gets shuffled, you'd get different results since it's only the first two children that are looked at.

I think doing an `assert len(bif_point.children) == 2, 'A bifurcation must have only 2 children'` which would raise an exception when it finds a multifurcation.

Alternative, having a warning printed would work, if non-bifurcations come up often, but aren't a show-stopper.

lidakanari: Sure, that sounds much better! Thanks for the suggestion.

lidakanari:
The last commit partly fixes #586, but the bifurcation angle tests are still missing.

lidakanari:
Please don't merge yet. I need to simplify the tests further.

lidakanari:
OK, I think that now all cases and bifurcation functions are covered with the tests, so #586 should be fixed. @mgeplf let me know if this works.

PR #588, opened 2017-01-11
Cleanup view

mgeplf:


neurom/view/tests/test_view.py
```diff
@@ -55,16 +55,18 @@ def test_tree():
...
 
 
 def test_soma():
     fig, ax = view.soma(soma0)
     fig, ax = view.soma(soma0, outline=False)
-    fig, ax = view.soma(soma0, plane='wrong')
-    nt.ok_(ax == 'No such plane found! Please select one of: xy, xz, yx, yz, zx, zy.')
+
+    nt.assert_raises(AssertionError, view.tree, soma0, plane='wrong')
+
```
lidakanari: Cool, I didn't know I could do that so neatly!

mgeplf:
This shouldn't have changed any behavior.
Plotting before:
plot_before
and after:
plot_after
Which has no differences:
2d

neurom/view/tests/test_view.py
mgeplf: It raises an exception now, which I think makes this sort of check easier.

lidakanari: Yes, that is definitely easier to check.`

lidakanari:
Thanks @mgeplf! Seems good to me and certainly much cleaner.

neurom/view/view.py
```diff
@@ -167,14 +172,10 @@ def _seg_2d(seg):
     kwargs['title'] = kwargs.get('title', 'Tree view')
```
arsenius7: alternatively, kwargs.setdefault()

neurom/view/view.py
```diff
@@ -96,6 +84,27 @@ def get_default(variable, **kwargs):
...
+    if get_default('diameter', parameters):
+        # TODO: This was originally a numpy array. Did it have to be one?
+        scale = get_default('diameter_scale', parameters)
+        linewidth = [2 * segment_radius(s) * scale for s in iter_segments(tr)]
+        if not linewidth:
+            linewidth = get_default('linewidth', parameters)
+    return linewidth
+
+
 def tree(tr, plane='xy', new_fig=True, subplot=False, **kwargs):
```
arsenius7: I would rather list the optional parameters with their defaults explicitly in function signatures (with defaults as constants to avoid code duplication).

neurom/view/view.py
mgeplf: Yeah, good call.

neurom/view/view.py
mgeplf: You mean not having **kwargs, right?  In principle I agree, but in this case it's hard, because this is a 'thin' wrapper over matplotlib, and there are a lot of options for plotting (the full list is here: https://github.com/BlueBrain/NeuroM/blob/master/neurom/view/common.py#L54) so then the function signature gets so long that it's hard to even read.

We've tried coming up with a work around before, but have never come to a good conclusion on how to do it (within the constraint of wanting single line viewing.)

PR #591, opened 2017-02-09
New features needed for validation / synthesis

lidakanari:
Prototypes on new -much needed- features, please optimize them when appropriate :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ addons:
     - libhdf5-dev
 cache: pip
 install:
+  - pip install 'matplotlib==1.5.2'
   - pip install .
 script:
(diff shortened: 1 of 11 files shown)


lidakanari:
Any idea why this is failing (since it has to do with the plots which were not modified to my knowledge). Also why does it repeat previous commits that I merged from the upstream?

mgeplf:
Not sure, I will have to look into it.  Might be due to newer versions of dependent libraries.

lidakanari:
Thanks! Let me know if I have to fix something....

lidakanari:
@mgeplf thanks!

mgeplf:
#595 is a copy of this, including a small change, so the changes are included.

Issue #470, opened 2016-08-16
document view `final` option


juanchopanza:
There seems to be an option `final` which controls whether a plot is shown or not (??!!)
This is needs to be documented. It may also make sense to give it a more intuitive name.

@eleftherioszisis I am assigning you because I think you introduced this option. Apologies if I am mistaken!

mgeplf:
Any update on this?

wizmer:
The `final` option is no longer in the codebase

PR #601, opened 2017-03-01
Cleanup view code

mgeplf:
* most plot_* functions don't return the same (figure, axis)
  that they received, since this suggests it's a copy, not the same one

* fix tests to not use a global fig/ax, because the order of the test
  means it's hard to find regressions

* tests don't create directories that would be lost if there's
  an exception

* make coverage rest itself on each nosetest runs, so things aren't
  cumulative

* reduce locations of matplotlib import, so there are fewer places to
  fix once the backend handling is done properly

### DIFF
--- a/neurom/view/common.py
+++ b/neurom/view/common.py
@@ -336,7 +336,7 @@ def save_plot(fig, **kwargs):
 
     output = os.path.join(output_path, prefile + output_name + postfile + "." + output_format)
 
-    plt.savefig(output, dpi=dpi, transparent=transparent)
+    fig.savefig(output, dpi=dpi, transparent=transparent)
 
 
 def plot_style(fig, ax, **kwargs):
@@ -371,7 +371,7 @@ def plot_style(fig, ax, **kwargs):
         fig.set_tight_layout(True)
 
     if output_path is not None:
-        fig = save_plot(fig=ax, **kwargs)
+        save_plot(fig=fig, **kwargs)
 
     show_plot = kwargs.get('show_plot', True)
     final = kwargs.get('final', False)
@@ -381,11 +381,8 @@ def plot_style(fig, ax, **kwargs):
         plt.show()  # pragma no cover
 
 
-def plot_title(fig, ax, **kwargs):
-
-    """
-    Function that defines the title options
-    of a matplotlib plot.
+def plot_title(_, ax, **kwargs):
+    """Function that defines the title options of a matplotlib plot.
 
     Parameters:
         fig: matplotlib figure
@@ -423,11 +420,8 @@ def plot_title(fig, ax, **kwargs):
                  fontsize=title_fontsize, **title_arg)
 
 
-def plot_labels(fig, ax, **kwargs):
-
-    """
-    Function that defines the labels options
-    of a matplotlib plot.
+def plot_labels(_, ax, **kwargs):
+    """ Function that defines the labels options of a matplotlib plot.
 
     Parameters:
         fig: matplotlib figure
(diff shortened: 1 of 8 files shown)


neurom/viewer.py
```diff
@@ -40,27 +40,21 @@
...
-    'neuron_2d': draw_neuron,
-    'neuron_dendrogram': draw_dendrogram,
-    'tree_3d': draw_tree3d,
-    'tree_2d': draw_tree,
-    'tree_dendrogram': draw_dendrogram,
-    'soma_3d': draw_soma3d,
-    'soma_2d': draw_soma
+    'neuron_3d': neuron3d,
+    'neuron_2d': neuron,
+    'neuron_dendrogram': dendrogram,
```
lidakanari: tree dendrogram and neuron dendrogram seem to be the same. Is this the expected behavior? If that is the case, why do we need two types of dendrogram calls that do the same thing?

neurom/view/common.py
```diff
@@ -353,55 +348,37 @@ def plot_style(fig, ax, **kwargs):
...
 
+    tight = kwargs.get('tight', False)
     if tight:
         fig.set_tight_layout(True)
 
     if output_path is not None:
         fig = save_plot(fig=ax, **kwargs)
 
+    show_plot = kwargs.get('show_plot', True)
+    final = kwargs.get('final', False)
```
lidakanari: Maybe fix according to https://github.com/BlueBrain/NeuroM/issues/470 ?

neurom/viewer.py
mgeplf: This is the same behavior as before.  The code concatenates the obj type (either neuron or tree) with the 'type' (ie: dendrogram) to produce 'neuron_dendrogram' or 'tree_dendrogram', which is why it looks the way it does

neurom/view/common.py
mgeplf: I don't understand what it's trying to do, and I don't see what the resolution is in #470.  I think it should probably be removed, as 'show_plot' controls similar behavior.

lidakanari: I'm not sure either :) Maybe @eleftherioszisis can help on this?

neurom/view/common.py
```diff
@@ -381,11 +381,8 @@ def plot_style(fig, ax, **kwargs):
         plt.show()  # pragma no cover
```
mgeplf: I think this usage of plt. (and the one above) might not work how people expect: if the figure passed in to the function is the 'active' one in matplotlibs internal state, the wrong diagram will be shown...
Any suggestions: @lidakanari or @eleftherioszisis

neurom/view/common.py
```diff
@@ -381,11 +381,8 @@ def plot_style(fig, ax, **kwargs):
...
 
 
-def plot_title(fig, ax, **kwargs):
-
-    """
-    Function that defines the title options
-    of a matplotlib plot.
+def plot_title(_, ax, **kwargs):
```
mgeplf: each of the plot_ functions take the 'fig', but then never use it.  For not I have given it an un-named  arg, but I don't think we need to pass them in.

I would prefer to remove them for functions that don't use them.

Any comments @lidakanari, @eleftherioszisis

lidakanari: It makes sense. I think you are right to assume that it is not used in those functions because for >90% of the matplotlib operations the axes object is sufficient. However, in some cases we need to manipulate the figure in the output (that's why I was passing this as a argument). I would propose to keep it in the plot_style function but I think we could remove it from the rest especially if it is indeed not used.

Issue #600, opened 2017-03-01
Three point somas are being used for .h5 files


mgeplf:
h5dump of `test_data/h5/v1/Neuron.h5` gives:
```
      (0,0): 0, 0, 0, 0.2,
      (1,0): 0, 0.2, 0, 0.2,
      (2,0): 0.1, 0.1, 0, 0.2,
```
When loading it, it's interpreted as a 3 point soma:
```
p = 'test_data/h5/v1/Neuron.h5'
import neurom as nm
m = nm.load_neuron(p)
m.soma
<neurom.core._soma.SomaThreePoint at 0x7fcac410dad0>
```

It really should be a pure contour.

mgeplf:
Need to also check for multiple sections in h5 files

PR #605, opened 2017-03-07
fix SWC soma handling

mgeplf:
* before, the interpretation of the points composing the soma was
  a contour, however, after discussion/research, it was decided
  that the points actually create a series of cylinders


neurom/apps/tests/test_morph_stats.py
```diff
@@ -51,7 +51,7 @@
 }
 
 REF_OUT = {
-    'mean_soma_radius': 0.17071067811865476,
+    'mean_soma_radius': 0.13065629648763766,
```
mgeplf: Because the soma type changed for the SWC files, all the area/volume comparisons have to be updated

neurom/core/_soma.py
```diff
@@ -27,30 +27,15 @@
...
-from neurom.morphmath import average_points_dist
+import logging
+import math
+
+from neurom import morphmath
 from neurom.core.dataformat import COLS
 from neurom.exceptions import SomaError
 import numpy as np
 
-
```
mgeplf: removed this to simplify, replaced w/ _get_type below

neurom/core/tests/test_soma.py
```diff
@@ -79,15 +79,6 @@
     [11, 22, 33, 44, 2, 1, 1]
 ]
 
-BIFURCATING_SOMA_PTS = [
```
mgeplf: wasn't used

neurom/io/tests/test_utils.py
```diff
@@ -147,28 +120,9 @@ def test_neuron_name():
         nt.eq_(nrn.name, nn)
 
 
-def test_load_contour_soma_neuron():
```
mgeplf: not applicable anymore, since swc is a cylinder

neurom/view/common.py
```diff
@@ -36,11 +36,17 @@
 #  TODO: Awful hack to use non-GUI backend when no display
 #  is available. For unixy systems.
 import matplotlib
-if 'DISPLAY' not in os.environ:  # noqa
-    matplotlib.use('Agg')  # noqa
+if 'DISPLAY' not in os.environ: # noqa
```
mgeplf: still need a better way to handle this, see #601

neurom/view/common.py
```diff
@@ -596,16 +591,94 @@ def plot_legend(ax, **kwargs):
...
+        ax: matplotlib axes
+        plane(tuple of int): where x, y, z = 0, 1, 2, so (0, 1) is the xy axis
+        start(np.array): start coordinates
+        end(np.array): end coordinates
+        start_radius(float): start radius
+        end_radius(float): end radius
+        color: matplotlib color
+        alpha(float): alpha value
+
+    Note: There are probably more efficient ways of doing this: here the
```
mgeplf: if you can think of better ways, let me know.  I will attach some examples of the output to the ticket

test_data/swc/soma/three_pt_soma.swc
```diff
@@ -1,9 +1,9 @@
 # Neuron with NeuroMorpho 3 point soma.
```
mgeplf: follow neuromorpho spec

neurom/view/common.py
mgeplf: XY: side view of simple cylinder:
xy

xz (bottom view of simple cylinder):
xz

XY: cylinder at an angle coming out from page: (from (0, 0, 0) -> (10, 10, 10))
xy_0_0_0_to_10_10_10

Side tapered cylinder:
xy_side_tapered

neurom/core/_soma.py
```diff
@@ -88,8 +72,95 @@ def __init__(self, points):
...
+        assert points[0, COLS.Y] - points[2, COLS.Y] == -points[0, COLS.R], \
+            'The second point must be one radius above 0 on the y-plane'
+
+        r = points[0, COLS.R]
+        h = points[2, COLS.Y] - points[1, COLS.Y]
+        self.area = 2.0 * math.pi * r * h  # ignores the 'end-caps' of the cylinder
+        self.radius = math.sqrt(self.area / (4. * math.pi))
+
+    @property
+    def center(self):
```
arsenius7: same as parent?

neurom/core/_soma.py
```diff
@@ -134,19 +207,53 @@ def center(self):
...
+            L.warning('Using neuromorpho 3-Point soma')
+            # NeuroMorpho is the main provider of morphologies, but they
+            # with SWC as their default file format: they convert all
+            # uploads to SWC.  In the process of conversion, they turn all
+            # somas into their custom 'Three-point soma representation':
+            #  http://neuromorpho.org/SomaFormat.html
+
+            return SomaThreePointCylinders
+
+        return {0: None,
```
arsenius7: why not raise a SomaError here?

neurom/fst/_core.py
```diff
@@ -129,12 +131,28 @@ def _check_soma_topology_swc(points):
...
+    'NL-ASCII': None,
 }
 
-_SOMA_ACTION = {
-    'SWC': _check_soma_topology_swc,
-    'H5V1': None,
-    'H5V2': None,
-    'NL-ASCII': None
+_SOMA_CONFIG = {
+    # after much debate (https://github.com/BlueBrain/NeuroM/issues/597)
```
arsenius7: I would say that a link to the issue is enough if these sources are already mentioned there

neurom/io/__init__.py
```diff
@@ -28,6 +28,4 @@
 
 ''' IO operations module for NeuroM '''
 
-from ..core.dataformat import COLS
-from ..core.dataformat import ROOT_ID
-from neurom.io.utils import load_data, load_neuron, load_neurons, NeuronLoader
+from neurom.io.utils import load_data, load_neuron, load_neurons
```
arsenius7: why removing NeuronLoader?

neurom/view/common.py
lidakanari: It looks reasonable to me! Is this plotting generalized to a collection of cylinders if the swc-soma is a set of cylinders?

neurom/core/tests/test_soma.py
lidakanari: This was added to exclude branchings within the soma (which are not allowed in any other type of soma except swc). I don't know why this test was removed but it would be good to have one as currently the valid set of neurons contains files that have issues because they have bifurcating somata.

neurom/core/_soma.py
```diff
@@ -134,19 +207,53 @@ def center(self):
...
+        points: Soma points
+        soma_class(str): one of 'contour' or 'cylinder' to specify the type
+    '''
+    assert soma_class in ('contour', 'cylinder', )
+    npoints = len(points)
+    if soma_class == 'contour':
+        return {0: None,
+                1: SomaSinglePoint,
+                3: SomaThreePoint,
+                2: None}.get(npoints, SomaSimpleContour)
```
lidakanari: Is the ThreePoint different from the SimpleContour? Do we need this one?

neurom/view/common.py
mgeplf: Yes, it's in the `neurom/view/view.py:soma` and `neurom/view/view.py:soma3d` code

neurom/core/_soma.py
mgeplf: fixed

neurom/core/_soma.py
mgeplf: For now, yes, the interpretation will probably change though

neurom/core/_soma.py
mgeplf: Seems easier to just use the dict lookup on point count, this way if we have something that has a 2 point soma, it's easy to find, for instance

neurom/core/tests/test_soma.py
mgeplf: h5 files should have bifurcations, since they are contours, I'm mainly fixing the SWC cylinder problem here

neurom/fst/_core.py
mgeplf: They weren't, I dug them up after

neurom/io/__init__.py
mgeplf: accident, fixing

neurom/core/tests/test_soma.py
lidakanari: Yes, the h5 soma is a contour but that means a _single_ section. Indeed the existence of bifurcations oppose to the h5 definition in https://bbpcode.epfl.ch/browse/code/common/morphology-documentation/
```
Soma: there is only one soma section
```
so there should be a test to ensure that bifurcations are not allowed (it doesn't have to be in this commit surely but at least there should be an issue for this).

mgeplf: I added it to #600

lidakanari: Thanks!

neurom/core/_soma.py
```diff
@@ -88,8 +72,95 @@ def __init__(self, points):
...
+
+        # X    Y     Z   R    P
+        # xs ys      zs rs   -1
+        # xs (ys-rs) zs rs    1
+        # xs (ys+rs) zs rs    1
+
+        # make sure the above invariant holds
+        assert points[0, COLS.R] == points[1, COLS.R] == points[2, COLS.R], \
+            'All radii must be the same'
+        assert points[0, COLS.Y] - points[1, COLS.Y] == points[0, COLS.R], \
```
arsenius7: floating point errors?

neurom/core/_soma.py
```diff
@@ -134,19 +207,53 @@ def center(self):
...
+            # somas into their custom 'Three-point soma representation':
+            #  http://neuromorpho.org/SomaFormat.html
+
+            return SomaThreePointCylinders
+
+        return {0: None,
+                1: SomaSinglePoint}.get(npoints, SomaCylinders)
+
+
+def make_soma(points, soma_check=None, soma_class='contour'):
```
arsenius7: constants for these ones?

arsenius7:
Seems OK for me from the stylistic point of view.
Can't say much about the maths behind it.

neurom/core/_soma.py
mgeplf: so, in *theory*, the spec suggests they should match exactly, but doing a tolerance check makes sense to me, so I'll add that.

neurom/core/_soma.py
mgeplf: yup, good call.  thanks

neurom/core/_soma.py
arsenius7: theory might work for the first one, I have strong doubts about the second one (involving arithmetics).

neurom/core/_soma.py
```diff
@@ -140,12 +140,17 @@ def __init__(self, points):
         # xs (ys-rs) zs rs    1
         # xs (ys+rs) zs rs    1
 
+        def tol_eq(p0, p1):
```
arsenius7: np.isclose?

neurom/core/_soma.py
```diff
@@ -206,21 +211,27 @@ def __str__(self):
...
     Args:
         points: Soma points
         soma_class(str): one of 'contour' or 'cylinder' to specify the type
     '''
-    assert soma_class in ('contour', 'cylinder', )
+    assert soma_class in (SOMA_CONTOUR, SOMA_CYLINDER)
+
     npoints = len(points)
-    if soma_class == 'contour':
+    if soma_class is SOMA_CONTOUR:
```
arsenius7: is 'is' intentional? I would still expect '==' for value comparison

neurom/core/_soma.py
mgeplf: sigh, yes...

neurom/core/_soma.py
arsenius7: I would say it could cause problems and I don't see why the values of these constants could be changed.
Up to you though.

mgeplf: I'll change it.

PR #608, opened 2017-03-10
WIP: Have plotting functions operate on axis

mgeplf:
* (neuron, neron3d, tree, tree3d, soma, soma3d, dendrite) in
  neurom/view/view.py all operate on matplotlib.Axes, so the user
  has more control over what is happening, and can compose plots
  more simply.

* They generally set the dataLim, so axes views can be inferred
  properly

### DIFF
--- a/examples/plot_somas.py
+++ b/examples/plot_somas.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 # Copyright (c) 2015, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
 # All rights reserved.
 #
--- a/examples/plot_somas.py
+++ b/examples/plot_somas.py
@@ -46,12 +46,11 @@ def random_color():
 
 def plot_somas(somas):
     '''Plot set of somas on same figure as spheres, each with different color'''
-    fig, ax = common.get_figure(new_fig=True, subplot=111,
-                                params={'projection': '3d', 'aspect': 'equal'})
+    _, ax = common.get_figure(new_fig=True, subplot=111,
+                              params={'projection': '3d', 'aspect': 'equal'})
     for s in somas:
         common.plot_sphere(ax, s.center, s.radius, color=random_color(), alpha=1)
-    fig.show()
-    plt.close(fig)
+    plt.show()
 
 
 if __name__ == '__main__':
--- a/neurom/core/_neuron.py
+++ b/neurom/core/_neuron.py
@@ -212,6 +212,11 @@ class Neurite(object):
         '''Deep copy of neurite object'''
         return Neurite(deepcopy(self.root_node, memo))
 
+    def __nonzero__(self):
+        return bool(self.root_node)
+
+    __bool__ = __nonzero__
+
     def __str__(self):
         return 'Neurite <type: %s>' % self.type
 
--- a/neurom/tests/test_viewer.py
+++ b/neurom/tests/test_viewer.py
@@ -27,12 +27,12 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import os
+import shutil
+import tempfile
 
 import matplotlib
 if 'DISPLAY' not in os.environ:  # noqa
     matplotlib.use('Agg')  # noqa
-
-import os
 
 from neurom.view import common
 from neurom import load_neuron
@@ -78,6 +78,7 @@ def test_draw_soma3d():
 
 
 def test_draw_dendrogram():
+    viewer.draw(nrn, mode='dendrogram')
     common.plt.close('all')
 
 
@@ -96,3 +97,18 @@ def test_invalid_object_raises():
 @nt.raises(viewer.NotDrawableError)
 def test_invalid_combo_raises():
     viewer.draw(nrn.soma, mode='dendrogram')
+
+
+def test_writing_output():
+    fig_name = 'Figure.png'
+
+    tempdir = tempfile.mkdtemp('test_viewer')
+    try:
+        old_dir = os.getcwd()
+        os.chdir(tempdir)
+        viewer.draw(nrn, mode='2d', output_path='subdir')
+        nt.ok_(os.path.isfile(os.path.join(tempdir, 'subdir', fig_name)))
+    finally:
+        os.chdir(old_dir)
+        shutil.rmtree(tempdir)
+        common.plt.close('all')
--- a/neurom/tests/test_viewer.py
+++ b/neurom/tests/test_viewer.py
@@ -28,6 +28,12 @@
 
 import os
 
+import matplotlib
+if 'DISPLAY' not in os.environ:  # noqa
+    matplotlib.use('Agg')  # noqa
+
+import os
+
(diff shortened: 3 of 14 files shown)


mgeplf:
Currently most of the tests are turned off.

The old neuron.viewer.draw works the same:
Old:
old
New:
new

I think the 3d view looks better (but I'm still having issues w/ a locked aspect ratio):
Old:
original_3d

New:
new_3d

The new code means it's easy to have good control over everything:
```
fig = plt.figure()
planes = ('xy', 'zy')
for i, plane in enumerate(planes):
    ax = fig.add_subplot(1, len(planes), i + 1)
    v.neuron(ax, m, plane=plane)
    common.plot_style(fig=fig, ax=ax, white_space=30)
    ax.set_title(plane)
    fig.set_tight_layout(True)
```
Results in:
multi

Which was possible before, but I think this is more clear, and offers more control over the axis.  Again, this is the develop API, so I think we should assume they are knowledgeable about matplotlib.

@lidakanari @eleftherioszisis thoughts?

lidakanari:
Looks good! Last part of your code is equivalent to:
```
common.plot_style(fig=fig, ax=ax, white_space=30, title=plane, tight=True)
```
I think the functionality is the same but it is probably easier to understand it :) -if you are of course knowledgeable about matplotlib-.

Is it possible to add back the control over the type of neurites that are plotted? For instance to choose to plot only axons.

examples/plot_somas.py
```diff
@@ -46,12 +47,11 @@ def random_color():
...
     '''Plot set of somas on same figure as spheres, each with different color'''
-    fig, ax = common.get_figure(new_fig=True, subplot=111,
-                                params={'projection': '3d', 'aspect': 'equal'})
+    _, ax = common.get_figure(new_fig=True, subplot=111,
+                              params={'projection': '3d', 'aspect': 'equal'})
     for s in somas:
         common.plot_sphere(ax, s.center, s.radius, color=random_color(), alpha=1)
-    fig.show()
-    plt.close(fig)
+    plt.show()
```
mgeplf: Changed so figure shows up until it's closed..

neurom/core/_neuron.py
```diff
@@ -212,6 +212,11 @@ def __deepcopy__(self, memo):
         '''Deep copy of neurite object'''
         return Neurite(deepcopy(self.root_node, memo))
 
+    def __nonzero__(self):
```
mgeplf: Make it simple to test if a tree/neurite/neuron is empty

neurom/tests/test_viewer.py
```diff
@@ -27,6 +27,12 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import os
+import shutil
+import tempfile
+
+import matplotlib
```
mgeplf: Saves us from Agg being set by view.common for everyone.

neurom/view/common.py
```diff
@@ -26,20 +26,9 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-"""
-Module containing the common functionality
-to be used by view-plot modules.
-"""
-from neurom import NeuriteType
+"""Functionality for styling plots"""
```
mgeplf: Most of the changes in this file have to do w/:
* remove kwargs when possible, make all args explicit: better for completion, better for error handling, easier to write docstrings for

neurom/view/view.py
```diff
@@ -25,13 +25,12 @@
 # ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-'''
-Python module of NeuroM to visualize morphologies
-'''
+'''visualize morphologies'''
```
mgeplf: Changes have to do with:
* removing kwargs
* removing automatic figure creation: this is a development API, and it's better to give complete control to the person using matplotlib
* remove automatic styling, again, a developer can call this if necessary
* rename functions neuron -> plot_neuron

neurom/viewer.py
```diff
@@ -112,6 +121,16 @@ def draw(obj, mode='2d', **kwargs):
 
     viewer = '%s_%s' % (tag, mode)
     try:
-        return _VIEWERS[viewer](obj, **kwargs)
+        plotter = _VIEWERS[viewer]
+        output_path = kwargs.pop('output_path', False)
```
lidakanari: None should have the same effect, right?

neurom/view/common.py
```diff
@@ -50,542 +39,311 @@
...
     if zticks is not None:
         ax.set_zticks(zticks)
-        ax.zaxis.set_tick_params(labelsize=tick_fontsize, **zticks_arg)
+        zticks_args = dict_if_none(zticks_args)
+        ax.zaxis.set_tick_params(labelsize=tick_fontsize, **zticks_args)
 
 
-def plot_limits(ax, **kwargs):
+def plot_limits(ax, white_space):
     """Sets the limit options of a matplotlib plot.
```
lidakanari: Shouldn't we also control if someone wants to pass the limits directly? Sorry if this is done somewhere else but this review is too long and my memory limited :P

neurom/view/tests/utils.py
```diff
@@ -0,0 +1,24 @@
...
+from neurom.view.common import plt
+
+
+@contextmanager
+def get_fig_2d():
+    fig = plt.figure()
+    ax = fig.add_subplot(111)
+    ax.plot([0, 0], [1, 2], label='test')
+    yield fig, ax
+    plt.close(fig)
```
mgeplf: Good point, I will fix that.

neurom/viewer.py
mgeplf: fixed

neurom/viewer.py
```diff
@@ -112,6 +121,16 @@ def draw(obj, mode='2d', **kwargs):
...
+        plotter(ax, obj, **kwargs)
+
+        if mode != 'dendrogram':
+            common.plot_style(fig=fig, ax=ax, **kwargs)
+
+        if output_path:
+            common.save_plot(fig=fig, output_path=output_path, **kwargs)
+
+        return fig, ax
     except KeyError:
```
mgeplf: agreed

neurom/view/common.py
mgeplf: Yeah, it's huge, and I apologize for that.

I should rename this function to more clearly state what it does, which I will do.  If we want another function to directly set the limits, I can add that, it just probably wouldn't be called by any of the internal functions, so I don't know if it's something people would use

lidakanari: OK, it is certainly trivial to call the limits, so I don't mind not including this.

PR #623, opened 2017-03-21
Fix incorrect summary information in checks w/ multiple files

### LINKED ISSUE #622: has_axon in CheckRunner returns always False
```python
from neurom.check.runner import CheckRunner

config = {'checks': {'neuron_checks': ['has_axon']}}

check_runner = CheckRunner(config)

results = check_runner.run(directory_of_Neuron.swc)
```
returns False which is not true for the ```Neuron.swc``` in ```test_data```. Basically for every cell it returns False.

I tried the same with ```morph_check```, and that returned the expected correct values.

mgeplf:
* class member was used to store results, so the last result
  was being propagated to all the results

Fixes #622

### DIFF
--- a/neurom/check/runner.py
+++ b/neurom/check/runner.py
@@ -47,7 +47,6 @@ class CheckRunner(object):
     '''Class managing checks, config and output'''
     def __init__(self, config):
         self._config = CheckRunner._sanitize_config(config)
-        self.summary = OrderedDict()
         self._check_modules = dict((k, import_module('neurom.check.%s' % k))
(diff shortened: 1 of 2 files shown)


mgeplf:
@eleftherioszisis can you check if this fixes the problem for you?

eleftherioszisis:
The results now are in agreement with the morphologies. Thus, yes it is fixed indeed!

mgeplf:
I will have to do a release for pip upgrades to work, which I can do.

PR #627, opened 2017-04-05
Add surface area to neurite functions.

liesbethvanherpe:


neurom/fst/_neuritefunc.py
```diff
@@ -81,6 +81,14 @@ def n_leaves(neurites, neurite_type=NeuriteType.all):
     return n_sections(neurites, neurite_type=neurite_type, iterator_type=Tree.ileaf)
 
 
+def total_surface_area_per_neurite(neurites, neurite_type=NeuriteType.all):
```
mgeplf: to make this a feature that's available from the command line, it has to be named in
neurom/fst/__init__.py, which I think would be a good idea

neurom/fst/_neuritefunc.py
```diff
@@ -81,6 +81,14 @@ def n_leaves(neurites, neurite_type=NeuriteType.all):
...
+    '''Surface area in a collection of neurites.
+
+    The area is defined as the sum of the area of the sections.
+    '''
+    return [neurite.area for neurite in iter_neurites(neurites, filt=is_type(neurite_type))]
+
+
```
lidakanari: Maybe total_area_per_neurite to be consistent with the terminology in the code (section_area etc)?

neurom/fst/__init__.py
```diff
@@ -85,7 +85,8 @@
     'segment_taper_rates': _nrt.segment_taper_rates,
     'segment_radial_distances': _nrt.segment_radial_distances,
     'segment_meander_angles': _nrt.segment_meander_angles,
-    'principal_direction_extents': _nrt.principal_direction_extents
+    'principal_direction_extents': _nrt.principal_direction_extents,
+    'total_surface_area_per_neurite': _nrt.total_surface_area_per_neurite
```
mgeplf: not sure why principal_direction_extents didn't have a trailing comma, but could you add one to the line you added?

neurom/fst/_neuritefunc.py
liesbethvanherpe: I was inspired by the "soma_surface_area". It's the same to me - which one is preferred?

lidakanari: It's not the same computations between section and soma (I don't mind whichever is used) but it has to be consistent between section_area etc...

mgeplf: Sadly, we already have inconsistencies.  I think `total_area_per_neurite` makes sense

mgeplf:
Thanks, and thanks for the nice test

PR #628, opened 2017-05-01
Strahler order

lidakanari:
Add strahler branch order as defined here: https://en.wikipedia.org/wiki/Strahler_number.
Test branch order and strahler order functions with simple examples.


lidakanari:
Close to update main code to recent version.

PR #629, opened 2017-05-01
Strahler updated

lidakanari:
Add strahler branch order as defined here: https://en.wikipedia.org/wiki/Strahler_number.
Test branch order and strahler order functions with simple examples (branch order was not tested before at the sectionfunc level as far as I know).


neurom/fst/sectionfunc.py
```diff
@@ -72,6 +72,26 @@ def branch_order(section):
...
+    plus one to start counting from 1.
+
+    Note:
+        The leaves have branch order 1.
+    '''
+    if len(section.children) == 0:
+        return 1
+    else:
+        max_bo_children = max([branch_order(i) for i in section.ipreorder()])
+        return max_bo_children - branch_order(section) + 1
```
mgeplf: Is there a more efficient way this can be calculated? calling branch_order on each section could get expensive

neurom/fst/tests/test_neuritefunc.py
```diff
@@ -213,6 +213,12 @@ def test_section_branch_orders():
                     (0, 1, 1,  # type 3, basal dendrite
                      0, 1, 1)) # type 2, axon
 
+def test_section_strahler_orders():
+    strahler_orders = list(_nf.section_strahler_orders(SIMPLE))
+    assert_allclose(strahler_orders,
+                    (2, 1, 1,  # type 3, basal dendrite
```
mgeplf: I think a more complicated test would be helpful

neurom/fst/sectionfunc.py
```diff
@@ -72,6 +72,26 @@ def branch_order(section):
     return sum(1 for _ in section.iupstream()) - 1
 
 
+def strahler_order(section):
+    '''Branching order of a tree section
```
lidakanari: Yes, thanks!

neurom/fst/sectionfunc.py
```diff
@@ -72,6 +72,26 @@ def branch_order(section):
...
+    a section's descendants minus the section's own branch order
+    plus one to start counting from 1.
+
+    Note:
+        The leaves have branch order 1.
+    '''
+    if len(section.children) == 0:
+        return 1
+    else:
+        max_bo_children = max([branch_order(i) for i in section.ipreorder()])
```
lidakanari: OK, will fix!

neurom/fst/sectionfunc.py
lidakanari: I am sure but I am not aware of a way since I haven't found any doc for this. For the neurons I tested it takes almost the same time as branch order so it wasn't an issue. Maybe add it as a issue to optimize this in the future?

neurom/fst/tests/test_neuritefunc.py
mgeplf: I made a more complicated test based on the wikipedia article, and I'm getting differences in the Strahler number:
```
# SWC structure:
# index, type, x, y, z, radius, parent
# Recreation of:
# https://commons.wikimedia.org/wiki/File:Flussordnung_(Strahler).svg

 1 1  0 0 0 1 -1
 2 3  0 2 0 1  1
 3 3 -1 2 0 1  2
 4 3  0 3 0 1  2
 5 3 10 3 0 1  4
 6 3 10 2 0 1  5
 7 3  9 2 0 1  6
 8 3 11 2 0 1  6
 9 3 10 4 0 1  5
10 3  9 4 0 1  9
11 3 11 4 0 1  9
12 3  0 4 0 1  4
13 3 -1 4 0 1 12
14 3  0 5 0 1 12
15 3 -5 5 0 1 14
16 3 -5 6 0 1 15
17 3 -5 4 0 1 15
18 3  5 5 0 1 14
19 3  5 6 0 1 18
20 3  5 4 0 1 18
```
Loaded with:
```
>>> n = nm.load_neuron('path/to/above.swc')
>>> sectionfunc.strahler_order(n.neurites[0].root_node)
6
```
which is higher than 4 in the wikipedia image.  Are there different interpretations of the Strahler number?

lidakanari: Thank you for the test! There seem to be something wrong with this figure in Wiki or I do not understand the definition :/ Could we meet tomorrow to check this (I'm not in Geneva today)?

lidakanari: The definition of this ordering is not consistent (see below). I will try to find the original paper for a proper reference.

Neurolucida definition:
```
    Order assignment progresses towards the root segment.
    At each node, if **all daughter** segments are assigned the same order, the parent segment is assigned the daughter segments order "+1."
    If the daughter segments are assigned different orders, the largest order is assigned to the parent segment.
```
Wiki definition:
```
    If the node is a leaf (has no children), its Strahler number is one.
    If the node has one child with Strahler number i, and all other children have Strahler numbers less than i, then the Strahler number of the node is i again.
    If the node has **two or more children** with Strahler number i, and no children with greater number, then the Strahler number of the node is i + 1.
```

lidakanari: After a discussion with @mgeplf we decided to use the Stahler order as defined here: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3928584/pdf/1755-7682-7-6.pdf 
I will modify the code accordingly and include this reference.

lidakanari:
@mgeplf I fail to see how I can efficiently parse the tree from the leaves backwards and keep the memory of the Strahler order, so that I don't have to compute it at each step. If you have any ideas please let me know.

christian-oreilly:
I think that the following code fix previous issues:

```
def strahler_order(section):
    '''Branching order of a tree section

    The strahler order is the inverse of the branch order,
    since this is computed from the tips of the tree
    towards the root.
    
    This implementation is a translation of the three steps described in
    Wikipedia (https://en.wikipedia.org/wiki/Strahler_number):
     - If the node is a leaf (has no children), its Strahler number is one.
     - If the node has one child with Strahler number i, and all other children
       have Strahler numbers less than i, then the Strahler number of the node 
       is i again.
     - If the node has two or more children with Strahler number i, and no 
       children with greater number, then the Strahler number of the node is 
       i + 1.

    No efforts have been invested in making it computationnaly efficient, but
    it computes acceptably fast on tested morphologies (i.e., no waiting time).
    '''

    if len(section.children) == 0:
        return 1
    else:
        child_orders = np.array([strahler_order(child) for child in section.children])
        max_so_children = max(child_orders)
        nb_max          = np.sum(child_orders == max_so_children)
        if nb_max == 1:
            return max_so_children
        else:
            return max_so_children+1
```

It gives more reasonable results on my use case, and it pass a test inspired by the one previously proposed by @mgeplf. The test has been checked as follow:

```
import neurom as nm
from neurom.fst import _neuritefunc as _nf
from numpy.testing import assert_allclose

test_file_str = """
# SWC structure:
# index, type, x, y, z, radius, parent
# Recreation of:
# https://commons.wikimedia.org/wiki/File:Flussordnung_(Strahler).svg

 1 1  0 0 0 1 -1
 2 3  0 2 0 1  1
 3 3 -1 2 0 1  2
 4 3  0 3 0 1  2
 5 3 10 3 0 1  4
 6 3 10 2 0 1  5
 7 3  9 2 0 1  6
 8 3 11 2 0 1  6
 9 3 10 4 0 1  5
10 3  9 4 0 1  9
11 3 11 4 0 1  9
12 3  0 4 0 1  4
13 3 -1 4 0 1 12
14 3  0 5 0 1 12
15 3 -5 5 0 1 14
16 3 -5 6 0 1 15
17 3 -5 4 0 1 15
18 3  5 5 0 1 14
19 3  5 6 0 1 18
20 3  5 4 0 1 18
"""

with open("test.swc", "w") as f:
    f.write(test_file_str)
    
SIMPLE = nm.load_neuron('test.swc')

def test_section_strahler_orders():
    strahler_orders = list(_nf.section_strahler_orders(SIMPLE))
    assert_allclose(strahler_orders,
                    [4, 1, 4, 3, 2, 1, 1, 2, 1, 1, 3, 1, 3, 2, 1, 1, 2, 1, 1])  
test_section_strahler_orders()
```

lidakanari:
From what I can remember this looks correct to me according to the definition of Strahler order in https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3928584/pdf/1755-7682-7-6.pdf

Thanks @christian-oreilly for the proposed implementation!

christian-oreilly:
@lidakanari My pleasure. Then I would propose that we (meaning you, because I don't have the right ;) ) reopen the issue and merge.

mgeplf:
Strahler order has been added, and will be included in the next release

christian-oreilly:
Great, thanks @mgeplf

PR #631, opened 2017-05-11
use IntEnum so for NeuriteType so they are sortable

### LINKED ISSUE #630: Support ordering for NeuriteType
Suppose I have a pandas DataFrame `df` with a column `type` of NeuriteType's.
Now if I do
``
df.groupby('type').size()
``
I get
``
unorderable types: NeuriteType() > NeuriteType()
``

Although it could be amended by specifying `sort=False` flag
``
df.groupby('type', sort=False).size()
``
I believe imposing an order on NeuriteType will do no harm.

mgeplf:
fixes #630

### DIFF
--- a/neurom/core/types.py
+++ b/neurom/core/types.py
@@ -28,11 +28,11 @@
 
 '''Type enumerations'''
 
-from enum import Enum, unique
+from enum import IntEnum, unique
 
 
 @unique
-class NeuriteType(Enum):
+class NeuriteType(IntEnum):
     '''Enum representing valid tree types'''
     undefined = 1
     soma = 2


mgeplf:
OrderedEnum doesn't exist in the enum34 backport we're using :(

arsenius7:
:sad:

arsenius7:
Something I have not noticed then:
```
print(pd.DataFrame({'x': [nm.AXON, nm.BASAL_DENDRITE]}))
   x
0  3
1  4
```
Previously we got:
```
   x
0  NeuriteType.axon
1  NeuriteType.basal_dendrite
```

mgeplf:
I think pandas is inferring the type as int, maybe force it to object?

arsenius7:
`x` dtype is `object` already; or you mean some extra casting?

mgeplf:
it's an int, no?

This works:
```
print(pd.DataFrame({'x': np.array([nm.AXON, nm.BASAL_DENDRITE], dtype=object)}))
```

arsenius7:
Sorry, removed the previous comment, as I was wrong there, it is an int indeed (I read the last line from `dtypes` output).
Anyway, looks like leaving the trap lying there.
Maybe we should just copy-paste OrderedEnum implementation if it is not available from the library?

mgeplf:
Good catch, btw.

Yeah, it's a trap.  I don't know enough about the OrderedEnum or the type coercion that numpy/pandas does to know if it's a solution.  Prob. should at least open a separate issue for this.

PR #634, opened 2017-06-23
add ability to selectively ignore exceptions in load_neurons

mgeplf:
* added -ignore-exceptions option to morph_stats, to allow for ignoring bad soma

### DIFF
--- a/apps/morph_stats
+++ b/apps/morph_stats
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-
 # Copyright (c) 2015, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
 # All rights reserved.
 #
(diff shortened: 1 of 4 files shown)


apps/morph_stats
```diff
@@ -84,32 +81,38 @@ def get_parser():
                         help=('Summary output file name, if it ends in .json, '
                               'a json file is created, if .csv, then a csv file'))
 
+    parser.add_argument('-I', '--ignored-exceptions', dest='ignored_exceptions', default='',
```
arsenius7: How would I know what to provide here?
Maybe, use `choices` here and remove `if hasattr` check on line 104?

mgeplf: I'll make the doc better.  I don't think choices handles python classes, does it?

arsenius7: str -> class dict?

mgeplf: Good call, simplified the logic too.  Thanks.

PR #636, opened 2017-07-17
add apical_point.py example

mgeplf:


examples/apical_point.py
```diff
@@ -0,0 +1,111 @@
...
+
+    if not apical:
+        L.warning('No apical found')
+        return None
+    elif 1 < len(apical):
+        L.warning('Too many apical dendrites')
+        return None
+
+    apical = apical[0]
+
```
wizmer: The distance is computed twice for the same points.
You could keep it in a dict:

`leaf_distances = {leaf: point_dist2(leaf.points[-1, COLS.XYZ], morph.soma.center) for leaf in pical.root_node.ileaf()}
max_distance2 = max(leaf_distances.values())
min_distance2 = max_distance2 * (1 - tuft_percent / 100.) ** 2
parents = [set(leaf.iupstream) for leaf, dist in leaf_distance.items() if dist > min_distances]
common_parents = set.intersection(*parents)
`

mgeplf: yeah, good point, memoization ftw.

I can't say this was written w/ performance in mind, tbh

wizmer:
I think apical_point.py should better live in morph_tool instead. What do you think @mgeplf ?

mgeplf:
yup, sounds good to me...this was always just a placeholder

PR #639, opened 2017-07-24
Fix having an extra 'default' section when the last row is a leaf

mgeplf:
* Rename Section -> to DataBlockSection so it's not confused with
  neurom.Section

### DIFF
--- a/neurom/core/dataformat.py
+++ b/neurom/core/dataformat.py
@@ -29,9 +29,13 @@
 '''Data format definitions'''
 
 
+_COL_COUNT = 7
+
+
 class COLS(object):
     '''Column labels for internal data representation'''
-    (X, Y, Z, R, TYPE, ID, P) = range(7)
+    COL_COUNT = _COL_COUNT
+    (X, Y, Z, R, TYPE, ID, P) = range(_COL_COUNT)
     XYZ = slice(0, 3)
     XYZR = slice(0, 4)
 
--- a/neurom/io/datawrapper.py
+++ b/neurom/io/datawrapper.py
@@ -54,6 +54,11 @@ class DataWrapper(object):
             ID(integer): unique integer given to each point, the `ROOT_ID` is -1
             P(integer): the ID of the parent
 
+        Args:
+            data_block: as defined above
+            fmt: File format designation, eg: SWC
+            sections: Already extracted sections, otherwise data_block will be used
+
         Notes:
             - there is no ordering constraint: a child can reference a parent ID that comes
               later in the block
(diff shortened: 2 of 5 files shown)


neurom/core/dataformat.py
```diff
@@ -29,9 +29,13 @@
 '''Data format definitions'''
 
 
+_COL_COUNT = 7
```
mgeplf: cleanup

neurom/io/datawrapper.py
```diff
@@ -168,31 +177,30 @@ def _extract_sections(data_block):
             curr_section.ids.append(row_id)
 
         if row_id in sec_end_pts:
-            parent_section[curr_section.ids[-1]] = len(_sections) - 1
-            _sections.append(Section())
-            curr_section = _sections[-1]
+            parent_section[curr_section.ids[-1]] = len(sections) - 1
             # Parent-child discontinuity section
             if gap:
+                sections.append(DataBlockSection())
```
mgeplf: this is where the problem is: an extra `DataBlockSection` was being created unnecessarily for the last row if it was a leaf

neurom/io/datawrapper.py
```diff
@@ -168,31 +177,30 @@ def _extract_sections(data_block):
...
-
-
-COL_COUNT = 7
-_, _, _, _, TYPE, ID, PARENT = range(COL_COUNT)
+    # TODO find a way to remove empty sections.  Currently they are
+    # required to maintain tree integrity.
+    return sections
 
 
 class BlockNeuronBuilder(object):
```
mgeplf: everything from here on is just cleanup

neurom/io/datawrapper.py
```diff
@@ -168,31 +177,30 @@ def _extract_sections(data_block):
...
             if gap:
+                sections.append(DataBlockSection())
+                curr_section = sections[-1]
                 curr_section.ids.extend((parent_id, row_id))
                 curr_section.ntype = int(row[COLS.TYPE])
-                _gap_sections.add(len(_sections) - 2)
+                gap_sections.add(len(sections) - 2)
+            elif row_id != len(data_block) - 1:
+                # avoid creating an extra DataBlockSection for last row if it's a leaf
+                sections.append(DataBlockSection())
```
arsenius7: Does it worth making these two lines a helper method or removing curr_section altogether?

mgeplf: I thought about that when writing it.  I think I'm going to rewrite some of this routine and its dependencies in a more 'numpy' way, so I'll do it then, if that's ok.

PR #646, opened 2017-09-15
Cut plane detection

wizmer:
Adding a tool to search for the cut plane

The idea is to look at the distribution of all neuron data points along the X, Y, Z axes and to look for the sharpest cut at the beginning/end of each distribution:

    1) The distribution of all points along X, Y and Z is computed
       and put into 3 histograms.

    2) For each histogram we look at the first and last empty bins
       (ie. the last bin before the histogram starts rising,
       and the first after it reaches zero again). Under the assumption
       that there is no cut plane, the posteriori probability
       of observing this empty bin given the value of the not-empty
       neighbour bin is then computed.
    3) The lowest probability of the 6 probabilities (2 for each axes)
       corresponds to the cut plane"""


wizmer:
Yes, I am currently writing them.
If there is no cut plane the computed probability will be higher than as usual.
Based on the morphologies I have, I can try to find a threshold above which to set a warning about the computed plane, or to return no plane at all.
I will try to reproduce your crash.

mgeplf:
@jdcourcol That error is due to how your python is installed: if you're using brew or fink (iirc), you get that error.  Can you try using the system python?

https://stackoverflow.com/questions/4130355/python-matplotlib-framework-under-macosx

mgeplf:
I should add that it's a mismatch between the matplotlib wheel that's being installed, and the how your python was compiled, so, in theory, you could recompile matplotlib (or install it from where you installed python) and that should work.

apps/cut_plane_detection
```diff
@@ -0,0 +1,100 @@
...
+from neurom import exceptions
+from neurom.io.utils import get_files_by_path
+from neurom.utils import NeuromJSON
+
+
+L = logging.getLogger(__name__)
+
+def get_parser():
+    '''Parse command line arguments'''
+    description = 'Cut plane detection: '
```
mgeplf: building strings using brackets is easier, imo:
```
description = ('asdf',
                       'bar',
                       'baz',
)
```

neurom/apps/cut_plane_detection.py
```diff
@@ -0,0 +1,176 @@
...
+import numpy as np
+import neurom as nm
+
+from neurom import viewer
+from neurom.core import Tree
+from neurom.core.dataformat import COLS
+
+L = logging.getLogger(__name__)
+
+'''Module for the cut plane detection'''
```
mgeplf: usually have this above the imports

neurom/apps/cut_plane_detection.py
```diff
@@ -0,0 +1,176 @@
...
+
+def _create_1d_distributions(points, bin_width):
+    '''Create point distribution integrated along each axis
```
mgeplf: `'''Create point distribution integrated along each axis`
->
`'''Create point histogram along each axis`
?

neurom/apps/cut_plane_detection.py
```diff
@@ -0,0 +1,176 @@
...
+
+def _get_probabilities(hist):
+    '''Returns -log(p) where p is the a posteriori probabilities of the observed values
+    in the bins X min, X max, Y min, Y max, Z min, Z max
+
+    Parameters:
+        hist: a dict of the X, Y and Z 1D histograms
+
+    Returns: a dict of -log(p) values'''
+    def compute_p(h):
```
mgeplf: more descriptive variable names are preferred over single letter ones except for indices, and the 'special' case of x, y, z axis

neurom/apps/cut_plane_detection.py
```diff
@@ -0,0 +1,176 @@
...
+        '''Compute the a posteriori probability to observe "k" counts
+        in bin given than the mean value was "mu"
+        The mean value is approximated by the number of counts in the previous bin.
+        The probability follows a Normal distribution'''
+        mu = h
+        sigma = np.sqrt(mu)
+        k = np.append(h[1:], [0])
+        proba = 1/sigma*np.exp(-0.5*np.power((k-mu)/sigma, 2))
```
mgeplf: does scipy.stats.norm supply this?

neurom/apps/cut_plane_detection.py
```diff
@@ -0,0 +1,176 @@
...
+        proba[np.where(np.isnan(proba))] = 1
+        return proba[-1]
+
+    prob = dict()
+    for plane, (histogram, bins) in hist.items():
+        prob['{} max'.format(plane)] = -np.log(compute_p(histogram))
```
mgeplf: would dict(x=(min_x, max_x), ...) make sense to allow it to be reused w/o having to use strings?

neurom/apps/cut_plane_detection.py
wizmer: totally: norm.pdf does

jdcourcol:
> @jdcourcol That error is due to how your python is installed: if you're using brew or fink (iirc), you get that error. Can you try using the system python?
> 
> https://stackoverflow.com/questions/4130355/python-matplotlib-framework-under-macosx

I changed my ~/.matplotlib/matplotlibrc with backend : TkAgg and that works. Thank you.

apps/cut_plane_detection
```diff
@@ -0,0 +1,117 @@
...
+
+    parser = argparse.ArgumentParser(description=description)
+
+    parser.add_argument('datapath', nargs='+', type=str,
+                        help='Path to a morphology data file or a directory')
+
+    parser.add_argument('-o', '--output', dest='output_file',
+                        help=('Output json file name'))
+
+    parser.add_argument('-w', '--width', type=int, default=3,
```
jdcourcol: Is the default 3 cherry-picked here ?

neurom/apps/cut_plane_detection.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Parameters:
+        points: a np.ndarray of points
+        bin_width: the bin width
+
+    Returns: a dict of the X, Y and Z 1D histograms'''
+    min_, max_ = np.min(points, axis=0), np.max(points, axis=0)
+    hist = dict()
+    for i, plane in enumerate('XYZ'):
+        bins_min = np.arange(min_[i], max_[i] + bin_width, bin_width)
```
jdcourcol: bins_min has greater number than bins_max here no ?

neurom/apps/cut_plane_detection.py
```diff
@@ -0,0 +1,191 @@
...
+        plt.legend()
+
+
+def _get_status(minus_log_p):
+    '''Returns ok if the probability that there is a cut plane is high enough'''
+    _THRESHOLD = 50
+    if minus_log_p < _THRESHOLD:
+        return 'The probability that there is in fact NO cut plane is high: -log(p) = {0} !'\
+            .format(minus_log_p)
+    else:
```
jdcourcol: the else is useless here.

neurom/apps/cut_plane_detection.py
```diff
@@ -0,0 +1,191 @@
...
+       of observing this empty bin given the value of the not-empty
+       neighbour bin is then computed.
+    3) The lowest probability of the 6 probabilities (2 for each axes)
+       corresponds to the cut plane"""
+
+    points = np.array([point
+                       for neurite in (neuron.neurites or [])
+                       for section in nm.iter_sections(neurite)
+                       for point in section.points])
+    if not len(points):
```
jdcourcol: if not points:

neurom/apps/cut_plane_detection.py
```diff
@@ -0,0 +1,191 @@
...
+
+from neurom import viewer
+from neurom.core import Tree
+from neurom.core.dataformat import COLS
+
+
+L = logging.getLogger(__name__)
+
+
+def _create_1d_distributions(points, bin_width):
```
jdcourcol: I am not sure to get why we go for the histogram here. why not just taking 2 slices of length width , one starting at the first point, and the other ending at the last point ?

apps/cut_plane_detection
wizmer: I tried values between 1 and 20 and it was the one giving best results on the 40 thalamus morphologies

neurom/apps/cut_plane_detection.py
wizmer: Yes, that's true. Strictly speaking, the algorithm does not need the whole histogram but it helps a lot to be able to see it by eye. That's why I kept it.

neurom/apps/cut_plane_detection.py
wizmer: bins_min corresponds to the binning used for computing the first slice
bins_max correspond to the binning used for computing the last slice.
The thing is that, if the binning starting at first point is used for the first and last slice, it produces a bias. Indeed, even if the point distribution is symetrical, the first slice will have more counts than the last slice. So to have perfect symetrical conditions, the first slice count must be using a binning starting at the first point found, and the last slice must be using a binning starting at the last point found.
Not sure I am being clear...
Anyway, I'll rename to binning_first_slice, binning_last_slice

neurom/apps/cut_plane_detection.py
wizmer: Indeed

neurom/apps/cut_plane_detection.py
jdcourcol: my bad, it is a np.array not a list

Makefile
```diff
@@ -62,7 +62,7 @@ run_pep8: $(VENV_INSTALLED)
 	$(VENV_BIN)/pep8 --config=pep8rc $(LINT_PYFILES) > pep8.txt
 
 run_pylint: $(VENV_INSTALLED)
-	$(VENV_BIN)/pylint --rcfile=pylintrc --extension-pkg-whitelist=numpy $(LINT_PYFILES) > pylint.txt
+	$(VENV_BIN)/pylint --rcfile=pylintrc --extension-pkg-whitelist=numpy $(LINT_PYFILES) #> pylint.txt
```
mgeplf: need to uncomment the redirect, surprised CI is passing

apps/cut_plane_detection
mgeplf: maybe say what the unit is, and also put that as a comment, so people know where it came from

neurom/apps/cut_plane_detection.py
```diff
@@ -0,0 +1,207 @@
...
+    in the bins X min, X max, Y min, Y max, Z min, Z max
+
+    Parameters:
+        hist: a dict of the X, Y and Z 1D histograms
+
+    Returns: a dict of -log(p) values'''
+    def minus_log_p(mu):
+        '''Compute -Log(p) where p is the a posteriori probability to observe 0 counts
+        in bin given than the mean value was "mu":
+        The number of counts follows a Poisson law so the result is simply... mu
```
mgeplf: the counts is assumed to follow Poisson's law, right?

neurom/apps/cut_plane_detection.py
```diff
@@ -0,0 +1,207 @@
...
+
+def _get_probabilities(hist):
+    '''Returns -log(p) where p is the a posteriori probabilities of the observed values
+    in the bins X min, X max, Y min, Y max, Z min, Z max
+
+    Parameters:
+        hist: a dict of the X, Y and Z 1D histograms
+
+    Returns: a dict of -log(p) values'''
+    def minus_log_p(mu):
```
mgeplf: I don't think this needs to be a function, but the comments are important

neurom/apps/cut_plane_detection.py
wizmer: Yep, I think it does. Don't you think ?

mgeplf: Dunno, haven't thought about it enough

PR #650, opened 2017-09-22
graft_neuron functionality

wizmer:
Returns a neuron with a single neurite starting at the given section.
Note that this is not a deep copy: the data points in the grafted neurite
are the same as in the original neuron

This PR will allow Lida (and others!) to use already existing neuromorpho features but on specific neuron parts (such as starting from the apical point for example)

### DIFF
--- a/neurom/core/_neuron.py
+++ b/neurom/core/_neuron.py
@@ -33,6 +33,7 @@ from copy import deepcopy
 import numpy as np
 
 from neurom import morphmath
+from neurom.core._soma import Soma
 from neurom.core.dataformat import COLS
 from neurom.utils import memoize
 from neurom._compat import filter, map, zip
@@ -114,6 +115,12 @@ def iter_segments(obj, neurite_filter=None):
                                for sec in sections)
 
 
+def graft_neuron(root_section):
+    '''Returns a neuron starting at root_section'''
+    assert isinstance(root_section, Section)
+    return Neuron(soma=Soma(root_section.points[:1]), neurites=[Neurite(root_section)])
+
+
 class Section(Tree):
     '''Class representing a neurite section'''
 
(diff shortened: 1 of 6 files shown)


neurom/core/_soma.py
```diff
@@ -44,8 +44,10 @@ class Soma(object):
     Holds a list of raw data rows corresponding to soma points
     and provides iterator access to them.
     '''
+
     def __init__(self, points):
         self._points = points
+        self.radius = 0
```
wizmer: I added the radius attribute here as I discovered that otherwise you can't call the draw method on a Neuron based on a Soma instance (instead of let's say SomaSinglePoint for example).

neurom/core/_neuron.py
```diff
@@ -113,8 +114,15 @@ def iter_segments(obj, neurite_filter=None):
                                for sec in sections)
 
 
+def graft_neuron(root_section):
```
mgeplf: I wonder whether there should be a recenter option, so that the soma is located at the origin?

neurom/core/_soma.py
mgeplf: yup, makes sense

neurom/core/_neuron.py
```diff
@@ -217,6 +227,12 @@ def __deepcopy__(self, memo):
     def __nonzero__(self):
         return bool(self.root_node)
 
+    def __eq__(self, other):
```
mgeplf: do you need `__eq__` and `__hash__` in this review?

neurom/core/_neuron.py
wizmer: You mean, take the soma from the original neuron and translate it at the new root ?

neurom/core/_neuron.py
wizmer: I needed __eq__ for the assert_equal of the unit tests.
And apparently you need to implement __hash__ whenever you implement __eq__: 
https://stackoverflow.com/questions/1608842/types-that-define-eq-are-unhashable

neurom/core/_neuron.py
wizmer: We could add a optional soma argument so the user can provide his own soma if he wants to.

PR #651, opened 2017-09-25
Clean pylint stuff

wizmer:
Reordering of the import statements so that pylint stops complaining (at last!).
As I use autopep8 some blank lines and other minor changes have appeared as well.


wizmer:
@mgeplf , pylint is actually less Nazi than I thought first. He just wants the built-in imports first :)

PR #653, opened 2017-09-28
re-enable pylint output

mgeplf:


wizmer:
It was in my next PR but thx :)

PR #657, opened 2017-10-23
MUK-Annotate replacement in NeuroM

wizmer:
Principal changes:
- Add annotate.py, a tool to write annotations (dangling dendrites, narrow start and such)
  with the Neurolucida format
- Still a lot of autopep8


examples/nl_fst_compat.py
```diff
@@ -50,9 +51,9 @@
       nm.get('principal_direction_extents', nrn_asc))
 
 print('\nNumber of neurites:')
-for nt in nm.NeuriteType:
+for nt in iter(nm.NeuriteType):
```
wizmer: Otherwise, pylint complains

PR #658, opened 2017-10-26
Stop supporting python 2.6 and bump h5py version

wizmer:
The install of h5py version 2.2.1 fails so bumping to a more recent one


mgeplf:
May aswell remove python 2.6 compat at the same time...

PR #660, opened 2017-11-22
Add narrow dendrite check

wizmer:
Add a function for checking that the neuron has no narrow dendrites.
Narrow dendrites are dendrites whose at least one segment greater than a given threshold has a mean radius smaller than a threshold radius.

See https://bbpteam.epfl.ch/project/issues/browse/NSETM-275

### DIFF
--- a/neurom/check/neuron_checks.py
+++ b/neurom/check/neuron_checks.py
@@ -287,3 +287,26 @@ def has_no_dangling_branch(n):
     bad_ids = [(neurite.root_node.id, [neurite.root_node.points[1]])
                for neurite in iter_neurites(n) if is_dangling(neurite)]
     return CheckResult(len(bad_ids) == 0, bad_ids)
+
+
+def has_no_narrow_neurite_section(neuron,
+                                  neurite_filter,
+                                  radius_threshold=0.05,
+                                  considered_section_min_length=50):
+    '''Check if the neuron has dendrites with narrow sections
+        sections below 'considered_section_min_length' are not taken into account
+    Returns:
+        CheckResult with result. result.info contains the narrow section ids and their
+        first point
+    '''
+
+    considered_sections = (sec for sec in iter_sections(neuron, neurite_filter=neurite_filter)
+                           if sec.length > considered_section_min_length)
+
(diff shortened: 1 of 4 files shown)


neurom/check/neuron_checks.py
```diff
@@ -287,3 +287,26 @@ def is_dangling(neurite):
...
+        first point
+    '''
+    def dendrite_filter(n):
+        '''Select only dendrites'''
+        return n.type == NeuriteType.basal_dendrite or n.type == NeuriteType.apical_dendrite
+
+    considered_sections = filter(lambda sec: sec.length > considered_section_min_length,
+                                 iter_sections(neuron, neurite_filter=dendrite_filter))
+
+    def bad_section(section):
```
jdcourcol: I would call it narrow_section

wizmer: What about narrow_dendritic_section ?
This makes me think, should the check also be applied to axons as well ?

jdcourcol: I meant replacing "bad_section" by "narrow_section". Then adding _dendritic_ does not bring much as you are not filtering if it is dendritic in that "bad_section" function.

jdcourcol: for the axons, I would ask one of the scientists if that is useful before.

wizmer: I see. Renamed.

neurom/check/neuron_checks.py
```diff
@@ -287,3 +287,27 @@ def is_dangling(neurite):
     bad_ids = [(neurite.root_node.id, [neurite.root_node.points[1]])
                for neurite in iter_neurites(n) if is_dangling(neurite)]
     return CheckResult(len(bad_ids) == 0, bad_ids)
+
+
+def has_no_narrow_dendritic_section(neuron, radius_threshold=0.05,
```
jdcourcol: can we have instead has_no_narrow_neurite_section ?
with a parameter than can select which neurite type are taken into account ?

wizmer: Good point. Added

PR #662, opened 2017-12-05
fix spelling errors found by Alexander

mgeplf:


mgeplf:
@alex4200 thanks for the help; documentation will be updated w/ the fixes when we do a new release

PR #669, opened 2017-12-12
Fix #663: Refactor Sholl analysis code

### LINKED ISSUE #663: Sholl analysis is performing a lot of unnecessary computations
The results of Sholl analysis return of a lot of zeros. I _think_ this is because the bounding box encloses the whole neuron while it should enclose only the trees of a specific neurite_type on which the computation is actually performed. See the following line:

https://github.com/BlueBrain/NeuroM/blob/49dbaef7ed78c351d8adb3085843010716afa217/neurom/fst/_neuronfunc.py#L201

wizmer:
Neuron attributes soma center and soma radius were being computed for each neurite.
Now they are computed only once.

### DIFF
--- a/neurom/fst/_neuronfunc.py
+++ b/neurom/fst/_neuronfunc.py
@@ -29,13 +29,15 @@
 '''Morphometrics functions for neurons or neuron populations'''
 
 import math
+
 import numpy as np
-from neurom.geom import bounding_box
-from neurom.core.types import NeuriteType
-from neurom.core.types import tree_type_checker as is_type
-from neurom.core.dataformat import COLS
-from neurom.core._neuron import iter_neurites, iter_segments
+
 from neurom import morphmath
+from neurom.core._neuron import iter_neurites, iter_segments
+from neurom.core.dataformat import COLS
+from neurom.core.types import tree_type_checker as is_type
+from neurom.core.types import NeuriteType
+from neurom.geom import bounding_box
 
 
 def neuron_population(nrns):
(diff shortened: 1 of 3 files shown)


neurom/fst/_neuronfunc.py
```diff
@@ -189,16 +191,13 @@ def sholl_frequency(nrn, neurite_type=NeuriteType.all, step_size=10):
...
-    for neuron in nrns:
-        neurites_list.extend(((neurites, neuron.soma.center)
-                              for neurites in neuron.neurites
-                              if neurite_filter(neurites)))
+    neurites_list = [(neurites, neuron.soma.center)
+                     for neuron in nrns
+                     for neurites in neuron.neurites
+                     if neurite_filter(neurites)]
 
-        min_soma_edge = min(min_soma_edge, neuron.soma.radius)
```
mgeplf: this is only getting computed once per neuron; now we're traversing nrns multiple times, no?

wizmer: Wow, I must have been tired when I wrote this code. I thought it was being computed for every neurite. Then I am not sure of what needs to be fixed now...

neurom/fst/tests/test_get_features.py
```diff
@@ -805,19 +805,19 @@ def test_soma_surface_areas():
 
 def test_sholl_frequency():
     assert_allclose(fst_get('sholl_frequency', NEURON),
-                    [4, 8, 8, 14, 9, 8, 7, 7])
+                    [4,  6, 10,  8,  8, 11,  7,  9,  8,  8])
```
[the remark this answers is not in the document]
mgeplf: nm

asanin-epfl:
@arnaudon would you need such change? How correct is to use `step_size` instead of `bins`? Currently `step_size` is used.

arnaudon:
this I'll have to take a closer look at it with @mariarv , as she was using sholl analysis lately. (@lidakanari , are you also using that in some codes?)

neurom/fst/_neuronfunc.py
```diff
@@ -187,22 +190,11 @@ def sholl_frequency(nrn, neurite_type=NeuriteType.all, step_size=10):
...
-    radii = np.arange(min_soma_edge, max_radii + step_size, step_size)
-    ret = np.zeros_like(radii)
-    for neurites, center in neurites_list:
-        ret += sholl_crossings(neurites, center, radii)
-
-    return ret
+    if bins is None:
+        min_soma_edge = min(neuron.soma.radius for neuron in nrns)
+        max_radii = np.max([np.abs(bounding_box(neuron)) for neuron in nrns])
+        bins = np.linspace(min_soma_edge, max_radii, 10)
```
arnaudon, line 196: 10 is a hardcoded number of bins, maybe we could have another argument: `n_bins`, or if `bins` is an int, it is used as n_bins? Not sure what is the best option, but I would definitely allow for an int option, for which the bounds are computed automatically as done if bins is None.

arnaudon (commented):
I think this is a good PR, it simplifies a bit of code, see my comment inline. Could we rebase to see how it works?

asanin-epfl:
Closing in favor of #905. Rebasing is pain here.

PR #676, opened 2018-03-07
Morphio

wizmer:
This PR aims at delegating the IO part (parsing and object creation) of NeuroM to the C++ library MorphIO.

This commit goes with MorphIO commit: 3fdd359b9490e098b92c16bc13d9378c3d49a12b

## API breaking changes :boom: :
- Property Neuron::sections no longer returns the soma section as its first element.
  It means that neuron.sections[0] returns the first neurite's first section.

- The enum NeuriteType is now directly imported from MorphIO. The new one does not have the value: `NeuriteType.all` so all functions that have a argument to filter on NeuriteType now have the default value neurite_type=None (where None=all). 

- NeuroM neurons were previously instanciated through a FstNeuron object from the `neurom.fst` module. For some reason, all neurom `features` were stored in this module. All FstNeuron code has been deleted and the module has been renamed to `fst.features`.

- Some structural checks no longer exists as NeuroM no longer has access to the SWC data and therefore can't perform them:
  - is_single_tree
  - has_sequential_ids (MorphIO will raise RawDataError if IDs are non-sequential)
  - has_increasing_ids (no one should care as MorphIO is able to read files with non increasing IDs)
  - has_valid_neurites (MorphIO will raise RawDataError if neurites are invalid)

## API non-breaking changes :blue_heart: :
- load_neuron instanciate a morphio::Morphology object and uses it to create the NeuroM object.

- The soma type inference has been delegated to MorphIO although the same logic is used.

- Some unit tests use a 800 lines long SWC file for the test, making it very hard to debug. For some of them, a simpler file has been used.

- Test file three_pt_soma.swc has been made specification conform.

## Unrelated change :zzz: :
- `morph_stats -l` list features with a colored output and can be filtered by passing an argument after `-l` option

- (*breaking*) Changed the expected argument for the "stream" version of `load_neuron`:
    from 2 arguments: load_neuron(stream, extension)
    to 1 2-tuple argument: load_neuron((extension, stream))
    and stream can now be a `StringIO` object or a `str`


apps/raw_data_check
```diff
@@ -91,14 +91,12 @@ def run(args):
         sys.exit('ERROR: Invalid data path %s' % data_path)
 
     for f in files:
-        raw_data = load_data(f)
+        raw_data = load_neuron(f)
```
eleftherioszisis: Are ```load_data``` and ```load_neuron``` the same through the morphio api (i.e. loading of the original file with warts and all)?

lidakanari:
I think the installation of morphio C++ module has not been added in the requirements file, so the default installation of NeuroM fails to load. See error below:

```python
In [1]: import neurom as nm
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-9e9ab1a9db19> in <module>()
----> 1 import neurom as nm

/home/kanari/github/NeuroM/Benoit_version/NeuroM/neurom/__init__.py in <module>()
     62 # to the neurom._core._morphio objects and won't need the following pragma:
     63 # pylint: disable=unused-import,import-error
---> 64 from morphio import Morphology, MorphologyVersion, SomaType, SectionType as NeuriteType
     65 
     66 

ImportError: No module named morphio
```

eleftherioszisis:
With the introduction of `MorphIO`, wouldn't it be a good opportunity for `fst`and `core` to merge?

Issue #670, opened 2018-01-18
neurite_type filter not available in viewer.draw


eleftherioszisis:
In the past there was the possibility to view different parts of the neuron depending on a neurite_type.
How I can do that with the current API?

``` python
import neurom
import neurom.viewer
nrn = neurom.load_neuron('test_data/h5/v1/Neuron.h5')

neuron.viewer.draw(nrn, neurite_type=neurom.AXON)
```

``` python
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-10-f0e17a29ff38> in <module>()
----> 1 viewer.draw(nrn, neurite_type=neurom.AXON)

/home/zisis/Projects/ARCH_NGV/ngv_venv/local/lib/python2.7/site-packages/neurom/viewer.pyc in draw(obj, mode, **kwargs)
    127 
    128     output_path = kwargs.pop('output_path', None)
--> 129     plotter(ax, obj, **kwargs)
    130 
    131     if mode != 'dendrogram':

TypeError: plot_neuron() got an unexpected keyword argument 'neurite_type'

```

mgeplf:
One way is:
```
from neurom.core.types import tree_type_checker as is_type
axon = list(nm.iter_neurites(nrn, filt=is_type(neurite_type)))[0]
neuron.viewer.draw(axon)
```

eleftherioszisis:
So there is no direct way to do it..
Anyway, this will do.
Why not incorporating it to the draw function? It's a fundamental functionality for plotting the morphologies.

mgeplf:
No direct way, at the moment, but I guess it's something that could be added at some point.  Making a PR would make it happen really soon :)

orena1:
Hi, I am getting the same issue, but I actually just want to plot anything but the axons (soma and dendrites) how can I do it?

I am trying with plotly viewer

Thanks,

wizmer:
Hi Orem,

Yes, as you can see, this is a known issue. We even have the pull request for it.
In the mean time you can use what @mgeplf suggested:
```
from neurom import load_neuron, NeuriteType
from neurom.viewer import draw                                          
from neurom.core.types import tree_type_checker as is_type 
nrn = load_neuron('blabla.swc')
axon = list(nm.iter_neurites(nrn, filt=is_type(NeuriteType.axon)))[0] 
draw(axon)
```

asanin-epfl:
[...]

PR #677, opened 2018-03-09
Issue #670: NeuriteType filter not available in viewer.draw

wizmer:
This adds a filter on neurite types when calling viewer.draw with a neuron
and with modes '2d' and '3d'.

### DIFF
--- a/neurom/tests/test_viewer.py
+++ b/neurom/tests/test_viewer.py
@@ -37,8 +37,10 @@ if 'DISPLAY' not in os.environ:  # noqa
 from neurom.view import common
 from neurom import load_neuron
 from neurom import viewer
+from neurom import NeuriteType
 
 from nose import tools as nt
+from numpy.testing import assert_allclose
 
 _PWD = os.path.dirname(os.path.abspath(__file__))
 DATA_PATH = os.path.join(_PWD, '../../test_data/swc')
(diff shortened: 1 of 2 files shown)


neurom/tests/test_viewer.py
```diff
@@ -52,6 +53,12 @@ def test_draw_neuron():
     common.plt.close('all')
 
 
+def test_draw_filter_neurite():
+    for mode in ['2d', '3d']:
+        viewer.draw(nrn, mode=mode, neurite_type=NeuriteType.basal_dendrite)
```
mgeplf: check some bounding boxes or something, to make sure that you're actually plotting only the basal_dendrite

neurom/view/view.py
```diff
@@ -156,14 +157,18 @@ def plot_soma(ax, soma, plane='xy',
                                    ignore=False)
 
 
-def plot_neuron(ax, nrn, plane='xy',
+# pylint: disable=too-many-arguments
+def plot_neuron(ax, nrn,
+                neurite_type=NeuriteType.all,
```
mgeplf: this really changes the signature of the method: anyone not using kwargs is going to be hit by this.

wizmer: Actually, the plot_ methods should not be public as neurom.viewer.draw is supposed to be the interface to all of them.

wizmer: @mgeplf Given the time it will take before NeuroM reaches a new major version I think we should still merge this nonetheless.
Reasons:
1) `plot_*` methods are not public
2) Even if it breaks someone's code (which I am not even sure), it is a trivial fix.
3) The issue it solves has been mentionned once again: https://github.com/BlueBrain/NeuroM/issues/670#issuecomment-442819943

mgeplf: ok, fine with me

PR #682, opened 2018-04-16
remove python 2.6 from travis

mgeplf:


mgeplf:
I think we should discuss this.

I'm not sure how much value there is in making sure everything runs under python26 anymore, 
but it certainly speeds up the testing time.

mgeplf:
Yeah, I'm hoping to do a release today, and that should update the documentation.

PR #683, opened 2018-04-16
add strahler order

mgeplf:
* thanks to Christian O'Reilly for initial implementation:
    https://github.com/BlueBrain/NeuroM/pull/629#issuecomment-372363500

### DIFF
--- a/neurom/fst/sectionfunc.py
+++ b/neurom/fst/sectionfunc.py
@@ -91,3 +91,35 @@ def section_meander_angles(section):
     p = section.points
     return [mm.angle_3points(p[i - 1], p[i - 2], p[i])
             for i in range(2, len(p))]
+
+
+def strahler_order(section):
+    '''Branching order of a tree section
+
+    The strahler order is the inverse of the branch order,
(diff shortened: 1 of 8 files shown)


neurom/fst/sectionfunc.py
```diff
@@ -91,3 +91,35 @@ def section_meander_angles(section):
...
+       children with greater number, then the Strahler number of the node is
+       i + 1.
+
+    No efforts have been invested in making it computationnaly efficient, but
+    it computes acceptably fast on tested morphologies (i.e., no waiting time).
+    '''
+    if section.children:
+        child_orders = [strahler_order(child) for child in section.children]
+        max_so_children = max(child_orders)
+        it = iter(co == max_so_children for co in child_orders)
```
wizmer: n_max_so_children = sum(co == max_so_children for co in child_orders)
seems more pythonic to me but that's a detail

mgeplf: that requires going through the list fully, as opposed to this way which only goes as far as finding 2 which match; probably a pointless optimization

PR #684, opened 2018-04-17
remove email, update author

mgeplf:


setup.py
```diff
@@ -54,9 +54,8 @@
 
 config = {
     'description': 'NeuroM: a light-weight neuron morphology analysis package',
-    'author': 'BBP Algorithm Development Team',
+    'author': 'BBP Neuroscientific Software Engineering',
     'url': 'http://https://github.com/BlueBrain/NeuroM',
-    'author_email': 'lida.kanari@epfl.ch',
```
wizmer: The software still needs a contact mail. Do we have a generic NSE address ?

mgeplf: We don't have an NSE address, so @jdcourcol just said to go with this.  I'm not sure if pypi will reject not having an address...something we will find out when I tag &release this.

PR #688, opened 2018-04-27
Introduced `locate_segment_position()` function

arsenius7:
Given that SONATA format stores synapse positions as section ID + relative distance along the section, the task of locating the corresponding segment might be something needed repeatedly.

Not sure if NeuroM is the appropriate place for it, though.

### DIFF
--- a/neurom/fst/sectionfunc.py
+++ b/neurom/fst/sectionfunc.py
@@ -123,3 +123,10 @@ def strahler_order(section):
             return max_so_children + 1
         return max_so_children
     return 1
+
+
+def locate_segment_position(section, fraction):
+    '''
+    Segment ID / offset corresponding to a given fraction of section length.
+    '''
+    return mm.path_fraction_id_offset(section.points, fraction)
--- a/neurom/morphmath.py
+++ b/neurom/morphmath.py
@@ -45,7 +45,7 @@ def vector(p1, p2):
     Returns:
         3-vector from p1 - p2
     '''
-    return np.subtract(p1[0:3], p2[0:3])
+    return np.subtract(p1[COLS.XYZ], p2[COLS.XYZ])
 
 
 def linear_interpolate(p1, p2, fraction):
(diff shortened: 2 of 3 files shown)


neurom/morphmath.py
```diff
@@ -81,6 +81,34 @@ def f(a, b, c):
...
+        fraction: path length fraction (0.0 <= fraction <= 1.0)
+        relative_offset: return absolute or relative segment distance
+
+    Returns:
+        (segment ID, segment offset) pair.
+    '''
+    if not (0. <= fraction <= 1.0):
+        raise ValueError("Invalid fraction: %.3f" % fraction)
+    pts = np.array(points)[:, COLS.XYZ]
+    lengths = np.linalg.norm(np.diff(pts, axis=0), axis=1)
```
arsenius7: I guess I'm supposed to use `point_dist` for calculating segment lengths, but it feels rather clumsy when I need all the lengths at once; hopefully reusing COLS.XYZ range alleviates this deviation.
In fact, I would argue that `point_dist` itself should be reusing NumPy `diff + norm`, but that's not relevant to this change.

neurom/fst/sectionfunc.py
```diff
@@ -123,3 +123,10 @@ def strahler_order(section):
             return max_so_children + 1
         return max_so_children
     return 1
+
+
+def locate_segment_position(section, fraction):
```
mgeplf: section id isn't something we have tried to make constant/public, and doesn't necessarily match what other software expects the section id to be.  I'm hesitant to have this exposed, unless it's understood that the segment id is only stable during one instance of a morphology, and that guarantees aren't made between other invocations/versions/etc

does that make sense?

arsenius7: Whether we want it or not, section IDs are _already_ used for accessing geometry through NeuroM.
I understand that it is problematic, but saying "we give no guarantees" is hardly an option.

lidakanari: That is quite useful for the morphology analysis as well. Thanks for the implementation!

arsenius7: Glad it found reuse so soon :); but please note that Mike is right: section / segment ID semantics is not well defined yet.

lidakanari: Yes, I am aware of that. The need for this was primarily internal so I assume it is consistent for the same morphology internally in NeuroM which works for me at the moment :)

PR #689, opened 2018-05-11
Synthesis stats

lidakanari:
This code includes some statistics for validations:
+    'section_bif_radial_distances'
+    'section_term_radial_distances'
+    'section_end_distances'
+    'partition_pairs'

and the code that is required for the synthesis input statistics:
+    'trunk_angles': _nrn.trunk_angles,
+    'trunk_vectors': _nrn.trunk_vectors,


neurom/fst/_neuritefunc.py
```diff
@@ -96,6 +96,32 @@ def _seclen(sec, **kwargs):
     return map_sections(_seclen, neurites, neurite_type=neurite_type)
 
 
+def section_term_lengths(neurites, neurite_type=NeuriteType.all):
```
mgeplf: I think a lot of this overlaps what was added in #595; section_term_lengths, section_bif_lengths, section_bif_branch_orders, partition_asymmetries.  How do these differ?

lidakanari: Yes, you are correct. This code was implemented a while back (so these functions were not there yet). I think the important additions that are missing and @wizmer needs are the two functions about the trunk measurements that are used as synthesis input.

mgeplf: I think it would probably be best to reduce the pull request to the parts that are needed, so it's clear what is being added

neurom/fst/_neuronfunc.py
```diff
@@ -140,6 +140,76 @@ def _elevation(section, soma):
...
+        a = _angle_between(selected_vectors[i], selected_vectors[i-1])
+
+        if np.abs(2 * np.pi - np.abs(a)) < np.abs(a):
+            angles.append(np.abs(2 * np.pi - np.abs(a)) - 2 * np.pi / len(vectors))
+        else:
+            angles.append(np.abs(a) - 2 * np.pi / len(vectors))
+
+    return angles
+
+    # angles = np.array([_angle_between(i/ np.linalg.norm(i), [0,1]) for i in vectors[:, 0:2]])
```
wizmer: Remove comments

lidakanari: Am I correct to assume that tests are required for the following functions?
```python
+    'section_bif_radial_distances': _nrt.section_bif_radial_distances,
+    'section_term_radial_distances': _nrt.section_term_radial_distances,
+    'section_end_distances': _nrt.section_end_distances,
+    'partition_pairs': _nrt.partition_pairs,
+    'trunk_angles': _nrn.trunk_angles,
+    'trunk_vectors': _nrn.trunk_vectors,
```

lidakanari:
Trying to run pylint tests I get errors outside this code, namely: 
```
************* Module neurom.fst._neuritefunc
E:386,32: Instance of 'ConvexHull' has no 'volume' member (no-member)
```
How can this be fixed or should I push it as it is?

mgeplf:
which version of scipy are you using?

lidakanari:
This should be controlled by the test environment, right? 

It looks like it requires the following:
```
scipy>=0.13.3 in /usr/lib/python2.7/dist-packages (from neurom==1.4.6) (0.13.3)
```

mgeplf:
Weird that it installs the minimal required, I should probably bump that, b/c volume doesn't exist before 0.16.

Anyways, upgrading it should fix the problem, but also you can push your commit, and the travis build will install a newer version, so you'll get past that sanity check.

lidakanari:
Yes, I was suspecting it has to do with local -vs- server versions. Thanks, I will try this one.

neurom/fst/_neuronfunc.py
wizmer: Indeed

neurom/fst/_neuritefunc.py
```diff
@@ -318,6 +326,30 @@ def section_radial_distances(neurites, neurite_type=NeuriteType.all, origin=None
...
+    for n in iter_neurites(neurites, filt=is_type(neurite_type)):
+        pos = n.root_node.points[0] if origin is None else origin
+        dist.extend([sectionfunc.section_radial_distance(s, pos)
+                     for s in iter_sections(n,
+                                            iterator_type=Tree.ileaf,
+                                            neurite_filter=is_type(neurite_type))])
+    return dist
+
+
+def section_bif_radial_distances(neurites, neurite_type=NeuriteType.all, origin=None):
```
mgeplf: is this the same as `section_term_radial_distances` except a different iterator? prob. easiest just to have a generic function, and pass in the different iterator types

neurom/fst/_neuritefunc.py
```diff
@@ -318,6 +326,30 @@ def section_radial_distances(neurites, neurite_type=NeuriteType.all, origin=None
     return dist
 
 
+def section_term_radial_distances(neurites, neurite_type=NeuriteType.all, origin=None):
+    '''Get the radial distances of the termination sections for a collection of neurites'''
+    dist = []
+    for n in iter_neurites(neurites, filt=is_type(neurite_type)):
+        pos = n.root_node.points[0] if origin is None else origin
+        dist.extend([sectionfunc.section_radial_distance(s, pos)
```
mgeplf: extend can take an iterator, so don't need the [ ]

neurom/fst/_neuronfunc.py
```diff
@@ -140,6 +140,61 @@ def _elevation(section, soma):
...
 
 
+def trunk_vectors(nrn, neurite_type=NeuriteType.all):
+    '''Calculates the vectors between all the trunks of the neuron
+    and the soma center.
+    '''
+    neurite_filter = is_type(neurite_type)
+    nrns = neuron_population(nrn)
+
+    def _vectors(section, soma):
```
mgeplf: don't think you need this helper, just put the morphmath.vector in the list comprehension

neurom/fst/_neuritefunc.py
lidakanari: Yes, but I don't want to improve the code with the current PR just extend the functionality that I need, so style improvements can be implemented later?

neurom/fst/_neuritefunc.py
lidakanari: OK, will change this.

neurom/fst/_neuronfunc.py
lidakanari: Ah, yes. Much simpler!

neurom/fst/_neuritefunc.py
lidakanari: Btw I noticed that all the other functions (such as section_radial_distances) also have the [ ] structure.. Should I keep it for consistency or modify it anyways? 
```
 dist.extend([sectionfunc.section_radial_distance(s, pos)
```

mgeplf: they should be fixed, too

neurom/fst/_neuritefunc.py
mgeplf: since both pieces of code are being added at the same time, I think it's best to do it now - it will be hard in the future to notice that they are doing almost the same thing

neurom/fst/_neuritefunc.py
mgeplf: yup

neurom/fst/_neuritefunc.py
lidakanari: The problem with merging those two functions is that then we will need to pass an argument to the "get" instead of having available the 'section_bif_radial_distances' (and respectively the 'term') as a feature. I have no idea how to do this since all the other features are done this way (one function per feature without specified arguments) so if you can propose a fix in order to pass a specific argument in the "get" to create the two required features I can update this.

mgeplf: I'd probably do something like:
```
def _section_radial_distances(..., iterator_type):
    ...

def section_term_radial_distances(...):
    return _section_radial_distances(..., iterator_type=Tree.ileaf)

def section_bif_radial_distances(...):
    return _section_radial_distances(..., iterator_type=Tree.ibifurcation_point)
```

This way it's apparent where the bulk of the algorithm is, and how the two functions differ

lidakanari: Sure, that certainly works :) Thanks!

neurom/fst/_neuritefunc.py
```diff
@@ -315,39 +315,31 @@ def partition_pairs(neurites, neurite_type=NeuriteType.all):
...
+    or only bifurcations (ibifurcation_point).'''
     dist = []
     for n in iter_neurites(neurites, filt=is_type(neurite_type)):
         pos = n.root_node.points[0] if origin is None else origin
         dist.extend(sectionfunc.section_radial_distance(s, pos)
-                    for s in n.iter_sections())
-
+                    for s in iter_sections(n,
+                                           iterator_type=iterator_type,
+                                           neurite_filter=is_type(neurite_type)))
```
lidakanari: I also just noticed that I use the neurite_type filter twice. I think the first one should be sufficient, right?

neurom/fst/_neuritefunc.py
lidakanari: Sure, if it is something simple there is no problem. Thanks for the suggestion.

neurom/fst/_neuritefunc.py
```diff
@@ -307,17 +307,40 @@ def partition_asymmetries(neurites, neurite_type=NeuriteType.all):
                              neurite_filter=is_type(neurite_type)))
 
 
-def section_radial_distances(neurites, neurite_type=NeuriteType.all, origin=None):
-    '''Remote bifurcation angles in a collection of neurites'''
+def partition_pairs(neurites, neurite_type=NeuriteType.all):
+    '''Partition pairs at bifurcation points of a collection of neurites'''
```
wizmer: Can you had the definition of a partition pair in the docstring :

The partition pair is the number of bifurcations in the two daughter subtrees at each branch point.

neurom/fst/_neuritefunc.py
```diff
@@ -307,17 +307,40 @@ def partition_asymmetries(neurites, neurite_type=NeuriteType.all):
...
+    return map(_bifurcationfunc.partition_pair,
+               iter_sections(neurites,
+                             iterator_type=Tree.ibifurcation_point,
+                             neurite_filter=is_type(neurite_type)))
+
+
+def section_radial_distances(neurites, neurite_type=NeuriteType.all, origin=None,
+                             iterator_type=Tree.ipreorder):
+    '''Remote bifurcation angles in a collection of neurites.
```
wizmer: I think this docstring is wrong

neurom/fst/_neuronfunc.py
```diff
@@ -140,6 +140,57 @@ def _elevation(section, soma):
...
+
+
+def trunk_angles(nrn, neurite_type=NeuriteType.all):
+    '''Calculates the angles between all the trunks of the neuron.
+    The angles are defined on the x-y plane and the trees
+    are sorted from the y axis and anticlock-wise.
+    '''
+    vectors = trunk_vectors(nrn, neurite_type=neurite_type)
+
+    def _angle_between(p1, p2):
```
wizmer: _angle_between should be defined in morphmath so it can be reused

neurom/fst/_neuronfunc.py
```diff
@@ -140,6 +140,57 @@ def _elevation(section, soma):
...
+        return np.arccos(np.clip(np.dot(v1, v2), -1.0, 1.0))
+
+    def _sort_angle(p1, p2):
+        """Angle between p1-p2 to sort vectors"""
+        ang1 = np.arctan2(*p1[::-1])
+        ang2 = np.arctan2(*p2[::-1])
+        return (ang1 - ang2)
+
+    # In order to avoid the failure of the process in case the neurite_type does not exist
+    if not vectors.size:
```
wizmer: put this right after the assignment of `vectors`

neurom/fst/_neuritefunc.py
lidakanari: True! Will correct this documentation.

neurom/fst/_neuritefunc.py
lidakanari: I did not write this function :) Should I correct it anyways?

neurom/fst/_neuronfunc.py
lidakanari: OK will move this

PR #691, opened 2018-05-25
update scipy version

mgeplf:
* no longer support py2.6, so we can use the scipy that has
    ConvexHull.volume


wizmer:
How is this related to py2.6 ?

PR #693, opened 2018-06-01
Only perform the deploy stage with python 2.7

wizmer:


mgeplf:
just have to remember this happened when we release again

PR #698, opened 2018-06-27
3-points contours are no longer interpreted as SomaThreePoint

wizmer:
But as SomaSimpleContour as it should be


mgeplf:
I'll review after it's passing travis

wizmer:
Cleant-up and re-opened at https://github.com/BlueBrain/NeuroM/pull/717

PR #699, opened 2018-07-10
Create plotly module

wizmer:
A new backend to create plots.

It can be used as:
```
from neurom.viewer.plotly import draw
draw(neuron, plane='xy')
draw(neuron, plane='3d')
```


mgeplf:
1) If we're doing a pylint bump, can you do that in a separate commit: too much is going on here
2) Plotly should be an optional dependency

jdcourcol:
> If we're doing a pylint bump, can you do that in a separate commit: too much is going on here
+1

jdcourcol:
Can you show what plotly brings ?

mgeplf:
@jdcourcol: Benoit showed me: the 3d plotting looks a lot better than the matplotlib version - makes debugging/viewing much more pleasant

wizmer:
Plotly is more interactive and more user friendly I feel. Especially the 3D plotter used to be ununsable as it was slow and stuttering while rotating the camera. Plotly is much more fluid.

Also, you can embed an interactive plot into a ipython notebook (I have added a notebook that shows how to do it).

wizmer:
You can have a look at the interactive plots here:
http://nbviewer.jupyter.org/github/wizmer/NeuroM/blob/2eedcc8f107b579eaeaa7c82f6ceec3a23c2a90c/tutorial/plotly.ipynb

wizmer:
@mgeplf Noob question. What is the benefit to put it as an external dependency ? Are you afraid that plotly might not be available on all platforms ?
Because I'd naturally go the other way around (aka. the rude way), putting it as a normal dependency and downgrading as an optional one if someone encounters an issue.

neurom/fst/tests/test_get_features.py
```diff
@@ -66,7 +66,7 @@ def assert_items_equal(a, b):
 
 def assert_features_for_neurite(feat, neurons, expected, exact=True):
     for neurite_type, expected_values in expected.items():
-        print('neurite_type: %s' % neurite_type)
+        # print('neurite_type: %s' % neurite_type)
```
wizmer: I kept it because I think it is used for debugging the tests

.travis.yml
```diff
@@ -23,6 +23,7 @@ script:
   - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then make -f Makefile run_tests; fi
 after_success:
   - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then pip install codecov; codecov; fi
+after_failure: "cat /home/travis/build/BlueBrain/NeuroM/pylint.txt; cat /home/travis/build/BlueBrain/NeuroM/pycodestyle.txt"
```
mgeplf: nice

neurom/fst/tests/test_get_features.py
mgeplf: k, fair enough

Issue #697, opened 2018-06-19
Section type constants are shown as integers


arsenius7:
Since #631, `NeuriteType` enum values are displayed as integers in pandas DataFrame unless explicitly casted to `object` dtype.
Perhaps we should achieve enum ordering by some other means rather than making it `IntEnum` (for instance, backport `OrderedEnum` implementation).

PR #700, opened 2018-07-11
Fix: Section type constants are shown as integers #697

wizmer:


arsenius7:
LGTM

requirements_dev.txt
```diff
@@ -29,6 +29,7 @@
 mock>=1.3.0
 pylint==1.7.4
 pycodestyle==2.3.1
+pandas==0.22.0
```
mgeplf: adding pandas as a dependency just to run a test seems like overkill

wizmer: I know but it's a only dev dependency so I don't think it's a big deal.
And that's the only way I can think of to test the expected behaviour

arsenius7: If it is such a big deal, checking the type of the value might suffice.

wizmer: Ah wait, maybe I can check on a np.array instead. Let me try

PR #705, opened 2018-07-20
Rename view module to viewer

wizmer:
and move neurom/viewer.py to neurom/viewer/\_\_init\_\_.py
so that public function are accessed the same way:
```python
from neurom.viewer import draw
```

Rationale:
As of today the structure is confusing since public functions can potentially
in one of the 3 following files.
- viewer.py
- view/\_\_init\_\_.py
- view/view.py

Now at least viewer.py is gone and public functions are exposed in the
viewer module though the \_\_init\_\_.py


eleftherioszisis:
Looks good to me!

neurom/view/__init__.py
```diff
@@ -1,33 +0,0 @@
-# Copyright (c) 2015, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
```
mgeplf: I think we should add a deprecation warning here, pointing people to where they should look for the functions now

neurom/viewer/view.py
```diff
@@ -37,7 +37,7 @@
 from neurom.core._soma import SomaCylinders
```
mgeplf: my only nit is whether this should be called view anymore; part of the reason for this refactor is because .viewer and .vew are a strange naming, but having viewer.view also seems weird to me

I wonder if it should be 'plot.py', and then the functions can be renamed from plot_tree -> tree.  I'm not convinced of this, though, so it's more of a request for discussion

lidakanari: I agree that viewer.view is weird and confusing. However, I would avoid to use plot because usually the convention is to use plot for functions, histograms, etc and visualize (or view, draw) for an object such as a neuron, tree, etc.

wizmer: Indeed view.py is not explicit. First thing, if we keep this file, regardless of the name, I think it should be preceeded by a leading underscore to make it clear that there are no public functions here.

Maybe we could restructure the following way: plot_dendrogram function goes into _dendrogram, 2D stuff goes to _view_2d.py and 3d stuff goes to _view_3d.py. The content of the folder would then be:

- __init__.py
- common.py (eventually to be renamed into _common.py)
- _view_2d.py
- _view_3d.py
- _dendrogram.py
- plotly.py (no underscore because it contains the public function `draw`)

eleftherioszisis:
Something that should be taken into account concerning the name `draw`. I have said this again in the past when `draw` was used as the neurom plotting function, but I would like to remind it.

If one launches ipython the following way (which is quite common).
```
ipython --pylab
```
Then the function `draw` is automatically imported from `matplotlib.pyplot`. Thus, if one imports the neurom module: `from neurom.viewer import draw`, it replaces the `matplotlib` function.

asanin-epfl:
Closing as `view/__init__.py` have been cleaned of imports. Now there are `viewer` and `view/view`. It's a bit misleading but for now it is fine. There are more urgent things to do.

PR #710, opened 2018-07-31
Fig dangling check crash when no soma

wizmer:
Instead, set all neurites as dangling

Note: I know right now NeuroM does not support empty soma but it will in the future and already can no the morphio branch. So I prefer to commit here and then to cherry pick on morphio.


wizmer:
I need this to fix a bug on MorphService which uses morphio branch.

asanin-epfl:
closing in favor of #900

PR #711, opened 2018-08-30
Require 'enum34' only for Python<3.4

arsenius7:


setup.py
```diff
@@ -36,7 +36,7 @@
 
 VERSION = "1.4.8"
 
-REQS = ['enum34>=1.0.4',
+REQS = ['enum34>=1.0.4;python_version<"3.4"',
```
mgeplf: fancy: python_version<"3.4"
that must be new

PR #713, opened 2018-08-31
Revert marker on enum34 for python >= 3.4

wizmer:
As it is not supported by old distutils


wizmer:
If you don't agree, shout now.

PR #715, opened 2018-09-24
Cut plane moved to its own python package

wizmer:


wizmer:
The repo I linked does not exist yet but I have asked for its creation

PR #717, opened 2018-11-29
3-points contours are no longer interpreted as SomaThreePoint

wizmer:
In NeuroM a soma made of 3 points could be of type:
- SomaThreePoint if the soma_class was SOMA_CONTOUR

And if soma_class is SOMA_CYLINDER, there was 2 options:
- SomaNeuromorphoThreePointCylinders if it follows the spec of the soma 3pt
cylinder (http://neuromorpho.org/SomaFormat.html)
- SomaCylinders else

The change is that now, in the case of a soma_class == SOMA_CONTOUR:
It is of type SomaSimpleContour instead of SomaThreePoint


neurom/fst/tests/test_neuronfunc.py
```diff
@@ -134,7 +135,7 @@ class Mock(object):
 
 @nt.raises(Exception)
 def test_trunk_elevation_zero_norm_vector_raises():
-    _nf.trunk_origin_elevations(NRN)
+    _nf.trunk_origin_elevations(SWC_NRN)
```
wizmer: With the change of soma type and interpretation, the soma center has moved from [0,0,0] to [0.33, 0.1, 0].
Which causes, this test not to pass with the old neuron.

PR #720, opened 2019-01-08
Switch to Tox

wizmer:
- Remove old Makefile
- Update pylint to 1.9.4 and make necessary changes in the code
- Add .coveragerc to enforce show missing lines even on newer nose versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,15 +15,14 @@ addons:
     - libhdf5-dev
 cache: pip
 install:
-  - pip install 'matplotlib==1.5.2'
+  - pip install 'matplotlib==1.5.2' tox
   - pip install .
 script:
-  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then make -f Makefile test; fi
-  - if [[ $TRAVIS_PYTHON_VERSION == 3.5 ]]; then make -f Makefile run_tests; fi
-  - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then make -f Makefile run_tests; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then tox -e py27 -e lint; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.5 ]]; then tox -e py35; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then tox -e py36; fi
 after_success:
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -15,15 +15,14 @@ addons:
...
 script:
-  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then make -f Makefile test; fi
-  - if [[ $TRAVIS_PYTHON_VERSION == 3.5 ]]; then make -f Makefile run_tests; fi
-  - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then make -f Makefile run_tests; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then tox -e py27 -e lint; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.5 ]]; then tox -e py35; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then tox -e py36; fi
 after_success:
-  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then pip install codecov; codecov; fi
-after_failure: "cat /home/travis/build/BlueBrain/NeuroM/pylint.txt; cat /home/travis/build/BlueBrain/NeuroM/pycodestyle.txt"
```
mgeplf: this is really useful, btw

neurom/apps/tests/test_morph_stats.py
```diff
@@ -125,13 +125,13 @@ def test_eval_stats_applies_numpy_function():
 def test_extract_stats_single_neuron():
     nrn = nm.load_neuron(os.path.join(DATA_PATH, 'Neuron.swc'))
     res = ms.extract_stats(nrn, REF_CONFIG)
-    nt.eq_(res.keys(), REF_OUT.keys())
+    nt.eq_(set(res.keys()), set(REF_OUT.keys()))
```
mgeplf: was doing pylint stuff wasn't necessary for tox? it makes the diff hard to read, so I prefer if only one thing is changed at a time

.travis.yml
wizmer: Tox does not support redirecting stdout from what I could see so I was not able to generate pylint.txt and pycodestyle.txt
But on the other hand, the output will be printed so it has the same effect as the `cat`.

mgeplf: cool

neurom/apps/tests/test_morph_stats.py
mgeplf: ic, good to add that to the commit message so it's known

PR #721, opened 2019-01-09
Remove Makefile

wizmer:
As Tox makes it obselete.

@mgeplf The Makefile seems to provide many function related to the documentation. In Tox, it only becomes `tox -e docs`. Do we need to support other make commands ?


Makefile
```diff
@@ -1,100 +0,0 @@
...
-else
-   VENV:=venv
-endif
-VENV_BIN:=$(VENV)/bin
-
-# simulate running in headless mode
-unexport DISPLAY
-
-# Test coverage pass threshold (percent)
-MIN_COV?=100
```
mgeplf: as long as this still tested somewhere

Makefile
```diff
@@ -1,100 +0,0 @@
...
-#see if we're in a virtualenv, and use that, otherwise use the default
-ifdef VIRTUAL_ENV
-   VENV=$(VIRTUAL_ENV)
-else
-   VENV:=venv
-endif
-VENV_BIN:=$(VENV)/bin
-
-# simulate running in headless mode
-unexport DISPLAY
```
mgeplf: dunno if we need this

Issue #678, opened 2018-04-06
'morph_stats' json outputs: {soma feature's name: value} pairs not collected into a unique key


pedroernesto:
Currently, NeuroM 'morph_stats' module provides a 'json' output file, containing a dictionary where each 'soma feature's name' key appears  at the same level as other cell parts (e.g. apical dendrite or axon), instead at the level for feature names and their values. Can the user customize 'morph_stats' output to correct that problem or such functionality is not implemented yet?

Reading the content of such dictionary might be simpler if all {soma feature's name: value} pairs were collected into a unique 'soma' key as well. For now, I am needing to re-format 'morph-stats' output with something like this to re-group those soma-features values into a unique key:
```
for dict1 in output_dict.values():
            soma_features = dict()
            for key, val in dict1.items():
                if key.find('soma') == -1:
                    continue
                soma_features.update({key: val})
                del dict1[key]
                dict1.update({"soma": soma_features})
```

asanin-epfl:
Closing by the same reason as #722 is closed. Outdated.

PR #722, opened 2019-01-09
morph_stats config can now handle and output a 'soma' block

wizmer:
Issue https://github.com/BlueBrain/NeuroM/issues/678

Soma properties can be specified in the 'neuron' block (legacy)

```
{
    'neuron': {
        'soma_radii': ['mean'],
    }
}
```

or in the 'soma' block:
```
{
    'soma': {
        'soma_radii': ['mean'],
    }
}
```

Which block was used (whether neuron or soma) will be reflected in the
output json.


arnaudon:
What about adding soma as another neurite_type?

asanin-epfl:
@arnaudon would you need this one? Otherwise I will close it.

arnaudon:
pfiuuuu, I completely forgot about that one, I think the soma things are now in neuron entry or something. Let's drop that, and we'll make another one more up to date if we need to.

asanin-epfl:
thanks for clarifying

PR #724, opened 2019-01-31
Add _get_figure method in plotly module

wizmer:
to be able to reuse the figure outside of neurom

Numpy version is constrained until the bug in pylint in conjunction
with numpy 1.16 is fixed.
PyCQA/pylint#2414


setup.py
```diff
@@ -41,7 +41,7 @@
         'future>=0.16.0',
         'h5py>=2.7.1',
         'matplotlib>=1.3.1',
-        'numpy>=1.8.0',
+        'numpy>=1.8.0,<1.16',
```
mgeplf: why?

wizmer: Cf updated summary.

mgeplf: wish it didn't have to be that way, but, lgtm

wizmer: Hopefully it should be for a short amount of time...

PR #727, opened 2019-01-31
Bump to next dev version 1.4.12.dev0

wizmer:


wizmer:
To check my assumption about readthedoc

PR #728, opened 2019-02-01
`tox -e lint` now defaults to Python 3

wizmer:
and pylint is 2.2.2


pylintrc
```diff
@@ -16,7 +16,7 @@
 #W0142 - Used * or ** magic
 #W0622 - redefined-builti
 #C0325 - superfluous-parens
-disable=C0103,R0904,R0903,W0511,W0232,R0922,R0801,R0921,W0141,R0401,I0013,W0142,W0622,C0325
+disable=C0103,R0904,R0903,W0511,W0232,R0922,R0801,R0921,W0141,R0401,I0013,W0142,W0622,C0325,R0205
```
mgeplf: write down what R0205 means, above

neurom/io/utils.py
```diff
@@ -106,7 +106,7 @@ def get_files_by_path(path):
     '''
     if os.path.isfile(path):
         return [path]
-    elif os.path.isdir(path):
+    if os.path.isdir(path):
```
mgeplf: what does pylint not like about this?

wizmer: it does not like an elif after a return

pylintrc
wizmer: done

PR #734, opened 2019-02-20
reduce error messages like: 'No handlers could be found for logger'

### LINKED ISSUE #362: neurolucida: No handlers could be found for logger

mgeplf:
* add NullLogger as per:
    https://docs.python-guide.org/writing/logging/#logging-in-a-library
* checked that if logging is setup, the logs are still propagated
* hat time @genric
* fixes #362

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -19,10 +19,13 @@ commands = nosetests
 [testenv:lint]
 basepython = python3
 deps =
-    pycodestyle==2.3.1
+    pycodestyle
+    # need to pin b/c https://github.com/PyCQA/astroid/issues/650
+    #                 https://github.com/PyCQA/astroid/issues/651
+    astroid==2.1.0
     pylint==2.2.2
 commands =
-    pycodestyle --config=pycodestylerc --exclude=tests neurom
+    pycodestyle --exclude=tests neurom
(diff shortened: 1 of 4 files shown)


mgeplf:
ping

wizmer:
From: https://pythonhosted.org/logutils/libraries.html
> to avoid this, the library developer simply needs to instantiate a NullHandler and add it to the top-level logger of the library module or package.
I haven't tried but this suggests that we should be able to put the NullHandler in the top level __init__.py and be fine with it. Have you tried it ?

mgeplf:
Didn't try that, does seem like a more elegant solution

tox.ini
```diff
@@ -19,10 +19,13 @@ commands = nosetests
 [testenv:lint]
 basepython = python3
 deps =
-    pycodestyle==2.3.1
+    pycodestyle
+    # need to pin b/c https://github.com/PyCQA/astroid/issues/650
```
wizmer: Ah!

mgeplf: yeah...we'll have to delete this once it's fixed upstream :(

Issue #735, opened 2019-03-08
swc custom type fails to load


eleftherioszisis:
Hi,
I tried to load hundreds of standardized glial cells in `neurom` from `neuromorpho` but the loader fails. Example file: [A1-CA1-L-A63x1zACR3.CNG.swc.zip](https://github.com/BlueBrain/NeuroM/files/2945440/A1-CA1-L-A63x1zACR3.CNG.swc.zip)

I think this is because of the partial support of `neurom` for the swc types (http://www.neuronland.org/NLMorphologyConverter/MorphologyFormats/SWC/Spec.html). Would it be possible to "relax" the `neurom` types, by allowing the user to define a custom type?

``` python
import neurom
nrn = neurom.load_neuron('A1-CA1-L-A63x1zACR3.CNG.swc')
```
``` python
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-2-d35986bfd206> in <module>
----> 1 nrn = neurom.load_neuron('A1-CA1-L-A63x1zACR3.CNG.swc')

~/Projects/ASTROCYTE_SYTHESIS/venv/lib/python3.7/site-packages/neurom/io/utils.py in load_neuron(handle, reader)
    120     else:
    121         name = None
--> 122     return FstNeuron(rdw, name)
    123 
    124 

~/Projects/ASTROCYTE_SYTHESIS/venv/lib/python3.7/site-packages/neurom/fst/_core.py in __init__(self, data_wrapper, name)
     43     def __init__(self, data_wrapper, name='Neuron'):
     44         self._data = data_wrapper
---> 45         neurites, sections = make_neurites(self._data)
     46         soma_check, soma_class = _SOMA_CONFIG[self._data.fmt]
     47         soma = make_soma(self._data.soma_points(), soma_check, soma_class)

~/Projects/ASTROCYTE_SYTHESIS/venv/lib/python3.7/site-packages/neurom/fst/_core.py in make_neurites(rdw)
     87                           points=rdw.data_block[sec.ids],
     88                           section_type=_TREE_TYPES[sec.ntype])
---> 89                   for i, sec in enumerate(rdw.sections))
     90 
     91     # One pass over nodes to connect children to parents

~/Projects/ASTROCYTE_SYTHESIS/venv/lib/python3.7/site-packages/neurom/fst/_core.py in <genexpr>(.0)
     87                           points=rdw.data_block[sec.ids],
[...]

PR #736, opened 2019-03-11
New NeuriteType: custom = 7

wizmer:
Change internal numbering of the enum
Fallback to NeuriteType.undefined when undefined type

Fix bug: https://github.com/BlueBrain/NeuroM/issues/735


neurom/fst/tests/test_feature_compat.py
```diff
@@ -99,33 +99,33 @@ def test_get_n_sections(self):
         nt.assert_equal(_nrt.n_sections(self.sec_nrn),
                         get('number_of_sections', self.ref_nrn)[0])
 
-        for t in NeuriteType:
+        for t in self.ref_types:
```
wizmer: I had to change this because we don't want to iterate on custom type here

neurom/core/types.py
```diff
@@ -47,11 +47,12 @@ class NeuriteIter(OrderedEnum):
 @unique
 class NeuriteType(OrderedEnum):
     '''Enum representing valid tree types'''
-    undefined = 1
-    soma = 2
-    axon = 3
-    basal_dendrite = 4
-    apical_dendrite = 5
+    undefined = 0
```
wizmer: @mgeplf I know we just discussed keeping the value different but I just did it for practicality reason as section_type is now accessed by a dict

eleftherioszisis:
Looks good to me!

neurom/core/types.py
mgeplf: I don't follow why it being access by a dict makes a difference?

test_data/swc/custom_type.swc
```diff
@@ -0,0 +1,11 @@
+# Use custom type: 7
```
mgeplf: what does this do to iteration when custom types exist? what about when undefined?

neurom/fst/_core.py
```diff
@@ -85,7 +85,7 @@ def make_neurites(rdw):
     # One pass over sections to build nodes
     nodes = tuple(Section(section_id=i,
                           points=rdw.data_block[sec.ids],
-                          section_type=_TREE_TYPES[sec.ntype])
```
wizmer: The reason dict is making a difference is here: _TREE_TYPES used to be accessed with sec.ntype used as the index in the _TREE_TYPES list. It was almost working by luck and was assuming consecutivity of the defined numbers. Doing _TREE_TYPES[custom = 7] should have meant that custom should be at the 8th position of the _TREE_TYPES list.

test_data/swc/custom_type.swc
wizmer: What do you mean ?

neurom/fst/_core.py
mgeplf: Ok, but I don't see why that requires renumber the constants?

wizmer: Because now sec.ntype is now used as the key of the dict. And sec.ntype is the value from the file you're reading so it has to be in sync.

test_data/swc/custom_type.swc
mgeplf: What would the user expect when iterating over NeuriteType.all?  Same with the other helpers, ie: is 'custom' a dendrite?

wizmer: That's a tough question. Two answers:
_ Yes, custom is a dendrite. There are little chances people are gonna define custom axon types. And if we don't do this all morphometrics for morphologies with custom types are gonna be broken.
_ No, custom is not a dendrite. all should only refer to properly defined types apical_dendrite and basal_dendrite

or should we rename custom to dendrite_custom to make it more obvious ?

neurom/core/types.py
```diff
@@ -44,15 +44,22 @@ class NeuriteIter(OrderedEnum):
     NRN = 2
 
 
+# Adding 100 to SWC actual values
+# to force people to use the Enum
+# and not rely on the fact that
+# the Enum value matches the SWC one
+_INTERFACE_DECOUPLING = 100
```
eleftherioszisis: Seriously? What about cells with types that do not correspond to the swc ones because the cells are not neurons? I don't understand why everybody has to use the Neuron types and create maps. What is wrong with things being intuitive in `NeuroM`...

lidakanari: I don't get what is this offset? Also if this comment is correct " # the Enum value matches the SWC one " it doesn't make much sense.

mgeplf: >Seriously? What about cells with types that do not correspond to the neurom ones because the cells are not neurons?

The point of this is to encourage the use of the named types, as opposed to just using integers to identify different neurite types.  By using the names values, it makes the code more readable, searchable and thus maintainable.

> I don't understand why everybody has to use the Neuron types and create maps.

I'm not sure I follow.  What maps are people creating?

wizmer: 2 things:
- Enum values are not exposed to the user and should NOT be used. Never. So, you should not care about them. It could be 47, 52 and -27, NeuroM would work the same. NeuroM is a neuron loader, it should not be attached to a given file format. What if we start supporting other file formats that have different numbering conventions ?
- The problem @eleftherioszisis mentionned does not stem in the NeuriteType enum values but in the fact that non neuron cells have section type of (python) type NeuriteType. This is a more valid debate to me.

wizmer: Yeah, to second @mgeplf, if you have use cases were you have to use integer value because Enums don't work out, brings them here so we can discuss them.

test_data/swc/custom_type.swc
eleftherioszisis: Is it possible to allow the user define a type mapping?

For example start with a default one, where anything else apart from the defined types is undefined unless the user passes their own types. BUT these types correspond to the types found in the file not an arbitrary set of integers.
```
class DefaultTypes(object):
    undefined = 0
    soma = 1
    axon = 2
    basal_dendrite = 3
    apical_denrite = 4
```
or the user could provide:
```
class UserTypes(object):
    undefined = 0
    soma = 1
    endfoot = 2
    glia_process = 3
    exotic_process = 7
```

neurom/core/types.py
wizmer: > I don't get what is this offset? Why should we not conform to the swc types?

We do conform to SWC types. When you load a section = 2 in NeuroM you end up with NeuriteType.axon.
It is just that the internal values (which you should never use) do not map to SWC values. It's done on purpose to prevent people to use this internal values.

test_data/swc/custom_type.swc
wizmer: The more we discuss about this the more I think we need different loader function for each type of cell.
Instead of allowing user define type mapping having AstrocyteSectionType that would be set in stones. And a function load_astrocyte that would be the equivalent of load_neuron but that would return sections of type AstrocyteSectionType ?

neurom/core/types.py
eleftherioszisis: When I use the `neurom`, `tmd`, `tns`, `morphio` tools I have to create a mapping from an integer file section type to all of these tools. So far, the last two agree, but `neurom`, which doesn't conform to the swc numbering, has to have a separate treatment. So instead of having a unified logic, I have to take cases for the arbitrary types. I don't understand why do we have to punish the users that for any reason would like to use the underlying types.

Furthermore, when i process cells, like astrocytes, I usually need to create intermediate types, from which I need to extract stats with `NeuroM`. This is why for example undefined or custom neurite types are useful. It is cleaner in my opinion to be able to say `endfoot -> 2` and expect `NeuroM` to map a neurite with type 2, instead of `endfoot -> 2 -> neuron axon -> neurom interface -> 102`

lidakanari: In my opinion, having different loaders might not generalize optimally. I have cells for various different types (microglia, oligodendrocytes etc) that obey the swc standards. On the other hand, I also have neurons from different labs with all sorts of "weird" neurite types. I'm not sure what's the best way to deal with all these inconsistencies, but separating loaders might not be the optimal strategy.

eleftherioszisis: It has to be said however that different loaders make sense for cells that carry additional datasets that are unique to their type. For example, astrocytes have perimeters that do not exist in neuron moprhs.

test_data/swc/custom_type.swc
mgeplf: I just talked to @eleftherioszisis and the use case he's running into is that some of the reconstructions are mis-numbered; ie: 7 is used for `endfoot` instead of 2 (?).  Instead of having a unique loader, one option, I think, would be to pass a mapping into a loader that says what types to map to the internal types:

```
import neurom as nm
nrn = nm.load_neuron('some/data/path/morph_file0.swc', mapping={7: NeuriteType.axon, ...})
```

This would partly solve his problem, however, it doesn't solve the problem of having to use the NeuroM typenames (axon/dendrite/etc) for Glia concepts (endfoot/process/etc), but it would allow for more flexibility when loading files.

lidakanari: That proposal is consistent with the tree_type definition that happens during loading in TMD morphologies. It would be useful if this is possible to extend to MorphIO so all approaches are synchronized.

eleftherioszisis: Yeah, that will provide us with much more flexibility in cases of custom typing. I am in favor also in favor of @wizmer suggestion for a different loader for astrocytes. However, this will imply that features will have to be further classified by `cell_type`.

wizmer: > I just talked to @eleftherioszisis and the use case he's running into is that some of the reconstructions are mis-numbered; ie: 7 is used for `endfoot` instead of 2 (?). Instead of having a unique loader, one option, I think, would be to pass a mapping into a loader that says what types to map to the internal types:
> 
> ```
> import neurom as nm
> nrn = nm.load_neuron('some/data/path/morph_file0.swc', mapping={7: NeuriteType.axon, ...})
> ```
> 
> This would partly solve his problem, however, it doesn't solve the problem of having to use the NeuroM typenames (axon/dendrite/etc) for Glia concepts (endfoot/process/etc), but it would allow for more flexibility when loading files.

From what you are saying, the morphology is broken, I don't think it is NeuroM's role to repair it. I'd rather provide a CLI tool neurite-type-fixer to fix broken morphologies than open a Pandora box by allowing people to pass custom type mappings.

wizmer: > However, this will imply that features will have to be further classified by `cell_type`.
Yes, and that actually makes sense

neurom/io/swc.py
```diff
@@ -50,4 +50,8 @@ def read(filename, data_wrapper=DataWrapper):
     if len(np.shape(data)) == 1:
         data = np.reshape(data, (1, -1))
     data = data[:, [X, Y, Z, R, TYPE, ID, P]]
+    # Setting all type ids > 4 to 5 (custom section type)
+    custom_types = data[:, 4] > 4
```
eleftherioszisis, line 54: Why not use the TYPE constant you have above ( `data[:, TYPE]` )?

wizmer: good point

PR #756, opened 2019-05-20
Do not depends on add_child in tests

wizmer:
So that the tests still pass when we merge to mut_morphio branch


wizmer:
https://github.com/BlueBrain/NeuroM/pull/757 must be merged first

PR #757, opened 2019-05-20
Bump scipy and matplotlib dependencies

wizmer:
whose recent versions somehow break some tests

To be merged before #756

### DIFF
--- a/neurom/tests/test_stats.py
+++ b/neurom/tests/test_stats.py
@@ -32,6 +32,7 @@ Since the stats module consists of simple wrappers to scipy.stats functions,
 these tests are only sanity checks.
 '''
 
+import sys
 from neurom import stats as st
 from nose import tools as nt
 import numpy as np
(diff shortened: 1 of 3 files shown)


neurom/view/common.py
```diff
@@ -219,7 +219,8 @@ def plot_style(fig, ax,  # pylint: disable=too-many-arguments, too-many-locals
         ax.xaxis.set_visible(False)
         ax.yaxis.set_visible(False)
 
-    ax.set_aspect(aspect_ratio)
+    if ax.name != '3d':
```
wizmer: Without the if, matplotlib says:
```
NotImplementedError: It is not currently possible to manually set the aspect on 3D axes
```

neurom/tests/test_stats.py
```diff
@@ -166,11 +167,17 @@ def test_compare_two():
 
     results2 = st.compare_two(data, data_close, test=st.StatTests.ks)
     nt.assert_almost_equal(results2.dist, 0.5)
-    nt.assert_almost_equal(results2.pvalue, 0.5344157, places=5)
+    if sys.version_info[0] == 2:
```
wizmer: If this is not scary, I don't know what is

mgeplf: something is wrong, it's too much of a divergence; I'd be surprised if it is on scipy's side, so I think we should find the root cause before implementing a version test

wizmer: Given that scipy is the only thing that changed wrt previous commit I think it is on scipy's side. But yes, I'll investigate more.

mgeplf: do the different python versions get different scipy versions?

wizmer: Yes, 1.3.0 has not been released for python2

mgeplf: ok, that explains it: between 1.2.1 and 1.3.0, the ks_2samp underlying the test now has a `mode="auto"` param, which `# 'auto' will attempt to be exact if n1,n2 <= LARGE_N`

mgeplf: @lidakanari Do you have any idea how much the stats functionality is used?

It seems to mostly be a wrapper around scipy, and since scipy is augmenting their api, I wonder how much we should mirror it as opposed to letting the user directly call scipy.

lidakanari: Personally I am running these tests independently, but I had a few questions from users taking advantage of our scipy wrappers. However, the validation suite uses scipy and ks_2samp so I think it is worth investigating what is the impact of this change.

wizmer: Maybe we could add a deprecation notice and remove it in a few months ?

lidakanari: Is this used in the validations? If so, we need to replace it otherwise it will break the pipeline for circuit validations.

wizmer: What is the validation pipeline ? Is it a repo ?

PR #760, opened 2019-07-08
Add has_no_root_node_jumps check

wizmer:
Check that the neurites have their first point not further than
`radius_multiplier * soma radius` from the soma center'''

### DIFF
--- a/neurom/check/neuron_checks.py
+++ b/neurom/check/neuron_checks.py
@@ -223,6 +223,19 @@ def has_no_jumps(neuron, max_distance=30.0, axis='z'):
     return CheckResult(len(bad_ids) == 0, bad_ids)
 
 
+def has_no_root_node_jumps(neuron, radius_multiplier=2):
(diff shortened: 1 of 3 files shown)


wizmer:
This is required for the validator web service I've been asked to build

neurom/check/neuron_checks.py
```diff
@@ -223,6 +223,19 @@ def has_no_jumps(neuron, max_distance=30.0, axis='z'):
     return CheckResult(len(bad_ids) == 0, bad_ids)
 
 
+def has_no_root_node_jumps(neuron, radius_multiplier=2):
```
wizmer: Not sure about the name of the function

wizmer: Not sure about the name `radius_multiplier`

mgeplf: I can't think of better names

Issue #761, opened 2019-07-25
Error when plotting dendrogram using .swc from the Janelia MouseLight dataset


DrTaDa:
Hello, 

I get the following error when trying to plot the dendogram of neurons from .swc coming from the Janelia MouseLight dataset. The issue comes from the _generate_dendro function, some child only have one segment/point which leads to an empty seg_lengths list. 

I attached a .swc to the issue, you can download the rest of the dataset from http://ml-neuronbrowser.janelia.org/ if needed.
[AA0363.swc.zip](https://github.com/BlueBrain/NeuroM/files/3430540/AA0363.swc.zip)

```
import neurom as nm
from neurom import viewer
neuron = nm.load_neuron('Janelia_morphos/AA0363.swc')
viewer.draw(neuron, mode='dendrogram', show_diameters=False)
```
```
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-8-a62820b94ea3> in <module>
      3 
      4 neuron = nm.load_neuron('Janelia_morphos/AA0363.swc')
----> 5 viewer.draw(neuron, mode='dendrogram', show_diameters=False)

/anaconda3/lib/python3.7/site-packages/neurom/viewer.py in draw(obj, mode, **kwargs)
    124 
    125     output_path = kwargs.pop('output_path', None)
--> 126     plotter(ax, obj, **kwargs)
    127 
    128     if mode != 'dendrogram':

/anaconda3/lib/python3.7/site-packages/neurom/view/view.py in plot_dendrogram(ax, obj, show_diameters)
    344     # create dendrogram and generate rectangle collection
    345     dnd = Dendrogram(obj, show_diameters=show_diameters)
--> 346     dnd.generate()
    347 
    348     # render dendrogram and take into account neurite displacement which

/anaconda3/lib/python3.7/site-packages/neurom/view/_dendrogram.py in generate(self)
    198 
    199                 dummy_section.add_child(neurite)
--> 200                 self._generate_dendro(dummy_section, (max_diameter, 0.), offsets)
    201 
    202                 # store in trees the indices for the slice which corresponds

/anaconda3/lib/python3.7/site-packages/neurom/view/_dendrogram.py in _generate_dendro(self, current_section, spacing, offsets)
    259             self._max_dims = max_dims
    260             # recursive call to self.
--> 261             self._generate_dendro(child, spacing, new_offsets)
    262 
    263             # update the starting position for the next child

[...]

PR #762, opened 2019-07-25
#761 Skip 1 point sections when computing the dendrogram

wizmer:
1 point sections make the dendrogram computation crash, skip them.


wvangeit:
Ok, I had a look at how NEURON loads this. As far as I can see it loads it in a similar way, so we're good.

PR #769, opened 2019-10-01
Fix strahler.swc

wizmer:
Add a first point at the root node in order not to start with a bifurcation.
Because starting with a bifurcation in SWC is broken (but fixed in morphio)
The test was passing, but a bit by luck.


test_data/swc/strahler.swc
```diff
@@ -4,22 +4,23 @@
 # https://commons.wikimedia.org/wiki/File:Flussordnung_(Strahler).svg
```
mgeplf: can you add a comment that describes that there is now a difference from the link above?

wizmer: You mean, with the strahler orders represented in the pictures ?
Because they haven't changed. Actually, if I am not mistaken, the file is now more morphologically correct than it used to be because on the picture the branch is not starting with a bifurcation.

PR #770, opened 2019-10-07
Add 'soma_volumes' feature

asanin-epfl:
According to the request of @lidakanari


neurom/fst/_neuronfunc.py
```diff
@@ -53,6 +53,16 @@ def soma_surface_area(nrn, neurite_type=NeuriteType.soma):
     return 4 * math.pi * nrn.soma.radius ** 2
 
 
+def soma_volume(nrn, neurite_type=NeuriteType.soma):
```
wizmer: Is the neurite_type argument mandatory ?

wvangeit:
For me a sphere is ok, but it's not really how NEURON necessarily calculates the surface area (or volume). For NEURON the soma is a set of cylinders, so it will calculate the area accordingly.

neurom/fst/_neuronfunc.py
jdcourcol: yeah, that looks strange. It seems it is already like that in the other functions, so I assume @asanin-epfl is following the style.

lidakanari:
I think for all representations (unless we have 3D contour) the sphere is the most meaningful simple implementation. Maybe we could name this sphere_volume for clarity or explain it in the documentation, but I don't think we need detailed computation for this.

wizmer:
Eventually what could be done is to implement the volume for each representation if it's doable and throw an exception for the ones where it's not doable.

wvangeit:
Something like soma_sphere_volume sounds like a good idea. Only question is of course which diameter to use.

asanin-epfl:
So, should I do `soma_sphere_volume` or implement a proper computation for each type of soma? Also what about `soma_surface_area`? It is basically the same and has the same problem.

wizmer:
Go for a proper computation for each type, `soma_sphere_volume` will lead to wrong results in most cases and will lead user to think the soma is a sphere, which is far from true ! I think our most common format is soma three points, which is a cylinder. Computing the volume as if it was a sphere makes no sense. And the computation of the volume of a cylinder is trivial. Soma cylinder is a stack of frustum so it's also easy to compute. Soma sphere, well, it's a sphere. I think you just need to raise for undefined and soma contour.

lidakanari:
@wizmer Actually no data are in swc for our current use-case, this is something we need for synthesis validations. Both input data (LNMC) and synthesized use contours for soma as in h5 format. In this case, the sphere is the most appropriate implementation. 

@asanin-epfl I say we implement the feature with a precise name (_sphere) and in the future if someone needs a different implementation we re-iterate. In any case, this is also currently done for surface area, so only for consistency reasons this is the only approach that makes sense.

wizmer:
Come on, the PR is still in 'requested changes' state so no, you are not going to merge it.
I thought I had been clear in the last commit, but I'll repeat myself one more time.
If you want a soma volume implementation, it has to work on all soma types. Otherwise people are going to be mistaken and use it for soma types where it makes no sense. I know you and Lida won't be, but you are not the only users of NeuroM.

mgeplf:
From what I gather from this discussion, we need to decide where we trade-off 'correctness' vs 'simplicity' and 'in practice approximations'

In general, I consider `NeuroM` to aim to be as 'correct' as possible.  However, `soma_surface_area` shows that there are approximations (but at least it's mentioned in the docstring).

The problem with the approximations is that they are different depending on the use case: Neuro**N** finds a surface area of a cylinder, and approximates the soma volume by a sphere with the same surface area (I think this is what @wvangeit is saying w/ "Only question is of course which diameter to use.")

Because of this, I get the impression that when morphometrics are calculated, the spherical approximation is used - as was done originally for the area.

I think `soma_sphere_volume` is a good name for that approximation.

I think we should make an Issue about having a 'correct' soma_[volume|area] calculation for each of the soma types.  This complicates the interface (ie: as a user, which `_volume` function do I use?) but hopefully the docstring is clear enough.

In addition, this has the benefit of allow *comparisons* between the approximated value, and the correct value - which is something I'd like to know: how much do they actually differ in practice (ie: there are some *weird* morphologies out there).

asanin-epfl:
`soma_radii` is also wrong in that sense. Only sphere somas have single radius. Other somas don't. Unfortunately having `soma_volume` as requested by @wizmer would be misleading and inconsistent at this point. We need to fix others as well in order to do that. Which in turn affects backward compatibility. So having `soma_sphere_volume` makes sense as to have an approximated benchmark.

asanin-epfl:
I need expertise of @wvangeit and @lidakanari on this one. What formula should we use for calculating volume of point somas (SomaSinglePoint)? NEURON simulator assumes them as cylinders and uses formula `Pi * h * r^2` but we assume it as sphere and use `4/3 * Pi * r^3`. My question is what do we use for `soma_volume` feature?

wvangeit:
Mmm, interesting observation. For the surface area we didn't have to care because there the value was the same for cylinder and sphere. But indeed, the volume will change. What is the function that's used to calculate the volume in NEURON ?  I couldn't find it (at least it's not on the same page as 'area'). For now our electrical models don't use the volume normally. But this could change if we use calcium shells, or the NEURON reaction-diffusion module.
Since we were anyway gonna use the feature soma_sphere_volume, I think it makes sense to use the sphere volume.
But for the general feature we have to be more careful.

lidakanari:
I support @wvangeit comment. A one-point sphere is a neuronal morphology indicates indeed a sphere. Therefore by definition its volume should be computed based on a sphere. I'm not sure why NeuroN assumes a cylindric structure.

mgeplf:
I don't think the goal is to match what NEURON does here - a literal spherical soma's volume should be that of a sphere.  If someone needs to calculate volume like NEURON would (ie: compartment wise), they will have to do it themselves.

For instance, I vaguely remember there is logic to split the soma into multiple compartments (?), which is something that would be hard to track from NeuroM's point of view.

wvangeit:
@mgeplf my only question is with NeuroM is doing with SWC's that have a soma consisting of cylinders. As in case 4 in http://neuromorpho.org/SomaFormat.html.
Would the soma be translated into a sphere ? Or would be a set of cylinders. If the latter is the case it would be a bit 'weird' that a soma that consists of 1 cylinder would be translated into a sphere, but from 2 cylinders onward it would be using cylinder volumes.
What about somas with different L and diam in SWC ? Or these converted into spheroids, spheres or cylinders ?
(for me this weirdness applies only to the soma_volume feature, for soma_sphere_volume things are a bit clearer)

asanin-epfl:
That is how NEURON calculates volume https://github.com/neuronsimulator/nrn/blob/4ed4067b5a730c83ccf688f5ae0f61507e72856a/src/nrnpython/nrnpy_nrn.cpp#L1217
For single point case it is:
```
      Node* nd = node_exact(sec, x);
      for (Prop* p = nd->prop; p; p = p->next) {
        if (p->type == MORPHOLOGY) {
          double diam = p->param[0];
          a = M_PI * diam * diam / 4 * length;
          break;
        }
      }
```

mgeplf:
> @mgeplf my only question is with NeuroM is doing with SWC's that have a soma consisting of cylinders. As in case 4 in http://neuromorpho.org/SomaFormat.html.
> Would the soma be translated into a sphere ? 

That is precisely the crux of this ticket, IMO - like I said about trade-offs above.

To move this forward, I propose we:
1) Create the `soma_sphererized_volume` and `soma_sphererized_area` methods - these make it clear that they are approximations
2) Put a deprecation warning on `soma_area`, saying that it will become more precise in the future
3) Add `soma_volume` and a new  `soma_area` after a grace period, that calculate `more` precise values - deciding what these are can be done in another ticket.

How do people feel about that? @wizmer, @wvangeit, @lidakanari, @asanin-epfl?

wvangeit:
I'm kind of ok with that.
I would only be a bit careful with giving people the illusions things are 'precise'. 
Reality is probably somewhere between a cylinder and a sphere. The 'most precise' would rather be:
https://github.com/BlueBrain/NeuroMorphoVis/blob/master/docs/artifacts/interface-images/interface.png
(To be sure, I'm not saying NeuroM has to generate the above to calculate the area/volume)

asanin-epfl:
I'm fine with this approach. Can we use `soma_as_sphere_volume` and `soma_as_sphere_area`?

wizmer:
Hi,

Short answer because I'm in a workshop. But in this PR, I would implement the strict minimum. What I mean is that if there is not a really demand of `soma_as_sphere_volume` I would not implement it. Considering both fact that the implementation is a trivial one liner (as we already have soma.radius defined) and is an approximation, I am not sure of the added value of such a function. However, you guys think it's worth it, I'll follow your wish.

asanin-epfl:
@mgeplf and @wizmer please look. Also I'm going to create new PR for proper volume and area calculation per soma type.

wizmer:
@asanin-epfl Can you comment on my previous question ? I am still not convinced by implementing an incorrect one liner in NeuroM. People already know the volume is 4/3 Pi R^3.
I'd like to have more context about who needs this and why this is needed.

mgeplf:
I'm not @asanin-epfl, but I'd say it's useful because then there is a uniform way to dump feature statistics using the applications  like `morph_stat`.

eleftherioszisis:
Random stranger here.
I agree with @mgeplf  it will be useful to have this function for feature statistics.

asanin-epfl:
> @asanin-epfl Can you comment on my previous question ?

Please point me what exact question is that?

lidakanari:
Agree with @mgeplf and @eleftherioszisis ; it's useful for consistency to extract all features the same way. There is no point implementing part of functionality outside NeuroM.

wizmer:
Here is my proposal. It comes down to computing the volume as a shperical soma except when it is a cylindrical soma (aka SomaNeuromorphoThreePointCylinders) or a cylinder stack (aka SomaCylinder). The volume is also exposed as a Soma property so that it can be accessed on a unified way for all soma types.
https://github.com/BlueBrain/NeuroM/pull/772/files

The main point I don't like with the former proposal is ending with 2 functions `soma_sphererized_volume` and `soma_volume` that do the same thing but with different algorithms. I find it pretty confusing.

asanin-epfl:
Closing in favor of #778

PR #771, opened 2019-10-11
Fix enum34 incompatibility with Python 3.6

wizmer:
Uses the package enum-compat that only installs enum34 for python < 3.4

The error on python 3.6 was:
AttributeError: module 'enum' has no attribute 'IntFlag'

See:
https://stackoverflow.com/questions/43124775/why-python-3-6-1-throws-attributeerror-module-enum-has-no-attribute-intflag

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ from setuptools import find_packages
 VERSION = "1.4.15.dev1"
 
 REQS = ['click>=7.0',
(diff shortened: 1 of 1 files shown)


wizmer:
To reproduce the problem on Python 3.6, simply execute `pip install neurom` twice in a row.

PR #772, opened 2019-10-17
Implement different volumes for different soma types

wizmer:


asanin-epfl:
As I mentioned before here in #770 [issuecomment-541563489](https://github.com/BlueBrain/NeuroM/pull/770#issuecomment-541563489), I've made these changes already. They are going to be published as soon as #770 is resolved. Also according to our discussion I moved these calculations into a feature function as you told that neurons/morphologies will be dynamic in future and we shouldn't store `radii`, `area`, `volume` on instances of them.

asanin-epfl:
Closing in favor of #778

mgeplf:
I'll be looking at both versions, so we can get #778 merged.

PR #776, opened 2019-10-28
Accept None as an equivalent of NeuriteType.all in check_tree_type

wizmer:
As a preparation of v2 where NeuriteType.all is replaced by None

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ from setuptools import setup
 from setuptools import find_packages
 
 
-VERSION = "1.4.15.dev1"
+VERSION = "1.4.15"
 
 REQS = ['click>=7.0',
         'enum-compat>=0.0.2',


mgeplf:
When would this be needed? For grepping and such, it's nice to have named enums like already exist, so I'm not convinced this is a good idea.

wizmer:
NeuriteType will be mapped to SectionType which currently does not have a "all" type.
I actually don't like the "all" type because it is not a real type.
The alternative of course is to add "all" to morphio.

The current use case a bluepy function which has a default argument that default to NeuriteType.all and leads to a SyntaxError when importing bluepy with NeuroM v2. Actually

Something I just tried in ipython that seems to work is to augment NeuriteType with the "all" type if we really keep the backward compatibility:

```
class NeuriteType(morphio.SectionType): 
    all = 30 
```

PR #778, opened 2019-10-31
Add `soma_volumes` feature per soma type

asanin-epfl:


neurom/core/_soma.py
```diff
@@ -147,21 +158,26 @@ def __init__(self, points):
...
             'All radii must be the same'
-        # These checks were turned off after https://github.com/BlueBrain/NeuroM/issues/614
-        # assert np.isclose(points[0, COLS.Y] - points[1, COLS.Y], points[0, COLS.R]), \
-        #     'The second point must be one radius below 0 on the y-plane'
-        # assert np.isclose(points[0, COLS.Y] - points[2, COLS.Y], -points[0, COLS.R]), \
-        #     'The third point must be one radius above 0 on the y-plane'
-
-        r = points[0, COLS.R]
+        # only warn users about invalid format
+        if r < 1e-5:
```
asanin-epfl: Please be notified that I brought back here format checks for ThreePointCylinders Soma but instead of asserting, I print warnings. Please share your opinions on this.
@wvangeit @wizmer @mgeplf

wvangeit: Just want to warn that 
```
# X    Y     Z   R    P
# xs ys      zs rs   -1
# xs (ys-rs) zs rs    1
# xs (ys+rs) zs rs    1
```
is a bit of a self-imposed format. It is described by neuromorpho in the context of how to convert a neurolucida soma into swc.
But there is nothing to prevent other tools from creating a soma that consists of 3 cylinders with different diameters (and imo still be within the standard).

lidakanari: @wvangeit yes the problem is a lot of data do not follow this description. In this case (when diameters for instance are not what they are supposed to be) the volume and surface computations are wrong. I don't think we should exclude these data (still valid other that soma issues) but I asked @asanin-epfl  to add a warning to the user that in this case, the morphometrics could be wrong.

asanin-epfl:
One week reminder to review.

neurom/core/_soma.py
mgeplf: I think we should use `warnings.warn` in this case, and the other L.warnings in this file: that way they can easily be filtered (using the `warnings.catch_warnings` context manager) and they should only be output once per run of a program.

PR #780, opened 2019-11-08
Refactor dendrogram

asanin-epfl:
Refactored dendrogram in order to add synapses to it later. Will be a separate PR.


asanin-epfl:
Example outputs. You can see that dendrogram became wider. It's because I've added a horizontal padding. Probably I should remove it for backward compatibility. Also I put the leftmost section at (0, 0). This way no dendrogram is on negative X axis. This change I would insist to keep.

#### Before refactor:
denrogram_diam
denrogram_nodiam

#### After refactor:
refactored_denrogram_diam
refactored_denrogram_nodiam

mgeplf:
Cool, looks great.
>  Probably I should remove it for backward compatibility.
I think it's ok - we try and maintain backwards compat, but for diagrams it doesn't have to be exact, imo.

asanin-epfl:
Example output for the biological cell.
#### Before refactor:
denrogram_diam_bio
denrogram_nodiam_bio

#### After refactor:
refactored_denrogram_diam_bio
refactored_denrogram_nodiam_bio

neurom/view/_dendrogram.py
```diff
@@ -27,280 +27,165 @@
...
+        if dendrogram_root is None:
+            dendrogram_root = self
+            dendrogram_root.processed_section_ids = []
+        if neurom_section.id in dendrogram_root.processed_section_ids:
+            raise ValueError('Cycled morphology {}'.format(neurom_section))
+        dendrogram_root.processed_section_ids.append(neurom_section.id)
+
+        segments = neurom_section.points
+        segment_lengths = np.linalg.norm(
+            np.subtract(segments[:-1, COLS.XYZ], segments[1:, COLS.XYZ]), axis=1)
```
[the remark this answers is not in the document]
asanin-epfl: Do you mean to save it as a property? `this.segments_lengths = segment_lengths`? Currently I don't see any usage for it later.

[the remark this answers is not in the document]
mgeplf: I mean use the function that does this already `neurom/fst/_neuritefunc.py`, but called using `get()` instead of directly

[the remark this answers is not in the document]
asanin-epfl: Ah, ok. I will do the changes after all suggestions have been made.

neurom/view/_dendrogram.py
```diff
@@ -27,280 +27,165 @@
...
+        segment_radii = segments[:, COLS.R]
+
+        self.section_id = neurom_section.id
```
mgeplf: this and `self.neurite_type` aren't used, is that expected?

neurom/view/_dendrogram.py
```diff
@@ -27,280 +27,165 @@
...
-    '''Vertices of a horizontal rectangle
+        Args:
+            neurom_section: NeuroM section tree.
+            dendrogram_root: root of dendrogram. This is a service arg, please don't set it on
+            your own. It is used to track cycles in ``neurom_section``.
+        '''
+        if dendrogram_root is None:
+            dendrogram_root = self
+            dendrogram_root.processed_section_ids = []
+        if neurom_section.id in dendrogram_root.processed_section_ids:
```
mgeplf: please use an assert for these kind of checks.

Also, is this necessary? I'm not sure how a cycle would happen - without the check, the `dendrogram_root` param wouldn't be necessary, and I think the logic would look neater?

neurom/view/_dendrogram.py
```diff
@@ -27,280 +27,165 @@
...
-                     (old_offs[0], old_offs[1] + spacing[1] - diameter)))
-
-
-def _spacingx(node, max_dims, xoffset, xspace):
-    '''Determine the spacing of the current node depending on the number
-       of the leaves of the tree
+    if isinstance(neuron, Neurite):
+        return Dendrogram(neuron.root_node)
+    SomaSection = namedtuple('NeuromSection', ['id', 'type', 'children', 'points'])
+    soma_section = SomaSection(
```
mgeplf: is there a way to do this with having a 'fake' soma section?  I'm thinking that `Dendrogram()` could check the type, and then create the 'fake' data inline, instead of it being bundled into an object here, and basically decomposed in the `__init__`

neurom/view/_dendrogram.py
```diff
@@ -27,280 +27,165 @@
...
     '''
-    x_spacing = _n_terminations(node) * xspace
-
-    if x_spacing > max_dims[0]:
-        max_dims[0] = x_spacing
-
-    return xoffset - x_spacing / 2.
 
+    class _PositionedDendrogram:
+        '''Wrapper around dendrogram that allows to layout it.
```
mgeplf:
```suggestion
        '''Wrapper around dendrogram that allows to lay it out.
```

neurom/view/_dendrogram.py
```diff
@@ -27,280 +27,165 @@
...
+         in `total_width`. If no children then the width is the dendrogram's width. After the
+         calculation we start layout. Each child gets its X coordinate as:
+         parent's X + previous sibling children widths + half of this child's width.
+        '''
+        HORIZONTAL_PADDING = 2
+
+        def __init__(self, dendrogram):
+            self.dendrogram = dendrogram
+            self.children = [_PositionedDendrogram(child) for child in dendrogram.children]
+            self.origin = np.empty(2)
```
mgeplf: is this needed?

neurom/view/view.py
```diff
@@ -281,80 +282,77 @@ def plot_neuron3d(ax, nrn, neurite_type=NeuriteType.all,
...
-        # create the polygonal collection of the dendrogram
-        # segments
-        _generate_collection(group, ax, ctype, colors)
+def _as_dendrogram_line(start, end, color):
+    return FancyArrowPatch(start, end, arrowstyle='-', color=color, lw=2, shrinkA=0, shrinkB=0)
 
-    soma_square = dnd.soma
 
-    if soma_square is not None:
+def _get_dendrogram_shapes(dendrogram, positions, show_diameter):
```
mgeplf: I assume this is here instead of in `_dendrogram.py` so that `_dendrogram.py` doesn't have matplotlib dependencies?

does that mean `_dendrogram.py` should be 'public', so you can reuse the tree building mechanism for other software?

neurom/view/_dendrogram.py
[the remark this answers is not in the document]
asanin-epfl: Unfortunately I couldn't reuse `segments_lengths` function because it is for neurites. I have sections. Please look [neurom/fst/_neuritefunc.py#L204](https://github.com/BlueBrain/NeuroM/blob/master/neurom/fst/_neuritefunc.py#L204).

neurom/view/_dendrogram.py
asanin-epfl: `neurite_type` is used in `view._get_dendrogram_shapes` and `view._get_dendrogram_legend`. `section_id` is not used. I kept it to use later for synapses extension but I will remove it for now.

asanin-epfl: However, I'm using `section_id` for cycles check. If cycles check is gone then I will remove it.

neurom/view/_dendrogram.py
asanin-epfl: In the previous version we had `sys.setrecursionlimit(max_depth)`. I thought to replace this recursion limit with the cycles check. If the recursion limit didn't target cycles validation then I will remove my cycles check.

neurom/view/view.py
asanin-epfl: Yes. I intend to make it public.

neurom/view/_dendrogram.py
asanin-epfl: This is the only major issue left to discuss. I would be happy to remove the cycles check but it would be nice to have a guarantee that we can't receive a cycled Neurite tree. Otherwise we need this check. I don't know where else to put it because we need this check per dendrogram tree instance. So each node needs an access to the tree instance. `dendrogram_root` is this instance.

asanin-epfl: @wizmer Do we check that the loaded morphology does is not cycled? By the loaded morphology I mean `neurom.load_neuron('morph.swc')`. I don't see that we do it but I might overlooked it.

mgeplf: For the structure of the file disallows cycles.  For SWC, we don't support non-connected components, so I think that's where it wouldn't load properly.  I can't remember for h5.

neurom/view/dendrogram.py
```diff
@@ -0,0 +1,176 @@
...
+                left_bottom_offset = [-.5 * self.total_width, 0]
+                children_origin = end_point + left_bottom_offset
+                for child in self.children:
+                    child_origin = children_origin + [.5 * child.total_width, 0]
+                    positions.update(child.position_at(child_origin))
+                    children_origin += [child.total_width + self.HORIZONTAL_PADDING, 0]
+            return positions
+
+    pos_dendrogram = _PositionedDendrogram(dendrogram)
+    return pos_dendrogram.position_at(origin)
```
wizmer:
```suggestion
    return _PositionedDendrogram(dendrogram).position_at(origin)
```

neurom/view/_dendrogram.py
wizmer: I don't think the `sys.setrecursionlimit(max_depth)` was for validation, but probably because without it it would crash for neurites with too many descendents. 

I just tried on SWC with the following file and NeuroM did not complain although it should have, scary:
```
    1 1  0  0 0 1. -1
    2 3  0  0 0 1.  1
    3 3  0  5 0 1.  2
    4 3 -5  5 0 0.  3
    5 3  6  5 0 0.  3
    6 2  0  0 0 1.  6  # <-- cyclic point
    7 2  0 -4 0 1.  6
    8 2  6 -4 0 0.  7
    9 2 -5 -4 0 0.  7
```
The `mut_morphio` branch already raises on this file so I don't know if it's worth an implementation.

neurom/view/dendrogram.py
```diff
@@ -0,0 +1,176 @@
...
+    '''Move positions to a new origin.
+
+    Args:
+        positions (dict of Dendrogram: np.array): positions
+        to_origin (np.array): where to move. np.array of (2,) shape for x,y coordindates.
+
+    Returns:
+        Moved positions.
+    '''
+    to_origin = np.array(to_origin)
```
wizmer:
```suggestion
    to_origin = np.asarray(to_origin)
```
will prevent a copy if `to_origin` is already a numpy array.

neurom/view/dendrogram.py
```diff
@@ -0,0 +1,176 @@
...
+                type=NeuriteType.soma,
+                children=[neurite.root_node for neurite in neurom_section.neurites],
+                points=np.array([
+                    np.array([0, 0, 0, .5]),
+                    np.array([.1, .1, .1, .5]),
+                ])
+            )
+        if dendrogram_root is None:
+            dendrogram_root = self
+            dendrogram_root.processed_section_ids = []
```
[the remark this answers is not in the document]
asanin-epfl: Here I would keep `dendrogram` to emphasize that there is a tree head and `self` is this head now.

neurom/view/dendrogram.py
```diff
@@ -0,0 +1,176 @@
...
+        assert neurom_section.id not in dendrogram_root.processed_section_ids
+        dendrogram_root.processed_section_ids.append(neurom_section.id)
+
+        segments = neurom_section.points
+        segment_lengths = np.linalg.norm(np.diff(segments[:, COLS.XYZ], axis=0), axis=1)
```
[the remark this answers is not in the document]
asanin-epfl: `path_distance` returns a single distance from one point to another. I need a list of distances among each neighbour point in the path.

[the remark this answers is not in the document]
wizmer: Ah, my bad, we really need to add this functionality in NeuroM.

neurom/view/dendrogram.py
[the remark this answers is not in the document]
asanin-epfl: I dropped the cycles check because it is implemented in `mut_morphio` branch and the NeuroM core is a more appropriate place for this check.

asanin-epfl:
I removed the cycles check and fake SomaSection. @mgeplf and @wizmer please verify.

asanin-epfl:
> Re-run the notebook in tutorial/neurom_tutorial.ipynb so that it displays the new-style dendrogram.

@wizmer I ran it and it worked fine. If you see a problem please say it directly.

wizmer:
It works, but the picture of the dendrogram cell #6 is the one from the old algo, so I just re-run this cell and update the notebook.

asanin-epfl:
> It works, but the picture of the dendrogram cell #6 is the one from the old algo, so I just re-run this cell and update the notebook.

Thank you for the details. Fixed it. I reran the first two cells for imports + #6 cell.

PR #781, opened 2019-11-18
Bump version to 1.4.15

asanin-epfl:


asanin-epfl:
> Going to have to delete/recreate the tags, I guess.

Yes, sorry about that.

Issue #768, opened 2019-08-19
Plotting neuron if axon originates from dendrite


INF-Rene:
In the viewer, the neurites are colored by it's root section type, so in neurons with the axon originating at the dendrite, the code below plots the entire neuron:
```
fig=plt.figure()
ax=plt.gca()
nm.view.view.plot_neuron(ax, nrn, neurite_type=NeuriteType.basal_dendrite, color='red')
ax.autoscale()
```
and 
`nm.view.view.plot_neuron(ax, nrn, neurite_type=NeuriteType.axon, color='red')
` plots nothing but the soma

This is because the plotting method iterates over neurites, which types are determined by the root section. 
Would it be an idea to let the `nm.view.view.plot_neuron` method iterate over sections instead of neurites?

wizmer:
Yes, that's something I have thought about for a while. I'll work on it this week.

INF-Rene:
@wizmer , awesome, let me know if there is something I can do to help

PR #787, opened 2020-01-03
The section color is chosen section by section

wizmer:
And is no longer determined only by the neurite type
Issue #768

### DIFF
--- a/neurom/view/tests/test_view.py
+++ b/neurom/view/tests/test_view.py
@@ -30,11 +30,12 @@ import os
 
 import itertools as it
 import numpy as np
+from numpy.testing import assert_array_almost_equal
 
 from nose import tools as nt
 from neurom import load_neuron
 from neurom.view import view, common
-from neurom.core import Section
+from neurom.core import Section, Neurite
 from neurom.core._soma import make_soma, SOMA_CONTOUR, SOMA_CYLINDER
 from neurom.core.types import NeuriteType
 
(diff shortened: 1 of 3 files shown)


neurom/view/tests/test_view.py
```diff
@@ -30,11 +30,12 @@
 
 import itertools as it
 import numpy as np
+from numpy.testing import assert_array_almost_equal
```
asanin-epfl: Can we use `np.testing.assert_array_almost_equal` instead of importing it separately? Otherwise should we import separately `assert_allclose`? Currently we use `np.testing.assert_allclose`

neurom/view/tests/test_view.py
```diff
@@ -46,12 +47,19 @@
...
 
 def test_tree():
-    with get_fig_2d() as (fig, ax):
-        tree = fst_neuron.neurites[0]
-        view.plot_tree(ax, tree, color='black', diameter_scale=None, alpha=1., linewidth=1.2)
-        collection = ax.collections[0]
-        nt.eq_(collection.get_linewidth()[0], 1.2)
-        np.testing.assert_allclose(collection.get_color(), np.array([[0., 0., 0., 1.]]))
+    neuron = load_neuron(os.path.join(SWC_PATH, 'simple-different-section-types.swc'))
+    expected_colors = {'black': np.array([[0., 0., 0., 1.] for _ in range(3)]),
```
asanin-epfl: What about `np.repeat([[0,0,0,1]], 3, axis=0)`?

neurom/view/tests/test_view.py
```diff
@@ -46,12 +47,19 @@
...
+                       None: [[1.      , 0.      , 0.      , 1.],
```
asanin-epfl: Can the formatting be improved here? Looks hard to read in my editor.

neurom/view/tests/test_view.py
wizmer: I find this more easily understandable than having to reason about numpy axes.

PR #788, opened 2020-01-03
Suppress warnings from Tox output

wizmer:


neurom/io/neurolucida.py
```diff
@@ -262,12 +262,9 @@ def read(morph_file, data_wrapper=DataWrapper):
     suitable to be wrapped by DataWrapper
     '''
 
-    msg = ('This is an experimental reader. '
-           'There are no guarantees regarding ability to parse '
-           'Neurolucida .asc files or correctness of output.')
-
-    warnings.warn(msg)
-    L.warning(msg)
```
asanin-epfl: Drop also
```
L = logging.getLogger(__name__)
```
and
```
import logging
```

neurom/fst/tests/test_sectionfunc.py
```diff
@@ -100,14 +101,8 @@ def test_section_tortuosity_looping_section():
         (0, 0, 0), (1, 0, 0), (1, 2, 0), (0, 2, 0), (0, 0, 0)
     ])
 
-    # is infinity
-    nt.eq_(_sf.section_tortuosity(sec), np.inf)
-    nt.ok_(math.isinf(_sf.section_tortuosity(sec)))
-    nt.ok_(np.isinf(_sf.section_tortuosity(sec)))
-
-    # is not NaN
-    nt.ok_(not math.isnan(_sf.section_tortuosity(sec)))
```
asanin-epfl: Is it ok to drop `NaN` tests?

neurom/io/neurolucida.py
wizmer: Good catch

neurom/fst/tests/test_sectionfunc.py
wizmer: This function was testing `_sf.section_tortuosity(sec)` against various types of NaN and inf, but since the returned value is np.inf, I feel it was more testing Numpy than NeuroM

PR #792, opened 2020-01-24
additional features

wizmer:
corrections + test


neurom/features/bifurcationfunc.py
```diff
@@ -120,8 +121,48 @@ def partition_pair(bif_point):
...
     The number of nodes in each child tree is counted. The partition
-    pairs is the number of bifurcations in the two daughter subtrees
+    pairs is the number of bifurcations in the two child subtrees
     at each branch point.'''
     n = float(sum(1 for _ in bif_point.children[0].ipreorder()))
     m = float(sum(1 for _ in bif_point.children[1].ipreorder()))
     return (n, m)
+
+
+def sibling_ratio(bif_point, method='first'):
```
wizmer: I still wonder if this should be called `sibling_diameter_asymmetry` ?

neurom/features/neuritefunc.py
```diff
@@ -316,8 +333,22 @@ def bifurcation_partitions(neurites, neurite_type=None):
                              neurite_filter=is_type(neurite_type)))
 
 
-def partition_asymmetries(neurites, neurite_type=None):
-    '''Partition asymmetry at bifurcation points of a collection of neurites'''
+def partition_asymmetries(neurites, neurite_type=None, variant=None):
```
wizmer: I don't think default variant should be None

wizmer:
@arnaudon could you add a unit test for section_taper_rate ?

neurom/features/bifurcationfunc.py
arnaudon: mmmh this is the name @lidakanari used, but indeed, I can't find it after a quick google search, so there may be a better name. It is not mentionned here neither: https://www.nature.com/articles/nrn2402.pdf . @lidakanari , what would be the best name?

neurom/features/neuritefunc.py
arnaudon: Shall I put a different name for each variant? I was thinking not to call the original definition by a variant name, but I can put variant='original' if you prefer.

PR #793, opened 2020-01-24
Fix dangling branch check

wizmer:
One should look at the first point of the neurite and not the second one.

Honestly, I am a bit puzzled as for why it was this way before. There might have been a reason ?


lidakanari:
Maybe they have assumed duplicate points? But I think your proposal is the correct one, there are no duplicates at the beginning of a tree.

PR #794, opened 2020-01-24
Reconcile mut_morphio and master a bit more

wizmer:
Bring back some harmless changes that were done in the mut_morphio branch
in the master branch.
The goal is to decrease the differences between the two branches.

### DIFF
--- a/neurom/check/tests/test_morphtree.py
+++ b/neurom/check/tests/test_morphtree.py
@@ -113,7 +113,7 @@ def _genetate_tree_non_monotonic_section_boundary():
 def _generate_back_track_tree(n, dev):
     points = np.array(dev) + np.array([1, 3 if n == 0 else -3, 0])
 
-    neuron = load_neuron(StringIO('''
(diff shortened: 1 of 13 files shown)


neurom/fst/tests/test_sectionfunc.py
```diff
@@ -70,13 +71,21 @@ def test_section_area():
 
 
 def test_section_tortuosity():
-    sec_a = Section([
```
wizmer: NeuroM v2 won't have a Section ctor so we need to get rid of its use in the tests.

neurom/check/tests/test_morphtree.py
```diff
@@ -110,28 +111,31 @@ def _genetate_tree_non_monotonic_section_boundary():
 
 
 def _generate_back_track_tree(n, dev):
-
```
wizmer: NeuroM v2 won't have a Section ctor so we need to get rid of its use in the tests.

PR #795, opened 2020-01-24
Moving features in features folder

wizmer:
- All features are moved to features/
- The sub-packages loose their underscore.
    As a reminder, `sectionfunc.py` had no underscore while `_bifurcationfunc.py`, `_neuritefunc.py` and `_neuronfunc.py` had one.

Example:
```python
from neurom.fst import _bifurcationfunc
```
can **also** be called through:
```python
from neurom.features import bifurcationfunc
```

The final goal is to get rid of the fst/ folder.


wizmer:
The diff is not very clear but what I did is:
move the files from fst/ to features/
Recreate empty files in fst/ where I do `from features.filename import *`

neurom/features/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+''' NeuroM, lightweight and fast
+
+Examples:
+
+    Obtain some morphometrics
+
+    >>> ap_seg_len = fst.get('segment_lengths', nrn, neurite_type=neurom.APICAL_DENDRITE)
```
haleepfl: No longer recommended usage with `fst`?

neurom/fst/__init__.py
```diff
@@ -1,177 +1,7 @@
-# Copyright (c) 2015, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
```
haleepfl: We should still retain this license shouldn't we?

neurom/fst/__init__.py
```diff
@@ -1,177 +1,7 @@
...
-    ret.append('\nNeuron features (neuron, neuron population):')
-    ret.extend(_INDENT + '- ' + feature + get_docstring(func)
-               for feature, func in sorted(NEURONFEATURES.items()))
-
-    return '\n'.join(ret)
-
-
-get.__doc__ += _indent('\nFeatures:\n', 1) + _indent(_get_doc(), 2)  # pylint: disable=no-member
+from neurom.features import NEURITEFEATURES, NEURONFEATURES, get, register_neurite_feature
+print('neurom.fst is being deprecated, please consider replacing it by neurom.features')
```
haleepfl: A couple of suggestions here and for other similar locations:

- Should this be a warning?
- If possible, note when it will be removed: next major version bump?
- Remove 'please consider' and just say what they need to change

pylintrc
```diff
@@ -17,7 +17,8 @@
 #W0622 - redefined-builti
 #C0325 - superfluous-parens
 #R0205 - useless-object-inheritance
-disable=C0103,R0904,R0903,W0511,W0232,R0922,R0801,R0921,W0141,R0401,I0013,W0142,W0622,C0325,R0205
+#W0614 - unused-wildcard-import
```
haleepfl: Why add this here if they are disabled inline? It might mask other bad usage of wildcard imports in the future

neurom/features/__init__.py
```diff
@@ -0,0 +1,176 @@
+# Copyright (c) 2015, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
```
haleepfl: Update all

neurom/fst/__init__.py
wizmer: From what I understood, having the licence at the root of your project is enough to protect it.

neurom/features/__init__.py
wizmer: You mean the date ?

neurom/fst/__init__.py
jdcourcol: no, it has to be in every file.

wizmer: Is it a BlueBrain directive ? Because most of the resources I've found on the internet seem to indicate the opposite. 
https://stackoverflow.com/questions/845895/putting-license-in-each-code-file
https://github.com/NethermindEth/nethermind/issues/1067

wizmer: Nevermind, @haleepfl showed me our internal directives where it is clearly specified.

wizmer: Well, well, our internal guidelines are not very clear about BSD. I have created a ticket about this discussion here: https://github.com/BlueBrain/NeuroM/issues/799

wizmer:
Good question @asanin-epfl 
What I have done is to duplicate the fst/tests in features/tests
I keep using the old way:
```from neurom.fst import _bifurcationfunc as _bf```
 in `fst/tests`
and I use the new way:
```from neurom.features import bifurcationfunc as _bf```
 in `features/tests`

**Comment by asanin-epfl** [comment]
My questions have been answered but why duplicating tests?

**Comment by wizmer** [comment]
To tests that old way and new way are both working. And we'll remove the old way when we remove the fst/ folder. I'm open to an alternative suggestion though.

**Comment by asanin-epfl** [comment]
Ok, my questions have been resolved. What about others?

File: neurom/fst/tests/test_bifurcationfunc.py [hunk 6]
```diff
@@ -26,58 +26,11 @@
...
-    root = SIMPLE2.neurites[0].root_node
-    assert_equal(bf.partition_asymmetry(root), 0.5)
-    assert_equal(bf.partition_asymmetry(root.children[0]), 0.0)
-
-    leaf = root.children[0].children[0]
-    assert_raises(NeuroMError, bf.partition_asymmetry, leaf)
-
-    multifurcation_section = MULTIFURCATION.neurites[0].root_node.children[0]
-    assert_raises(NeuroMError, bf.partition_asymmetry, multifurcation_section)
+from neurom.fst import _bifurcationfunc as bifurcation
```
mgeplf: Could you:
a) have a comment explaining how this override works (in all places it's used)
b) in `neurom.features.tests.test_bifurcationfunc` point to all the files that need to be updated if one of the names that are being over-written here changes - w/ a TODO to say remove it when these tests go away.

PR #800, opened 2020-01-28
Use plotly in tutorial notebook

haleepfl:
This PR updates the main tutorial notebook to use plotly instead of matplotlib. It also moves the required Binder config files out of root, and into the .binder subdirectory.

To test the new notebook: https://mybinder.org/v2/gh/haleepfl/NeuroM.git/plotly-notebook?filepath=tutorial%2Fneurom_tutorial.ipynb


wizmer:
Two things:
- The GitHub preview used to show the matplotlib (https://github.com/BlueBrain/NeuroM/blob/master/tutorial/neurom_tutorial.ipynb), it seems to be unable to render the plotly one (https://github.com/haleepfl/NeuroM/blob/plotly-notebook/tutorial/neurom_tutorial.ipynb). I guess this is no big deal but worth noting.
- I find it a bit weird to use an 'extra' as the main. And there is nothing more frustrating than a tutorial that does not work out of the box. What you could do is always use `draw` when calling the function instead of `plotly.draw`. And in the first cell something around:

```python
try:
    from neurom.plotly import draw
except:
    print('It seems you haven't installed the extra package neurom['plotly']'
             'You can install it via...'
             'The matplotlib viewer will be used')
    from neurom import draw
```

wizmer:
Could you rename the notebook to `getting-started.ipynb` or something in those lines ?

jdcourcol:
>     * I find it a bit weird to use an 'extra' as the main. And there is nothing more frustrating than a tutorial that does not work out of the box. What you could do is always use `draw` when calling the function instead of `plotly.draw`. And in the first cell something around:

The origin of the ticket is to have the mybinder notebook (where everything is installed) to showcase what NeuroM is capable of. So showing the beautiful plotly is better than the default one in that context.

PR #807, opened 2020-02-03
All checks should return list of points

wizmer:
Even if there is only one point to return


neurom/check/neuron_checks.py
```diff
@@ -273,7 +273,7 @@ def has_no_narrow_start(neuron, frac=0.9):
     Returns:
         CheckResult with a list of all first segments of neurites with a narrow start
     '''
-    bad_ids = [(neurite.root_node.id, [neurite.root_node.points[1]])
+    bad_ids = [(neurite.root_node.id, neurite.root_node.points[np.newaxis, 1])
```
mgeplf: How does this interact with the `morph_check` output summary?

wizmer: The output went from:
```
DEBUG: Has no narrow start: 1 failing ids detected: [(1, [array([0., 0., 2., 2., 3., 3., 2.])])]
```
to:
```
DEBUG: Has no narrow start: 1 failing ids detected: [(1, array([[0., 0., 2., 2., 3., 3., 2.]]))]
```
Is this your question ?

asanin-epfl:
Looks good to me but I will let Mike to resolve his questions.

neurom/check/neuron_checks.py
mgeplf: Ah, I thought we output some sort of json report or something.  I prefer stability for those types of outputs, but hopefully not many people will have to change their code in this case.

PR #808, opened 2020-02-05
Fix coverage

wizmer:
Coverage is back to 100%

One `continue` is not being covered even though it is being run in the tests.
It is probably a 'coverage' bug so I added a TODO for later

I also changed the 'show_diameter' argument to 'show_diameters' in
_get_dendrogram_shapes to be consistent with the argement of the function
plot_dendrogram


mgeplf:
> One continue is not being covered even though it is being run in the tests.

iirc, that's a known thing w/ coverage - that's fine

PR #809, opened 2020-02-05
Travis build now fails if coverage is less than 100%

wizmer:
The coverage command was performed in stage 'after_success' but its return
code is ignored so it was basically doing nothing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,11 @@ dist: trusty
 language: python
 python:
   - 2.7
-  - 3.5
   - 3.6
+  - 3.7
 cache: pip
 install:
-  - pip install 'matplotlib==1.5.2' tox-travis
-  - pip install .
+  - pip install tox-travis
 script:
   tox
 before_deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,23 +4,12 @@ python:
   - 2.7
   - 3.5
   - 3.6
-addons:
-  apt:
-    packages:
-    - libatlas-dev
-    - libatlas-base-dev
-    - liblapack-dev
-    - gfortran
-    - libmpfr-dev
-    - libhdf5-dev
 cache: pip
 install:
-  - pip install 'matplotlib==1.5.2' tox
+  - pip install 'matplotlib==1.5.2' tox-travis
   - pip install .
 script:
-  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then tox -e py27; fi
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -18,11 +18,9 @@ install:
   - pip install 'matplotlib==1.5.2' tox
   - pip install .
 script:
-  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then tox -e py27; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 2.7 ]]; then tox -e py27 -e coverage; fi
   - if [[ $TRAVIS_PYTHON_VERSION == 3.5 ]]; then tox -e py35; fi
   - if [[ $TRAVIS_PYTHON_VERSION == 3.6 ]]; then tox -e py36 -e lint; fi
```
mgeplf: should we be doing the coverage on a newer python?

wizmer: Good point. I changed that.

wizmer:
I also changed also the default tox env to python3, hence the changes in the source files.

mgeplf:
Yup, great idea.  We could probably add newer python versions too; but we can do that in another PR, if you want.

wizmer:
I'll take the opportunity of the failing Travis build to try some new stuff. It may take a while. You can unsubscribe from this PR if you want. I'll remove you from the reviewers.

.travis.yml
```diff
@@ -1,28 +1,13 @@
...
-    - gfortran
-    - libmpfr-dev
-    - libhdf5-dev
+  - "2.7"
+  - "3.6"
+  - "3.7"
 cache: pip
 install:
-  - pip install 'matplotlib==1.5.2' tox
-  - pip install .
```
wizmer: I'm wondering if this line is required for the deployement stage ?

wizmer:
@mgeplf That should be good now.
I have added thet use of tox-travis to manage the Travis tox envs.
Also I replaced the py35 by py37.

tox.ini
```diff
@@ -7,43 +7,37 @@ testdeps =
...
 [tox]
 envlist =
-    lint
-    py27
-    py36
-    coverage
+    {py27,py36}
+    py36-lint
+    py36-coverage
+    py36-docs
```
wizmer: Does the 'docs' env name matters for readthedoc ?

haleepfl: readthedocs doesn't use tox, so it shouldn't be a problem

.travis.yml
mgeplf: iirc, pinning the matplotlib version was to make sure the tests passed

PR #810, opened 2020-02-10
Load in custom type neurite swc

AF1111:
This fixes issue BlueBrain#735 by loading in custom type (!=1-4) neurites with load_neuron. Specifically, this allows for the loading of janelia mouselight consensus swc data which is not loaded with a soma on the first entry nor does it contain the standard neurite labeling(only has 0,5,6 labels). The soma is set as the first entry to avoid the out of range index or SomaErrors.


wizmer:
Hi,

Thanks for you PR. I took the part relevant to the section type and merged it in #736.
We are not to sure we want to support files without a soma as that could lead to an undefined behavior. That's why I did not merge this part. Still thanks, your PR made me save some time for the custom type fix !

PR #811, opened 2020-02-12
Remove all uses of make_soma in tests

wizmer:
To make the migration to morphio easier

### DIFF
--- a/neurom/core/tests/test_soma.py
+++ b/neurom/core/tests/test_soma.py
@@ -26,158 +26,151 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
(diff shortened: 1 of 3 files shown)


neurom/core/tests/test_soma.py
```diff
@@ -26,184 +26,175 @@
...
 
     # cylinder: h = 10, r = 20
-    soma_2pt_normal = np.array([
-        [0.0,   0.0,  0.0, 20.0, 1, 1, -1],
-        [0.0, -10.0,  0.0, 20.0, 1, 2,  1],
-    ])
-    s = _soma.make_soma(soma_2pt_normal, soma_class=_soma.SOMA_CYLINDER)
+    s = load_neuron(StringIO(u'''
+                1 1 0   0 0 20 -1
+                2 1 0 -10 0 20  1'''), reader='swc').soma
```
wizmer: I'm now wondering if those 2 point cylinder correspond to a valid configuration ?

asanin-epfl: I don't know to be honest

PR #818, opened 2020-03-12
Make Neuron class a public member

wizmer:
In order to be able to use `isinstance` or type hints

### DIFF
--- a/neurom/__init__.py
+++ b/neurom/__init__.py
@@ -59,14 +59,15 @@ Examples:
 '''
 
 import logging as _logging
-from .version import VERSION as __version__
-from .core import iter_neurites, iter_sections, graft_neuron, iter_segments, NeuriteType
+
+from .core import (NeuriteType, graft_neuron, iter_neurites, iter_sections,
+                   iter_segments)
+from .core._neuron import Neuron
 from .core.dataformat import COLS
 from .core.types import NEURITES as NEURITE_TYPES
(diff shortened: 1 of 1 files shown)


mgeplf:
Can you explain why it needs to be public in the commit msg?

wizmer:
Ok, I will put what I've just added in the description of the PR

neurom/__init__.py
```diff
@@ -59,13 +59,15 @@
 '''
 
 import logging as _logging
-from .version import VERSION as __version__
-from .core import iter_neurites, iter_sections, graft_neuron, iter_segments, NeuriteType
+
+from .core import (NeuriteType, graft_neuron, iter_neurites, iter_sections,
+                   iter_segments)
+from .core._neuron import Neuron
```
asanin-epfl: Why not include it in the upper import? In `core/__init__.py` we have `from ._neuron import (Section, Neurite, Neuron, `

wizmer: That's a good question. I actually don't know what is the proper Python standard policy but I thought since Neuron is really a base object of NeuroM it deserved to be avalailable from the top level.

asanin-epfl: So, can we use a single import statement with `Neuron` included?:
```
from .core import (NeuriteType, Neuron, graft_neuron, iter_neurites, iter_sections,
                   iter_segments)
```

PR #819, opened 2020-03-19
Documentation build and config cleanup

haleepfl:


setup.py
```diff
@@ -51,10 +51,7 @@
 # Hack to avoid installation of modules with C extensions
 # in readthedocs documentation building environment.
 if os.environ.get('READTHEDOCS') == 'True':
-    REQS = ['future>=0.16.0',
-            'enum34>=1.0.4',
-            'pyyaml>=3.10',
-            ]
+    REQS = []
```
wizmer: Do we still need this at all even with `autodoc_mock_imports` ?

haleepfl: This seems to be causing issues in NeuroR since the packages are not mocked, but NeuroM is being installed on ReadTheDocs. I'll remove it and see if the NeuroM docs build succeeds.

PR #822, opened 2020-03-26
path_distance issue in section_taper

arnaudon:
In the last PR, I made a mistake in computing the path distances for taper rates of sections.


eleftherioszisis:
No test?

arnaudon:
There is one, but it passes as it uses SIMPLE neuron, which only section with two segments. I guess we could add a more complex neuron for this test. @wizmer ?

eleftherioszisis:
One section with two segments? Is this SIMPLE morphology used for all features? If it is then all the features that make calculations taking branching into account are not tested.

arnaudon:
Nah, it has branches as it has several sections. The modif I made does not matter if you have sections with two segments, only if you have more. That's why the test still passes.

eleftherioszisis:
Ah I see. In that case a slightly more complex neuron that fails in the old case should be added.

wizmer:
Agreed !

arnaudon:
Where can I find one? I can modify the test now.

wizmer:
You can take test_data/swc/Neuron.swc

PR #823, opened 2020-04-01
Fix dangling axons

wizmer:
Since axon can start from dendritic trees at any location,
the check must include all dendritic points.
Not only `iter_neurites(axon)` but `iter_neurites(neuron)`

Also fixed the returned point. It was returning the 2nd point of the
dangling neurite instead of the 1st one.


neurom/check/tests/test_neuron_checks.py
```diff
@@ -401,12 +404,23 @@ def test_has_no_narrow_dendritic_section():
...
-    nrn_chk.has_no_dangling_branch(nrn)
-    nt.ok_(not nrn_chk.has_no_dangling_branch(nrn).status)
+    res = nrn_chk.has_no_dangling_branch(nrn)
+    nt.ok_(not res.status)
+    nt.assert_equal(len(res.info), 1)
+    assert_array_equal(res.info[0][1][0][COLS.XYZ],
+                       [0., 49.,  0.])
+
+    _, nrn = _load_neuron('axon-sprout-from-dendrite.asc')
+    status = nrn_chk.has_no_dangling_branch(nrn)
```
asanin-epfl, line 421: Can you please use here `res` as above? So they all in the same style

PR #827, opened 2020-04-22
Cleanup docstrings for pydocstyle

haleepfl:
There are currently still some failures, however I need some additional input on these to make sure the fixes are sensible.

Note: there are some checks disabled in the tox.ini config (e.g. triple double quotes, and line endings), these are disabled due to the sheer volume of changes required. We can decide if we should address these as part of this PR.


tox.ini
```diff
@@ -41,3 +43,11 @@ whitelist_externals = make
 
 [pycodestyle]
 max-line-length=100
+
+[pydocstyle]
+# ignore the following
+#   - D300: found ''' not """
+#   - D413: no blank line afer last section
+#   - D415: First line should end with a period, question mark, or exclamation point
+add-ignore = D300, D413, D415
```
wizmer, line 52: I'll take care of D415

PR #828, opened 2020-04-30
Add option to scale diameters with axis + bug in soma contour

arnaudon:
This PR contains the following improvements:

1. updated docstrings to comply with pydocstyle
2. with soma_contour=False, the plt.plot(points) was not working as points had the wrong shape (possible clash with newest matplotlib version). It was points = [(0,1), (1,2)], but needs to be points = [0, 1], [1, 2] 
3. added an option to scale linewidths (diameters of dendrites) according to data coordindate, such that they look realistic, for any figsize, or axes limits.

For 3. there are 2 outstanding issues which are partially fixed:

1. we compute the difference of positions using ax.transData.transform of matplotlib in x an y directions, which give slightly different results. We take the mean for now, as an approximation of true value. Because these values are very close to each other, this should not affect the outcome by much
2. Matplotlib has some internal 'stale' state, which needs to be updated by hand with ax._unstale_viewLim() , as we compute the scaling before the first add_collection is called, where this function is called by default in matplotlib. This works as it is, but may be worth further investigation, an eventually an issue/PR on matplotlib. I used the current github version of matplotlib for that test. 

PS: I found the hack for 3 in: https://stackoverflow.com/questions/19394505/matplotlib-expand-the-line-with-specified-width-in-data-unit/42972469#42972469


neurom/view/view.py
```diff
@@ -163,14 +158,39 @@ def plot_soma(ax, soma, plane='xy',
...
+
+    WARNING: this works best if both axis are on the same scale, and will not follow
+    a dynamics change of figure size.
+
+    Args:
+        ax(matplotlib axes): on what to plot
+        scale_diameter_with(str): axis along which to scale diameter for
+                                  correct unit (None, 'x' or 'y')
+
+    Returns: flaot
```
wizmer, line 172:
```suggestion
    Returns (float): describe whatever it returns
```

neurom/view/view.py
```diff
@@ -163,14 +158,39 @@ def plot_soma(ax, soma, plane='xy',
...
 
+def _get_scaled_linewidth(ax, scale_with=None):
+    """Get linewidth to scale diameters according to the axis of current figure.
+
+    WARNING: this works best if both axis are on the same scale, and will not follow
+    a dynamics change of figure size.
+
+    Args:
+        ax(matplotlib axes): on what to plot
+        scale_diameter_with(str): axis along which to scale diameter for
```
wizmer, line 169:
```suggestion
        scale_with(str): axis along which to scale diameter for
```

neurom/view/view.py
```diff
@@ -25,7 +25,7 @@
 # ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-"""Visualize morphologies."""
```
wizmer, line 28: Can you disable the """ -> ''' and keep the docstrings with the capital letter as first char and period at the end ?
Thanks

neurom/view/view.py
```diff
@@ -182,10 +202,13 @@ def plot_neuron(ax, nrn,
         linewidth(float): all segments are plotted with this width, but only if diameter_scale=None
         color(str or None): Color of plotted values, None corresponds to default choice
         alpha(float): Transparency of plotted values
-    """
+        scale_diameter_with(str): axis along which to scale diameter for
+                                  correct unit (None, 'x' or 'y')
```
wizmer, line 206: Is there a way to have the scaling be always correct without having to provide this additional argument ?

wizmer:
Can you add some explanation about the bug in the commit message. That will help me review.

arnaudon:
I think coverage fails, I can try to fix that if you want by adding a test.

neurom/view/view.py
```diff
@@ -163,14 +158,38 @@ def plot_soma(ax, soma, plane='xy',
                                    ignore=False)
 
 
+def _get_scaled_linewidth(ax, scale_with=None):
+    """Get linewidth to scale diameters according to the axis of current figure.
+
+    WARNING: this works best if both axis are on the same scale, and will not follow
```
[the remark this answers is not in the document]
arnaudon: ah yes, I should remove the 'axis on the same scale', this was for the first implementation I made, now it works independently of that. For your second point, fixing it would involve more coding (see the link in the main description of the PR), as lines need to be redrawn if the user rescale things on the fly... this option is not on by default, and should be used safely once we just plt.savefig. If the user start modifying the figure, diameter will not follow properly (as it has always been the case).

arnaudon:
For the collage example I was making, the cells had diameters 4 times the correct thickness. I let you imagine how the figure looked like.

neurom/view/view.py
```diff
@@ -95,6 +124,7 @@ def plot_tree(ax, tree, plane='xy',
         linewidth(float): all segments are plotted with this width, but only if diameter_scale=None
         color(str or None): Color of plotted values, None corresponds to default choice
         alpha(float): Transparency of plotted values
+        scale_with_axis(bool): scale linewidths with axis data coordinates
```
wizmer, line 127: At the level of this public function, I would prefer a simpler argument name that appeals to anyone. Maybe `realistic_diameters` ?

wizmer:
It's not a coverage problem but a unit tests one:
https://travis-ci.org/github/BlueBrain/NeuroM/jobs/681875915#L1062

arnaudon:
Indeed, the hidden matplotlib function is not there in previous versions, so I just added a try: except. I think it is not needed for previous versions. I found it by just printing ax.get_xlim(), which calls that function  for latest matplotlib. Maybe better to just call ax.get_xlim() in stead of try: and the calling this obscure hidden function? 

For the test the value is a bit large, not sure why. It works on my use case, see figure:
[example_realistic.pdf](https://github.com/BlueBrain/NeuroM/files/4587728/example_realistic.pdf)
[example_not_realistic.pdf](https://github.com/BlueBrain/NeuroM/files/4587729/example_not_realistic.pdf)

arnaudon:
lint does not like bare except, but not sure how to best do that, I'll follow your suggestions.

wizmer:
Maybe it's time to say farewell to the Python 2 support ?
I wouldn't mind.

arnaudon:
Ahah, it seems the every time I do a PR on a new code, I have to drop python2 for some reason ;)

arnaudon:
let's move to python2 in another PR, I don't wan to take responsability for that haha

neurom/view/view.py
```diff
@@ -72,12 +72,9 @@ def _scale_linewidth_with_axis(ax, linewidth):
         linewidth (float):  linewidth in data coordinates
     Returns (float): rescaled linewidth
     """
-    try:
+    if sys.version_info.major == 3:
         # this is done after first add_collection, but we need it here for latest matplotlib
```
mgeplf, line 76: I think the `latest matplotlib` comment will go out date too fast, so some description of what is going on, and why it's needed, IMO

arnaudon: just added more details. I was thinking of puting an issue on matplotlib, as I think it should be fixable from matplotlib side. I'll try to make a toy example that fails to send it to them and see what they say.

mgeplf: I wonder if there is a better choice than `LineCollection` w/ `linewidth`, such that the units are scalled correctly, and we don't have to do any of this.  For instance `Rectange` scales correctly?

I haven't looked at it thoroughly, though.

wizmer:
> let's move to python2 in another PR, I don't wan to take responsability for that haha

Ok, I will prepare the PR.

neurom/view/view.py
arnaudon: Yes, it seems rectangle will work: https://matplotlib.org/devdocs/api/_as_gen/matplotlib.patches.Rectangle.html
I can give it a try if you want, but I'm not sure I'll be able to do that correctly

mgeplf: > Yes, it seems rectangle will work

Work as in correctly scale?  That's what I don't know, and don't have time to test.  If you're think it will, and you're willing to give it a shot, that would be appreciated.

arnaudon: I'll give it a try and update this PR if it works better, I agree it will be cleaner to get rid of this rescale thing, I don't like it either

arnaudon:
So I replaced the hack with linewidth scaling to using Rectangle if realistic_diameter=True. It is a bit slower (0.2s vs. 1.8s on my test), but it gives correct thickness (which is much smaller than what it was before!). There may be a faster way to code it up, but at least this seems to work.

PR #829, opened 2020-05-07
Drop python 2 support

wizmer:


setup.py
```diff
@@ -34,7 +34,7 @@
 from setuptools import find_packages
 
 
-VERSION = "1.4.20"
+VERSION = "1.5.0"
 
 REQS = ['click>=7.0',
         'enum-compat>=0.0.2',
```
haleepfl, line 40: Can you drop this? What about pylru below?

setup.py
```diff
@@ -68,7 +68,7 @@
         'plotly': ['plotly>=3.6.0'],
     },
     'include_package_data': True,
-
+    'python_requires': '>=3.6',
     'classifiers': [
         'Development Status :: 6 - Mature',
         'Intended Audience :: Education',
```
haleepfl, line 74: Remove py2/py27/py35 below

tox.ini
```diff
@@ -7,7 +7,7 @@ testdeps =
```
haleepfl, line 7: Can you remove pathlib2 now?

mgeplf:
Instead of doing this in the v1.* version, why not deprecate python in the v2 branch? Then it's a clean break opportunity.

setup.py
haleepfl: I should have included `future` here too, can that be removed?

jdcourcol:
@alex4200 do we still use python2 in the BSP use cases ?
Usually we install latest package so this would break our use cases if it is still the case.

alex4200:
There are still notebooks using python2, but most use cases should be using python3. I am trying to remove all python2 usecases by the time of the review.

wizmer:
> Usually we install latest package so this would break our use cases if it is still the case.
It should not. The existing NeuroM version with python 2 support will still be available in Pypi. It will be pick up when someone tries to install NeuroM with python 2.

wizmer:
> Instead of doing this in the v1.* version, why not deprecate python in the v2 branch? Then it's a clean break opportunity.

Yes. That's a good point. I'll try to build the Windows MorphIO wheels today or next weeks so that we can finally go forward on this.

setup.py
wizmer: What's the deal with `pylru` ?

haleepfl: There is now a built-in [LRU cache](https://docs.python.org/3/library/functools.html#functools.lru_cache), I assumed `pylru` was only for py2.7 support.

wizmer:
@mgeplf It's ready to be reviewed

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.7"
   - "3.6"
   - "3.7"
```
mgeplf, line 4: add 3.8?

tox.ini
```diff
@@ -3,11 +3,10 @@ name = neurom
 testdeps =
     mock
     nose
-    pathlib2
 
 [tox]
 envlist =
-    {py27,py36}
+    {py36,py37}
```
mgeplf, line 9: do we need to list all of them, or does tox handle this automatically?

wizmer: Yes, there is no way for tox (locally, on Travis it's another story since it is stated in the yaml file) to know under which python to run those tests except that having this list.

.travis.yml
wizmer: I was wondering. Okay, done.

wizmer:
They say we only live once...

PR #831, opened 2020-05-18
Use setuptools for versionning

wizmer:


setup.py
```diff
@@ -81,7 +78,7 @@
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Topic :: Scientific/Engineering :: Bio-Informatics',
-    ]
-}
-
-setup(**config)
+    ],
+    use_scm_version=True,
```
haleepfl, line 82: We should consider whether the default local scheme is what we prefer. I noticed that libsonata uses "no-local-version": https://github.com/BlueBrain/libsonata/blob/master/setup.py#L181

wizmer: Good point. I'd say that using "no-local-version" is probably less error prone.

wizmer:
Looks much nice indeed

PR #840, opened 2020-07-13
load_neuron accepts morphio.Morphology/morphio.mut.Morphology

wizmer:


neurom/io/utils.py
```diff
@@ -115,6 +116,16 @@ def get_files_by_path(path):
 
 def load_neuron(handle, reader=None):
     """Build section trees from an h5 or swc file."""
+    if isinstance(handle, (morphio.Morphology, morphio.mut.Morphology)):
+        fd, temp_file = tempfile.mkstemp(str(uuid.uuid4()) + '.h5', prefix='neurom-')
```
asanin-epfl, line 120: Rather a question. Should we delete this `temp_file` after it's been used? According to [tempfile.mkstemp](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp) the user is responsible for its removal. We also produce temp files in `utils._get_file`. Should we clean them there also?

mgeplf: Seconded, temp files should be cleaned up

neurom/tests/test_stats.py
```diff
@@ -68,7 +70,11 @@ def test_fit_normal_regression():
     nt.assert_almost_equal(fit_.params[0], 10.019332055822, 12)
     nt.assert_almost_equal(fit_.params[1], 0.978726207747, 12)
     nt.assert_almost_equal(fit_.errs[0], 0.021479979161, 12)
-    nt.assert_almost_equal(fit_.errs[1], 0.745431659944, 12)
+
+    if sys.version_info[0] == 2:
```
HDictus, line 74: This failure depends on the version of scipy, but I think here I think you're checking the neurom version? alternative to this is to ensure that the same mode for kstest is used regardless of scipy version (e.g. #839 ).

neurom/io/utils.py
```diff
@@ -115,6 +116,16 @@ def get_files_by_path(path):
...
+        os.close(fd)
+
+        # TODO: make a better morphio loader that does not require a tempfile
+        if isinstance(handle, morphio.Morphology):
+            handle = handle.as_mutable()
```
HDictus, line 125: Probably just my ignorance about morphio, but these two lines seem unnecessary to me

neurom/tests/test_stats.py
wizmer: Since 3 days ago we dropped Python 2 so this is no longer an issue.

neurom/io/utils.py
wizmer: morphio has 2 classes:
- `morphio.Morphology` 
- `morphio.mut.Morphology`

This line is to make sure we end up with a `morphio.mut.Morphology` regardless of the input class type of `handle` (`morphio.Morphology` or `morphio.mut.Morphology`)

setup.py
```diff
@@ -34,24 +34,22 @@
...
+                      'morphio>=2.3.10',
+                      'numpy>=1.8.0',
+                      'pylru>=1.0',
+                      'pyyaml>=3.10',
+                      'scipy>=1.2.0',
+                      'tqdm>=4.8.4',
+                      ],
     packages=find_packages(),
     license='BSD',
     scripts=['apps/raw_data_check',
```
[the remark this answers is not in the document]
wizmer: I don't see it in my IDE. It might have disappear during the rebase. Can you tell me your IDE still warns you for something ?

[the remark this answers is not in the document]
asanin-epfl: it's been fixed after rebasing

asanin-epfl:
@wizmer is it ready for a review? can you assign reviewers?

neurom/tests/test_stats.py
HDictus: I failed this test with python 3.7 after updating scipy, but this now passes CI so i dunno

neurom/io/utils.py
```diff
@@ -113,7 +116,15 @@ def get_files_by_path(path):
 
 def load_neuron(handle, reader=None):
     """Build section trees from an h5 or swc file."""
-    rdw = load_data(handle, reader)
+    if isinstance(handle, (morphio.Morphology, morphio.mut.Morphology)):
+        with tempfile.NamedTemporaryFile(mode='w+b', suffix='.h5') as f:
+            # TODO: make a better morphio loader that does not require a tempfile
```
mgeplf, line 121: perhaps reference the MorphIO issue where this feature is described.

setup.py
```diff
@@ -41,6 +37,7 @@
         'click>=7.0',
         'h5py>=2.7.1',
         'matplotlib>=3.2.1',
+        'morphio>=2.3.10',
```
mgeplf, line 40: I thought the plan was to move to MorphIO once it's ready - if it can be a dependency, does that mean it's ready?

asanin-epfl:
All good from me. I will wait for answers to MikeG questions.

wizmer:
Well, I did this PR for the purpose of putting the sampling in `morph-tool`. At the end this solution did not get chosen.
Mike's point about the MorphIO dependency is also very good. 
So either we dismiss this PR or we remove the MorphIO dependency from the setup.py. 
After all, this PR is about adding the capability to create NeuroM neuron from MorphIO neurons. It means the user must already have MorphIO installed if they want to use the feature.

asanin-epfl:
Implemented in v2

PR #841, opened 2020-07-14
New function extract_dataframe that extract morphometrics as a dataframe

wizmer:
Similar to how `extract_stats` works except that it returns a dataframe instead of a dict.

The dataframe looks like:

```
     name     neurite_type  ...  max_segment_midpoint_Y  max_segment_midpoint_Z
0  Neuron             axon  ...                0.000000               49.520306
1  Neuron  apical_dendrite  ...                0.000000               53.750948
4  simple             axon  ...               -2.000000                0.000000
5  simple  apical_dendrite  ...                     NaN                     NaN
```


wizmer:
@arnaudon All you would need then for your usecase would be to call `extract_stats` with the following dict:

```
neurite:
    total_length_per_neurite:
        - total

neurite_type:
    - AXON
    - APICAL_DENDRITE
    - BASAL_DENDRITE
```

arnaudon:
Lovely!

wizmer:
@mgeplf The context is to have something usable for the repair validation https://bbpcode.epfl.ch/code/#/c/49859/

wizmer:
Python 2 is causing 2 issues in this PR.
- there is a mismatch in the returned value of a numpy function between both python.
- Pandas 1.0 is not available for Python 2

I know we've discussed in the past postponing the python 2 support drop but I think this is causing more harm than good though.

mgeplf:
> there is a mismatch in the returned value of a numpy function between both python.

Can you point to these errors.

> Pandas 1.0 is not available for Python 2

Why is pandas 1 required?

neurom/apps/morph_stats.py
```diff
@@ -73,19 +82,59 @@ def _stat_name(feat_name, stat_mode):
     return '%s_%s' % (stat_mode, feat_name)
 
 
-def extract_stats(neurons, config):
+def _fill_compoundified(data, stat_name, stat):
```
mgeplf, line 85: I think this needs a better name and description; also, does this work for all stats?

wizmer: The diff UI is confusing but I haven't changed this code. It was a nested function that I have simply extracted out so it should work as before. I can change its name and description though.

wizmer:
> Can you point to these errors.

It's the change in `test_stats.py` in the git diff. I had to do an if/else on the python version to make the test pass.

> Why is pandas 1 required?

I'm not sure I understand the question. Because I want to return a dataframe. Or you mean why specifically version 1 ? Because having a dependency spanning over multiple major version is scary.

Moreover after more thinking, I don't think dropping a python version require a major version bump. Numpy did not do it for example. I think dropping a support is actually less drastic than introducing a breaking change. It is just freezing support but not breaking anything. People under python 2 will still be able to do `pip install -U neurom` and nothing will break. They won't have anymore updates, that's all.

neurom/apps/morph_stats.py
```diff
@@ -73,34 +82,77 @@ def _stat_name(feat_name, stat_mode):
     return '%s_%s' % (stat_mode, feat_name)
 
 
-def extract_stats(neurons, config):
+def extract_stats(neurons, config, as_frame=False):
```
wizmer, line 85: I still wonder if it is best to have the `as_frame` to decide which function to return or if I should simply provide a different function ?

arnaudon: In the cli, one would only use as_frame=False, right? This will be mostly for internal use I guess, so I'd be for leaving it as it is. I'm going to update the morph-validator code to use that function now.

neurom/apps/morph_stats.py
```diff
@@ -73,34 +82,77 @@ def _stat_name(feat_name, stat_mode):
...
+    if as_frame:
+        return _extract_stats_as_frame(neurons, config)
+    return _extract_stats_as_dict(neurons, config)
+
+
+extract_stats.__doc__ = extract_stats.__doc__ + EXAMPLE_CONFIG + '\n'
+
+
+def _extract_stats_as_frame(neurons, config):
+    del config['neuron']
```
arnaudon, line 113: why that? in the config you mentionned above, it does not have a 'neuron' entry, so it crashed here

neurom/apps/morph_stats.py
```diff
@@ -73,34 +82,77 @@ def _stat_name(feat_name, stat_mode):
...
         neurite_type = _NEURITE_MAP[neurite_type]
         for mode in modes:
             stat_name = _stat_name(feature_name, mode)
             stat = eval_stats(nm.get(feature_name, neurons, neurite_type=neurite_type), mode)
-            _fill_compoundified(stats[neurite_type.name], stat_name, stat)
+            _fill_stats_dict(stats[neurite_type.name], stat_name, stat)
 
-    for feature_name, modes in config['neuron'].items():
+    for feature_name, modes in config.get('neuron', {}).items():
         for mode in modes:
```
arnaudon, line 150: in the config you mentionned in the discussion, there is a single mode not in a list, so it crashed here as well. Maybe we can add
```
if not isinstance(modes, list):
    modes = [modes]
```

neurom/apps/morph_stats.py
```diff
@@ -73,34 +82,77 @@ def _stat_name(feat_name, stat_mode):
...
+        return _extract_stats_as_frame(neurons, config)
+    return _extract_stats_as_dict(neurons, config)
+
+
+extract_stats.__doc__ = extract_stats.__doc__ + EXAMPLE_CONFIG + '\n'
+
+
+def _extract_stats_as_frame(neurons, config):
+    del config['neuron']
+    stats = {nrn.name: extract_stats(nrn, config) for nrn in neurons}
```
arnaudon, line 114: If I give a single neuron as input, it fails here

arnaudon: also, I use neurom.load_neurons(list_of_paths), and the neuron.name=None, so I only get the last neuron in the dataframe

neurom/apps/morph_stats.py
wizmer: `config['neuron']` contains the config for stats an neuron level. These cannot be splitted by neurite type so they can't be incorporated into the dataframe.
I'm not sure how to handle that.

neurom/apps/morph_stats.py
wizmer: I don't get what is crashing. This seems to work:
```python
    nrns = nm.load_neurons([os.path.join(SWC_PATH, name)
                            for name in ['Neuron.swc', 'simple.swc']])
    config = {'neurite': {'total_length_per_neurite': ['total']},
              'neurite_type': ['AXON', 'APICAL_DENDRITE', 'BASAL_DENDRITE']}
    actual = ms.extract_dataframe(nrns, config)
```
Can you elaborate ?

neurom/apps/morph_stats.py
arnaudon:
```
if 'neuron' in config:
    del config['neuron']
    L.warning(' neuron based config is not handles with to_frame option')
```

neurom/apps/morph_stats.py
arnaudon: I use that config:
```
CONFIG_TOTAL_LENGTH = {
    "neurite": {"total_length_per_neurite": "total"},
    "neurite_type": {"AXON", "APICAL_DENDRITE", "BASAL_DENDRITE"},
}
```

arnaudon: I use that config:
```
CONFIG_TOTAL_LENGTH = {
    "neurite": {"total_length_per_neurite": "total"},
    "neurite_type": {"AXON", "APICAL_DENDRITE", "BASAL_DENDRITE"},
}
```

arnaudon: But this is fine, we can pass a list, but I thought it may be more user friendly to be able to pass a single string

neurom/apps/morph_stats.py
arnaudon: alternatively, you can add another neurite_type = 'neuron' in the dataframe

neurom/apps/morph_stats.py
wizmer: I'm not a huge fan either of this config format but since it is already in use in NeuroM we cannot really change it.
However I have made the "neurite_type" key optional.

neurom/apps/morph_stats.py
wizmer: I did that:
```python
    nrns = nm.load_neurons([os.path.join(SWC_PATH, name)
                            for name in ['Neuron.swc', 'simple.swc']])
    print(ms.extract_dataframe(nrns, REF_CONFIG))
```

and the output is:
```
     name     neurite_type  ...  max_segment_midpoint_Y  max_segment_midpoint_Z
0  Neuron             axon  ...                0.000000               49.520306
1  Neuron  apical_dendrite  ...                0.000000               53.750948
2  Neuron   basal_dendrite  ...               48.481977               51.575581
3  Neuron              all  ...               48.481977               53.750948
4  simple             axon  ...               -2.000000                0.000000
5  simple  apical_dendrite  ...                     NaN                     NaN
6  simple   basal_dendrite  ...                5.000000                0.000000
7  simple              all  ...                5.000000                0.000000
```

Do you know how did you end up with neuron.name = None ?

neurom/apps/morph_stats.py
```diff
@@ -73,34 +83,97 @@ def _stat_name(feat_name, stat_mode):
...
+        An example config can be found at:
+    """
+    if isinstance(neurons, FstNeuron):
+        neurons = [neurons]
+    config = config.copy()
+    if 'neuron' in config:
+        del config['neuron']
+
+    stats = {nrn.name: extract_stats(nrn, config) for nrn in neurons}
+    columns = list(next(iter(next(iter(stats.values())).values())).keys())
```
asanin-epfl, line 112: Lets expand here for readability.

neurom/apps/morph_stats.py
```diff
@@ -73,34 +83,97 @@ def _stat_name(feat_name, stat_mode):
...
+    Returns:
+        The extracted statistics
+
+    Note:
+        An example config can be found at:
+    """
+    if isinstance(neurons, FstNeuron):
+        neurons = [neurons]
+    config = config.copy()
+    if 'neuron' in config:
```
asanin-epfl, line 108: Lets provide an explanation comment on why we need to delete this.

neurom/apps/morph_stats.py
```diff
@@ -73,34 +83,97 @@ def _stat_name(feat_name, stat_mode):
...
+    columns = list(next(iter(next(iter(stats.values())).values())).keys())
+
+    rows = [[name, neurite_type] + list(features.values())
+            for name, data in stats.items()
+            for neurite_type, features in data.items()]
+    return pd.DataFrame(columns=['name', 'neurite_type'] + columns,
+                        data=rows)
+
+
+extract_dataframe.__doc__ = extract_dataframe.__doc__ + EXAMPLE_CONFIG + '\n'
```
asanin-epfl, line 121: Can we use `+=` here? The same for `extract_stats`.

neurom/apps/tests/test_morph_stats.py
```diff
@@ -140,6 +144,63 @@ def test_extract_stats_single_neuron():
...
+    actual = ms.extract_dataframe(nrns, REF_CONFIG)
+    expected = pd.read_csv(os.path.join(DATA_PATH, 'extracted-stats.csv'), index_col=0)
+    assert_frame_equal(actual, expected)
+
+    # Test with a single neuron in the population
+    nrns = nm.load_neurons(os.path.join(SWC_PATH, 'Neuron.swc'))
+    actual = ms.extract_dataframe(nrns, REF_CONFIG)
+    assert_frame_equal(actual, expected[expected.name == 'Neuron'], check_dtype=False)
+
+    # Test with a config with the 'neuron' key
```
asanin-epfl, line 160: > Test with a config with the 'neuron' key

Where is this 'neuron' in config?

neurom/apps/tests/test_morph_stats.py
```diff
@@ -140,6 +144,63 @@ def test_extract_stats_single_neuron():
...
+    nrns = nm.load_neurons([os.path.join(SWC_PATH, name)
+                            for name in ['Neuron.swc', 'simple.swc']])
+    config = {'neurite': {'total_length_per_neurite': ['total']},
+              'neurite_type': ['AXON', 'APICAL_DENDRITE', 'BASAL_DENDRITE']}
+    actual = ms.extract_dataframe(nrns, config)
+    expected = pd.DataFrame(
```
asanin-epfl, line 166: Can we reuse `expected` from before? Something like `expected[['total_total_length_per_neurite']]`?

neurom/tests/test_stats.py
```diff
@@ -31,6 +31,7 @@
 Since the stats module consists of simple wrappers to scipy.stats functions,
 these tests are only sanity checks.
 """
+import sys
```
asanin-epfl, line 34: Why this import?

neurom/apps/morph_stats.py
wizmer: It's a lame line to dig into the first val of a nested dict. I can expand but I doubt it will look better anyway.
Maybe I can simply add a comment...

neurom/apps/tests/test_morph_stats.py
wizmer: The comment was supposed to say "without"

neurom/apps/morph_stats.py
```diff
@@ -73,34 +83,94 @@ def _stat_name(feat_name, stat_mode):
...
+            - neurite: a dictionary {neurite_feature: mode} where:
+                - neurite_feature is a string from NEURITEFEATURES
+                - mode is an aggregation operation provided as a string such as:
+                  ['min', 'max', 'median', 'mean', 'std', 'raw', 'total']
+
+    Returns:
+        The extracted statistics
+
+    Note:
+        An example config can be found at:
```
asanin-epfl, line 103: Just had an idea. What if we use placeholder here for this config example? Because currently it does not look obvious that we adjust this docstring later. With placeholder it would be clearer.

`An example config can be found at: {config_path}`

Later we would use `extract_stats.__doc__ = extract_stats.__doc__.format({'config_path': EXAMPLE_CONFIG})`

arnaudon:
It would be nice to have some basic multiprocessing to extract features ;)

PR #842, opened 2020-07-15
Drop Python 2.7 support

mgeplf:
* Make it Python 3.5+ only


mgeplf:
I think we should create another issue to actually remove all the compat stuff, so you guys can move forward, what do you think?

wizmer:
I'm actually currently rebasing https://github.com/BlueBrain/NeuroM/pull/829

mgeplf:
cool, I'll drop this

PR #843, opened 2020-07-16
Realistic diameters

arnaudon:
Continuation of https://github.com/BlueBrain/NeuroM/pull/828


neurom/view/view.py
```diff
@@ -276,6 +297,9 @@ def plot_neuron3d(ax, nrn, neurite_type=NeuriteType.all,
         color(str or None): Color of plotted values, None corresponds to default choice
         alpha(float): Transparency of plotted values
     """
+    if realistic_diameters:
+        print("Warning: realistic_diameters option not implemented for 3d plots.")
```
wizmer, line 301:
```python
from warnings import warn

warn("Option 'realistic_diameters' not implemented for 3d plots.)
```
I will merge the CLI unit test PR and then I think we are good

wizmer:
Maybe at some point we change the default so that realistic diameters are used by default.

PR #844, opened 2020-07-16
Add unit tests for CLI 'neurom view'

wizmer:
And move the cli file within the package so that it can be tested

### DIFF
--- a/neurom/apps/tests/test_cli.py
+++ b/neurom/apps/tests/test_cli.py
@@ -1,0 +1,36 @@
+from pathlib import Path
+
+from click.testing import CliRunner
+from mock import MagicMock, patch
+from nose.tools import assert_equal
+
+from neurom.apps.cli import cli
+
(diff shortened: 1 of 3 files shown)


neurom/apps/tests/test_cli.py
```diff
@@ -0,0 +1,36 @@
...
+from click.testing import CliRunner
+from mock import MagicMock, patch
+from nose.tools import assert_equal
+
+from neurom.apps.cli import cli
+
+DATA = Path(__file__).parent.parent.parent.parent.resolve() / 'test_data'
+
+
+@patch('neurom.apps.cli.plt.show', MagicMock())
```
asanin-epfl, line 12: Can we verify that the mocked method was called once? The same applies to another test.

wizmer: Good point

wizmer:
Thanks @asanin-epfl !

PR #845, opened 2020-07-16
Fortify load_neurons to handle pathlib.Path and sequences

wizmer:
And remove remnants of python 2 compat code


neurom/io/tests/test_utils.py
```diff
@@ -92,21 +93,53 @@ def _check_neurites_have_no_parent(nrn):
...
+def test_get_morph_files():
+    ref = set(['Neuron_h5v2.h5', 'Neuron_2_branch_h5v2.h5', 'Neuron_slice.h5',
+               'Neuron.swc', 'Neuron_h5v1.h5', 'Neuron_2_branch_h5v1.h5'])
+
+    FILE_PATH = os.path.abspath(os.path.join(DATA_PATH, 'valid_set'))
+    files = set(os.path.basename(f) for f in utils.get_morph_files(FILE_PATH))
+
+    nt.assert_equal(ref, files)
+
+
```
wizmer, line 105: I have simply moved this function above test_load_neurons so that it is tested first. It is not a new code.

neurom/io/tests/test_utils.py
```diff
@@ -92,21 +93,53 @@ def _check_neurites_have_no_parent(nrn):
...
 
-def test_get_morph_files():
-    ref = set(['Neuron_h5v2.h5', 'Neuron_2_branch_h5v2.h5', 'Neuron_slice.h5',
-               'Neuron.swc', 'Neuron_h5v1.h5', 'Neuron_2_branch_h5v1.h5'])
+    # Single Path
+    nrns = utils.load_neurons(Path(FILES[0]), neuron_loader=_mock_load_neuron)
+    nt.assert_equal(nrns[0].name, _get_name(FILES[0]))
 
-    FILE_PATH = os.path.abspath(os.path.join(DATA_PATH, 'valid_set'))
-    files = set(os.path.basename(f) for f in utils.get_morph_files(FILE_PATH))
```
wizmer, line 107: GH is messing with the UI, the red is the function I moved. The green is the new tests I have added

neurom/io/tests/test_utils.py
```diff
@@ -92,21 +93,53 @@ def _check_neurites_have_no_parent(nrn):
...
+    # sequence of strings
+    iter_file = (f for f in FILES)
+    nrns = utils.load_neurons(iter_file, neuron_loader=_mock_load_neuron)
+    for i, nrn in enumerate(nrns):
+        nt.assert_equal(nrn.name, _get_name(FILES[i]))
+    nt.assert_raises(SomaError, utils.load_neurons, NO_SOMA_FILE)
```
asanin-epfl, line 126: Why do we keep repeating this line below? Should not be one time enough?

neurom/io/tests/test_utils.py
asanin-epfl: Thank you! It saved me time.

neurom/io/tests/test_utils.py
asanin-epfl: ok

neurom/io/tests/test_utils.py
wizmer: Because I'm dumb XD

wizmer:
>  Where do you plan to do a mass replace of os.path with Path

Yes, I should do this one day. But I did not have time yet.

> Do we still need OrderedEnum implementation in utils.py?

I think we do. The code provided at https://docs.python.org/3/library/enum.html#orderedenum is just a recipe. So I think we need to implement it even in python 3. Correct me if I'm wrong.

arnaudon:
The bug is at l.117 of io/utils.py, you need to add Path along with str ;)

wizmer:
Ahh, you mean the name = None bug. Okay, I'll hotfix that. Thanks for finding out.

arnaudon:
Yes! I just converted to Path -> str to make it work on morph-validator, but maybe good to have this consistent in neurom loaders.

PR #846, opened 2020-07-21
Widespread the use of pathlib.Path

wizmer:
- neurom.io.utils `get_files_by_path` and `get_morph_files` returns pathlib objects
- fix name == None when calling load_neuron with a pathlib object
- all unit tests use pathlib instead of os.path

### DIFF
--- a/neurom/check/runner.py
+++ b/neurom/check/runner.py
@@ -58,7 +58,7 @@ class CheckRunner(object):
 
         for _f in utils.get_files_by_path(path):
             L.info(SEPARATOR)
-            status, summ = self._check_file(_f)
+            status, summ = self._check_file(str(_f))
             res &= status
             if summ is not None:
                 summary.update(summ)
--- a/neurom/check/runner.py
+++ b/neurom/check/runner.py
@@ -58,7 +58,7 @@ class CheckRunner(object):
 
         for _f in utils.get_files_by_path(path):
             L.info(SEPARATOR)
-            status, summ = self._check_file(str(_f))
+            status, summ = self._check_file(_f)
             res &= status
             if summ is not None:
                 summary.update(summ)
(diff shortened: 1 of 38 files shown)


neurom/check/runner.py
```diff
@@ -58,7 +58,7 @@ def run(self, path):
 
         for _f in utils.get_files_by_path(path):
             L.info(SEPARATOR)
-            status, summ = self._check_file(_f)
+            status, summ = self._check_file(str(_f))
```
mgeplf, line 61: _check_file should take a Path?

neurom/check/tests/test_runner.py
```diff
@@ -76,7 +77,7 @@
 
 REF_0 = {
     'files': {
-        NRN_PATH_0: {
+        str(NRN_PATH_0): {
```
mgeplf, line 80: not sure it's worth changing to Path here

neurom/io/utils.py
```diff
@@ -94,30 +94,31 @@ def get_morph_files(directory):
     Returns:
         list with all files with extensions '.swc' , 'h5' or '.asc' (case insensitive)
     """
-    lsdir = (os.path.join(directory, m) for m in os.listdir(directory))
-    return list(filter(_is_morphology_file, lsdir))
+    directory = Path(directory)
+    return [Path(directory, m) for m in directory.iterdir() if _is_morphology_file(m)]
```
mgeplf, line 98: `iterdir` doesn't return Path items?

neurom/check/runner.py
wizmer: Ack

asanin-epfl:
You are still using `os.path` in tests. What's the plan on them?

asanin-epfl:
Here are the occurrences:
Screen Shot 2020-07-21 at 2 05 20 PM

wizmer:
> You are still using `os.path` in tests. What's the plan on them?

Ah yes, my find and replaced froze and did not finish the job at the first try. I forgot.
That should be alright now.

wizmer:
Thank you !

PR #847, opened 2020-07-22
Use multiprocessing in morphs_stats

arnaudon:
Small refactoring of morphs_stats module for computational speedup, with the following changes:

1. uses pool.imap_unordered in extract_dataframe in the loop overs neurons
2. allow giving a list of str/Path instead of neuron collection (so that the expansive loadine is done in parallel, see 1.)
3. moved neurom.get outside mode loop to avoid duplicate computations

### DIFF
--- a/neurom/apps/morph_stats.py
+++ b/neurom/apps/morph_stats.py
@@ -27,12 +27,14 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 """Core code for morph_stats application."""
+import os
 import logging
 from collections import defaultdict
 from itertools import product
 from pathlib import Path
 import multiprocessing
 from functools import partial
+import warnings
 
 import numpy as np
 import pandas as pd
@@ -126,6 +128,8 @@ def extract_dataframe(neurons, config, n_workers=1):
     if n_workers == 1:
         stats = dict(map(func, neurons))
     else:
+        if n_workers > os.cpu_count():
+            warnings.warn(f'n_workers ({n_workers}) > os.cpu_count() ({os.cpu_count()}))')
         with multiprocessing.Pool(n_workers) as pool:
             stats = dict(pool.imap_unordered(func, neurons))
 
@@ -159,9 +163,6 @@ def extract_stats(neurons, config):
 
     {config_path}
     """
-    if isinstance(neurons, (str, Path)) or (isinstance(neurons, list) and
-                                            all(isinstance(nrn, (str, Path)) for nrn in neurons)):
-        neurons = nm.load_neurons(neurons)
 
     def _fill_stats_dict(data, stat_name, stat):
         """Insert the stat data in the dict.
--- a/neurom/apps/morph_stats.py
+++ b/neurom/apps/morph_stats.py
@@ -31,6 +31,9 @@ import logging
 from collections import defaultdict
 from itertools import product
 from pathlib import Path
+import multiprocessing
+from tqdm import tqdm
+from functools import partial
 
 import numpy as np
 import pandas as pd
@@ -82,11 +85,11 @@ def _stat_name(feat_name, stat_mode):
     return '%s_%s' % (stat_mode, feat_name)
 
 
-def extract_dataframe(neurons, config):
+def extract_dataframe(neurons, config, n_workers=1):
     """Extract stats grouped by neurite type from neurons.
 
     Arguments:
-        neurons: a neuron, population or neurite tree
+        neurons: a neuron, population, neurite tree or list of neuron paths
         config (dict): configuration dict. The keys are:
             - neurite_type: a list of neurite types for which features are extracted
               If not provided, all neurite_type will be used
@@ -94,7 +97,7 @@ def extract_dataframe(neurons, config):
                 - neurite_feature is a string from NEURITEFEATURES
                 - mode is an aggregation operation provided as a string such as:
                   ['min', 'max', 'median', 'mean', 'std', 'raw', 'total']
-
+        n_workers (int): number of workers for multiprocessing (on collection of neurons)
     Returns:
         The extracted statistics
 
@@ -112,7 +115,14 @@ def extract_dataframe(neurons, config):
     if 'neuron' in config:
         del config['neuron']
 
-    stats = {nrn.name: extract_stats(nrn, config) for nrn in neurons}
+    with multiprocessing.Pool(n_workers) as pool:
+        stats = {nrn if isinstance(nrn, str) else nrn.name:
+                 stat for nrn, stat in tqdm(zip(neurons,
+                                                pool.imap(partial(extract_stats, config=config),
+                                                          neurons)
+                                                ), total=len(neurons)
+                                            )
+                 }
     columns = list(next(iter(next(iter(stats.values())).values())).keys())
 
     rows = [[name, neurite_type] + list(features.values())
@@ -126,7 +136,7 @@ def extract_stats(neurons, config):
     """Extract stats from neurons.
 
     Arguments:
-        neurons: a neuron, population or neurite tree
+        neurons: a neuron, population, neurite tree or list of neuron paths/str
         config (dict): configuration dict. The keys are:
             - neurite_type: a list of neurite types for which features are extracted
               If not provided, all neurite_type will be used
@@ -143,6 +153,10 @@ def extract_stats(neurons, config):
 
     {config_path}
     """
+    from neurom import load_neurons
+    if isinstance(neurons, (str, Path)) or (isinstance(neurons, list) and
+                                            all(isinstance(nrn, (str, Path)) for nrn in neurons)):
+        neurons = load_neurons(neurons)
 
     def _fill_stats_dict(data, stat_name, stat):
         """Insert the stat data in the dict.
@@ -162,15 +176,17 @@ def extract_stats(neurons, config):
                                                        config.get('neurite_type',
                                                                   _NEURITE_MAP.keys())):
         neurite_type = _NEURITE_MAP[neurite_type]
+        feature = nm.get(feature_name, neurons, neurite_type=neurite_type)
         for mode in modes:
             stat_name = _stat_name(feature_name, mode)
-            stat = eval_stats(nm.get(feature_name, neurons, neurite_type=neurite_type), mode)
+            stat = eval_stats(feature, mode)
             _fill_stats_dict(stats[neurite_type.name], stat_name, stat)
 
     for feature_name, modes in config.get('neuron', {}).items():
+        feature = nm.get(feature_name, neurons)
         for mode in modes:
             stat_name = _stat_name(feature_name, mode)
-            stat = eval_stats(nm.get(feature_name, neurons), mode)
+            stat = eval_stats(feature, mode)
             _fill_stats_dict(stats, stat_name, stat)
 
     return dict(stats)
--- a/neurom/apps/morph_stats.py
+++ b/neurom/apps/morph_stats.py
@@ -30,7 +30,6 @@
 import logging
 from collections import defaultdict
 from itertools import product
-from functools import partial
 from pathlib import Path
 import multiprocessing
 from tqdm import tqdm
@@ -122,13 +121,13 @@ def extract_dataframe(neurons, config, n_workers=1):
         pool = multiprocessing.Pool(n_workers)
         mapper = pool.imap
 
-    stats = {nrn if isinstance(nrn, str) else nrn.name:
-             stat for nrn, stat in tqdm(zip(neurons,
-                                            mapper(partial(extract_stats, config=config), neurons)
-                                            ),
-                                        total=len(neurons)
-                                        )
-             }
+    def _compute(nrn):
+        if not isinstance(nrn, FstNeuron):
+            nrn = nm.load_neuron(nrn)
+        return nrn.name, extract_stats(nrn, config=config)
+
+    stats = dict(tqdm(mapper(_compute, neurons), total=len(neurons)))
+
     if n_workers > 1:
         pool.close()
 
--- a/neurom/apps/morph_stats.py
+++ b/neurom/apps/morph_stats.py
@@ -116,14 +116,22 @@ def extract_dataframe(neurons, config, n_workers=1):
     if 'neuron' in config:
         del config['neuron']
 
-    with multiprocessing.Pool(n_workers) as pool:
-        stats = {nrn if isinstance(nrn, str) else nrn.name:
-                 stat for nrn, stat in tqdm(zip(neurons,
-                                                pool.imap(partial(extract_stats, config=config),
-                                                          neurons)
-                                                ), total=len(neurons)
-                                            )
-                 }
+    if n_workers == 1:
+        mapper = map
+    else:
+        pool = multiprocessing.Pool(n_workers)
+        mapper = pool.imap
+
+    stats = {nrn if isinstance(nrn, str) else nrn.name:
+             stat for nrn, stat in tqdm(zip(neurons,
+                                            mapper(partial(extract_stats, config=config), neurons)
+                                            ),
+                                        total=len(neurons)
+                                        )
+             }
+    if n_workers > 1:
+        pool.close()
+
     columns = list(next(iter(next(iter(stats.values())).values())).keys())
 
     rows = [[name, neurite_type] + list(features.values())
--- a/neurom/apps/morph_stats.py
+++ b/neurom/apps/morph_stats.py
@@ -32,7 +32,7 @@ from collections import defaultdict
 from itertools import product
 from pathlib import Path
 import multiprocessing
-from tqdm import tqdm
+from functools import partial
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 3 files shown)


arnaudon:
I don't know why I got this high number of previous commit here, but the changes are only the latest commits, sorry about that!

neurom/apps/morph_stats.py
```diff
@@ -112,7 +116,14 @@ def extract_dataframe(neurons, config):
     if 'neuron' in config:
         del config['neuron']
 
-    stats = {nrn.name: extract_stats(nrn, config) for nrn in neurons}
+    with multiprocessing.Pool(n_workers) as pool:
```
wizmer, line 119: Can you make a `if/else` and only use Pool when `n_workers > 1`

neurom/apps/morph_stats.py
```diff
@@ -143,6 +154,9 @@ def extract_stats(neurons, config):
 
     {config_path}
     """
+    if isinstance(neurons, (str, Path)) or (isinstance(neurons, list) and
+                                            all(isinstance(nrn, (str, Path)) for nrn in neurons)):
+        neurons = nm.load_neurons(neurons)
```
wizmer, line 159: I'm going to see if I can make `load_neurons` lazy instead of doing this.

arnaudon: this is such a nice line of code!

neurom/apps/morph_stats.py
arnaudon: will do!

neurom/apps/morph_stats.py
```diff
@@ -112,7 +116,14 @@ def extract_dataframe(neurons, config):
...
         del config['neuron']
 
-    stats = {nrn.name: extract_stats(nrn, config) for nrn in neurons}
+    with multiprocessing.Pool(n_workers) as pool:
+        stats = {nrn if isinstance(nrn, str) else nrn.name:
+                 stat for nrn, stat in tqdm(zip(neurons,
+                                                pool.imap(partial(extract_stats, config=config),
+                                                          neurons)
+                                                ), total=len(neurons)
+                                            )
```
wizmer, line 125: Could you use dask instead. It should be something like:
```python
dask_bag.from_sequence(neurons, npartitions=n_workers).map(extract_stats, config=config).compute()
```

neurom/apps/tests/test_morph_stats.py
```diff
@@ -143,6 +143,20 @@ def test_extract_stats_single_neuron():
             assert_almost_equal(res[k][kk], REF_OUT[k][kk], places=3)
 
 
+def test_extract_stats_single_neuron_path():
```
wizmer, line 146: Reuse the previous test and add a for loop top level on the argument passed to `extract_stats`
```python
path = SWC_PATH / 'Neuron.swc'
for nrn in (nm.load_neuron(path), path, str(path)):
```

neurom/apps/morph_stats.py
mgeplf: Is that extra dependency needed?

wizmer: Alexis is trying and saying it's not working well. Let's stay on Pool.

arnaudon: no I'm going to stick to multiprocess, it does the job for now

neurom/apps/morph_stats.py
arnaudon: do you want to move this line to the load_neurons? So that if a collection of neuron is passed, it will not load it?

neurom/apps/morph_stats.py
wizmer, line 125: Instead of changing `extract_stats` signature, you could do this

```suggestion
        def _compute(nrn):
            if not isinstance(nrn, Neuron):
                nrn = load_neuron(nrn)
            return nrn.name, extract_stats(nrn, config=config)
        stats = dict(tqdm(pool.imap(_compute, neurons)), total=len(neurons))
```
This will also fix the fact that passing a string in your case would return an incorrect neuron name since it would use the whole string instead of just the last part of the path.

neurom/apps/morph_stats.py
arnaudon: this is a pain, I'll leave it like that

neurom/apps/morph_stats.py
mgeplf: > Alexis is trying and saying it's not working well.

Weird, what kind of problem is there?  Is it a dask related thing?

arnaudon: aaah I did the stupid thing of copy pasting and you messed up the brakets, so it was added a key total in the dict haha

arnaudon: For my setting, it needs to have OPENBLAS_NUM_THREADS=1, and it is much slower. I still have to undersand how this dask works, but it is not so easy to have it working out of the box so for here, I guess standard multiprocess is good enough. The main slow part was loading neurons I think.

arnaudon:
I let you fix coverage, I'm tired of doing that... sorry

neurom/apps/morph_stats.py
mgeplf: Thanks for the info - appreciated.

wizmer:
Ok, I'll take over from there. Thanks for the PR !

arnaudon:
Thanks @wizmer !

neurom/apps/morph_stats.py
```diff
@@ -112,7 +122,13 @@ def extract_dataframe(neurons, config):
     if 'neuron' in config:
         del config['neuron']
 
-    stats = {nrn.name: extract_stats(nrn, config) for nrn in neurons}
+    func = partial(_run_extract_stats, config=config)
+    if n_workers == 1:
```
tomdele, line 126: I would have set a limit on the number of worker

tomdele: and adding a warning for the user saying : 
nworker too big setting nworker to X

arnaudon: with max_worker = os.cpu_count() ?

neurom/apps/morph_stats.py
```diff
@@ -112,7 +122,13 @@ def extract_dataframe(neurons, config):
...
+        stats = dict(map(func, neurons))
+    else:
+        with multiprocessing.Pool(n_workers) as pool:
+            stats = dict(pool.imap_unordered(func, neurons))
```
tomdele, line 130: you never use the key ordering from the stats to do anything ?

neurom/apps/morph_stats.py
tomdele: could be yes

neurom/apps/morph_stats.py
arnaudon: see line 132, it is converted to 'name' in line 134

arnaudon: so in the final dataframe we have the name of the cells for which the features have been extracted

neurom/apps/tests/test_morph_stats.py
```diff
@@ -130,17 +130,18 @@ def test_eval_stats_applies_numpy_function():
...
+        assert_almost_equal(res['mean_soma_radius'], REF_OUT['mean_soma_radius'])
 
-    for k in ('all', 'axon', 'basal_dendrite', 'apical_dendrite'):
-        assert_equal(set(res[k].keys()), set(REF_OUT[k].keys()))
-        for kk in res[k].keys():
-            assert_almost_equal(res[k][kk], REF_OUT[k][kk], places=3)
+        for k in ('all', 'axon', 'basal_dendrite', 'apical_dendrite'):
+            assert_equal(set(res[k].keys()), set(REF_OUT[k].keys()))
+            for kk in res[k].keys():
+                assert_almost_equal(res[k][kk], REF_OUT[k][kk], places=3)
```
tomdele, line 145: you don t test : 
```
(isinstance(neurons, list) and all(isinstance(nrn, (str, Path)) for nrn in neurons))
```
even if the coverage is ok.

tomdele: (or maybe I missed it)

arnaudon: line 134 of the test ;)

neurom/apps/morph_stats.py
tomdele: Then I would have sorted the 
```
pd.DataFrame(columns=['name', 'neurite_type'] + columns, data=rows).sort(axis=1)
```

neurom/apps/morph_stats.py
```diff
@@ -112,7 +122,13 @@ def extract_dataframe(neurons, config):
...
+
     columns = list(next(iter(next(iter(stats.values())).values())).keys())
```
tomdele, line 132: 0_o

neurom/apps/tests/test_morph_stats.py
tomdele: Actually my initial though was : 
I would expect extract_stats to raise if `neurons` does not have the correct format. 
and you don t test a failing list

neurom/apps/morph_stats.py
tomdele: this avoids a tonne of problems when you will try to compare stuffs

arnaudon: sure, or just use .imap()

neurom/apps/morph_stats.py
arnaudon: (0(o_o(ooo)a_a)A)

neurom/apps/morph_stats.py
tomdele: > sure, or just use .imap() 
hum I don t think so. iirc, imap is blocking.

arnaudon: yes, we don't care about blocking here

tomdele: ok up to you then

tomdele: I let you check if you really have the same results also with the single worker for the same input. 
this deserve a test.

arnaudon: but we don't care about ordering with pandas, right? I avoid using .iloc unless really necessary.

tomdele: This is not for querying but for the comparisons :
```
>> import pandas as pd 
>> a = {"a": [1,2], "b" : [1,2]}   
>> b = {"b": [1,2], "a" : [1,2]}   
>> pd.DataFrame(a) == pd.DataFrame(b)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-8-311c305e4e29> in <module>
----> 1 pd.DataFrame(a) == pd.DataFrame(b)

~/.virtualenvs/snap/lib/python3.6/site-packages/pandas/core/ops/__init__.py in f(self, other)
    837             if not self._indexed_same(other):
    838                 raise ValueError(
--> 839                     "Can only compare identically-labeled DataFrame objects"
    840                 )
    841             new_data = dispatch_to_series(self, other, op, str_rep)

ValueError: Can only compare identically-labeled DataFrame objects
```
```
import pandas.testing as npt
npt.assert_frame_equal(pd.DataFrame(a), pd.DataFrame(b))                                                                                                                                           
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-12-cff663a5701a> in <module>
----> 1 npt.assert_frame_equal(pd.DataFrame(a), pd.DataFrame(b))

~/.virtualenvs/snap/lib/python3.6/site-packages/pandas/_testing.py in assert_frame_equal(left, right, check_dtype, check_index_type, check_column_type, check_frame_type, check_less_precise, check_names, by_blocks, check_exact, check_datetimelike_compat, check_categorical, check_like, obj)
   1350         check_exact=check_exact,
   1351         check_categorical=check_categorical,
-> 1352         obj=f"{obj}.columns",
   1353     )
   1354 

~/.virtualenvs/snap/lib/python3.6/site-packages/pandas/_testing.py in assert_index_equal(left, right, exact, check_names, check_less_precise, check_exact, check_categorical, obj)
    684             obj=obj,
    685             lobj=left,
--> 686             robj=right,
    687         )
    688 

pandas/_libs/testing.pyx in pandas._libs.testing.assert_almost_equal()

pandas/_libs/testing.pyx in pandas._libs.testing.assert_almost_equal()

~/.virtualenvs/snap/lib/python3.6/site-packages/pandas/_testing.py in raise_assert_detail(obj, message, left, right, diff)
    913         msg += f"\n[diff]: {diff}"
    914 
--> 915     raise AssertionError(msg)
    916 
    917 

AssertionError: DataFrame.columns are different

```

arnaudon: mmh here you compare on columns ordering, not index, right? The sorting you want to do on 'name' is on indices of dataframe. But yes, it's cleaner to ensure the format of df is the same, we should add a .sort() at the end. (I never had to compare dataframes, but maybe one day, who knows!)

tomdele: You compare on both.

So the checks are : 
a.columns == b.columns which includes the ordering
a.index == b.index which includes the ordering
a.data == a.data

why ordering is important ? 
- because if you have (one day) multi level indexers (https://pandas.pydata.org/pandas-docs/stable/user_guide/advanced.html) the algorithm (and so the performances) are different with sorted or unsorted index.
- if you want to use `iloc` the results would differs


Be careful, you need to sort on both axis ! `.sort()` only sorts on indices and not columns.

neurom/apps/morph_stats.py
```diff
@@ -112,7 +122,13 @@ def extract_dataframe(neurons, config):
...
 
     rows = [[name, neurite_type] + list(features.values())
```
tomdele, line 134: by the way, I benchmarked this one day and using numpy arrays if you have the same types for all data is MUCH MUCH MUCH more efficient for this constructor : 
```
pd.DataFrame(columns=['name', 'neurite_type'] + columns, data=rows)
```
and you benefit a lot in doing the conversion beforehand. 
 ```
pd.DataFrame(columns=['name', 'neurite_type'] + columns, data=np.asarray(rows))
```

tomdele: see : 
https://github.com/BlueBrain/libsonata/pull/99#issuecomment-656694175

neurom/apps/morph_stats.py
wizmer: I don't think it's NeuroM responsability to restrain the number of workers. In my mind, it's really up to the user to pick the correct number of workers.

neurom/apps/morph_stats.py
wizmer: I did not think order was important that's why I transformed `imap` to `imap_unordered`. To me, `.sort()` only solves half of the issue.  Adding a `.sort()` would ensure stability, but it would not even ensure the result is the same as the one with running the code with `n_workers==1`. And it would not even be the order of the `neurons` list.
It think it's either:
- `imap_unordered` if we don't care about the ordering. I would say, we don't. morph_stat is about aggregating data, it will most of the time be grouped by mtype afterward, so the order will disappear anyway.
- switch back to `imap` if we do. 
- add a `sort` flag to switch between `imap` and `imap_unordered`

neurom/apps/morph_stats.py
wizmer: Aleksei had more or less the same reaction.

neurom/apps/morph_stats.py
tomdele: meahhhh.
This is a question of risks/benefits. 
To me there is a small benefits in having a number of worker > os.cpu_count() (except maybe in very specific cluster related analyses where the number of available processors can be different than os.cpu_count() and or in I/O bound analaysis) and but there are some risks to slowdown a normal analysis. 

This is up to you !

wizmer: I will cut the pear in 2 then and set a warning but not restrict.

neurom/apps/morph_stats.py
arnaudon: +1 for leave it as it is

neurom/apps/morph_stats.py
arnaudon: me and @wizmer are the only users of that, and we know our stuff!

neurom/apps/morph_stats.py
tomdele: >  Adding a .sort() would ensure stability, but it would not even ensure the result is the same as the one with running the code with n_workers==1. 

Should be or I am missing something ... And if the ordering makes sense is something else. 

But up to you again. I don t know the final goal of this code so maybe this is just pure pain for nothing indeed. I just share my experience with dataframes.

neurom/apps/morph_stats.py
tomdele: I am fine with that

neurom/apps/morph_stats.py
tomdele: @asanin-epfl You are a wise man.

neurom/apps/tests/test_morph_stats.py
wizmer: I have removed this change. I think they are no longer relevant.

neurom/apps/morph_stats.py
wizmer: I prefer to stick to the idiomatic way of doing it if you don't mind...

tomdele: I don t mind

PR #849, opened 2020-09-03
Add maximum radial distance feature

adrien-berchet:
Add a neurite feature to get the maximum radial distance from the soma of a collection of neurites.

### DIFF
--- a/neurom/features/__init__.py
+++ b/neurom/features/__init__.py
@@ -89,12 +89,10 @@ def _get_feature_value_and_func(feature_name, obj, **kwargs):
     feat = _find_feature_func(feature_name)
 
     res = feat(obj, **kwargs)
-    if len(feat.shape) == 0:
-        res_list = [res]
-    else:
-        res_list = list(res)
+    if len(feat.shape) != 0:
+        res = np.array(list(res))
 
-    return np.array(res_list), feat
+    return res, feat
 
 
 def get(feature_name, obj, **kwargs):
--- a/neurom/features/__init__.py
+++ b/neurom/features/__init__.py
@@ -88,7 +88,13 @@ def _get_feature_value_and_func(feature_name, obj, **kwargs):
     """
     feat = _find_feature_func(feature_name)
 
-    return np.array(list(feat(obj, **kwargs))), feat
+    res = feat(obj, **kwargs)
+    if len(feat.shape) == 0:
+        res_list = [res]
+    else:
+        res_list = list(res)
+
+    return np.array(res_list), feat
 
 
 def get(feature_name, obj, **kwargs):
--- a/neurom/features/neuritefunc.py
+++ b/neurom/features/neuritefunc.py
@@ -65,6 +65,12 @@ def total_length(nrn_pop, neurite_type=NeuriteType.all):
 
 
 @feature(shape=())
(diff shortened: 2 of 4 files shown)


arnaudon:
@adrien-berchet , in the end, we don't need that, right? I guess we can close it?

adrien-berchet:
Indeed we don't use it for now. But I am wondering if it is better to close or to merge (after rebase)? Someone might need it someday.

arnaudon:
ah yes, we could merge, if @wizmer has nothing against it!

wizmer:
We can merge it but you'd need to pass it as a PR (no longer a draft) and it needs a rebase

adrien-berchet:
@wizmer I had to update `neurom.features._get_feature_value_and_func` to make it work with your new registering system but I am not sure how you intended to deal with features with `shape=()`, can you check it plz?

arnaudon:
So `_get_feature_value_and_func` was not working previously with `shape=()` types of features?

adrien-berchet:
Exactly, because `neurom/features/__init__.py::91` is trying to build a list from a scalar instead of an iterable.

arnaudon:
So is this a real bug in neurom? The test were still working?

adrien-berchet:
The tests zere zorking because the features with `shape=()` were actually not registered because they were filtered here: `neurom/features/__init__.py::177`. But I don't know the reason of this filter.

arnaudon:
@asanin-epfl , could you merge that? Appart from adding a new feature, it fixes a small bug, see adrien last comment. Thank you!

arnaudon:
@adrien-berchet , can you rebase so we can merge to master?

asanin-epfl:
I will review after the rebase.

adrien-berchet:
> I will review after the rebase.

Done

neurom/features/__init__.py
```diff
@@ -88,7 +88,13 @@ def _get_feature_value_and_func(feature_name, obj, **kwargs):
     """
     feat = _find_feature_func(feature_name)
 
-    return np.array(list(feat(obj, **kwargs))), feat
+    res = feat(obj, **kwargs)
```
asanin-epfl, line 91: Why this change? Tests still pass with the old code.

adrien-berchet: See my two previous comments:
https://github.com/BlueBrain/NeuroM/pull/849#issuecomment-787809109
https://github.com/BlueBrain/NeuroM/pull/849#issuecomment-787846393

And there is a TODO at ``neurom/features/__init__.py::176`` so I guess this change was more or less planned.

adrien-berchet:
I added a more usual feature so it makes it usable and points out the issue before these changes: it is not possible to use ``get_feature()`` for functions with ``shape=()``. So it is not possible to call it to just get a scalar, we always get a list of at least 1 element (I updated the behavior in the last commit).

asanin-epfl (approved):

asanin-epfl:
Ok, good to go then. Thank you!

adrien-berchet:
Thanks!

arnaudon:
Sorry, I'm late to the party: this is fixed here: https://github.com/BlueBrain/NeuroM/pull/884

PR #852, opened 2020-09-17
Deprecate COLS.TYPE, COLS.ID and COLS.P

wizmer:
They represent internal states and should not be public.

A section type is supposed to be accessed with section.type
and ID and parent ID are just some details of the datastructure we use
internally.


asanin-epfl:
They have been dropped in v2

PR #861, opened 2021-01-07
Fix various morph_stats issues

### LINKED ISSUE #859: morph_stats `extract_dataframe` might have a column mismatch
There is a heuristic in morph_stats that split features that return a 3D array into 3 one dimentional features.
For example the feature `segment_midpoint` is splitted into `segment_midpoint_X`, `segment_midpoint_Y` and `segment_midpoint_Z`.

`extract_dataframe` currently returns one row for each neurite_type. The problem arises when the feature does not exist for a given neurite_type. In this case the heuristic `if stat is None or not isinstance(stat, np.ndarray) or stat.shape not in ((3, ), ):` cannot apply because `stat` is None. That means the feature won't be split and the dataframe will be missing some columns.
This can potentially offset columns of the final dataset.

wizmer:
# Fixes handling of multi-dimensional features in `morph_stats`

Context:

Previously, 3 dimensional features where split into 3 components X, Y and Z.
However this way of handling multidim features have issues:

- It only handles 3-dim features and not other dimensionality
- 3-dim features may not be spatial so X, Y and Z might not be appropriate.

Solution:

Use an integer suffix of arbitrary length (ie. `_0`, `_1`, ..., `_N`) instead of `_X`, `_Y` and `_Z`.

:information_source: This fixes #859 

# Fixed issue with computing the mean, std and median of empty features

When features are returning an empty array, the mean, median and std of those
values will be set to None instead of crashing.

# Fixed issue with the convex hull computation in feature `neurite_volume_density`

The neurite volume computation made using a convex hull is now wrapped in a
try/catch block and the feature will return `np.nan` if the convex hull crashes.

# <cleanup> Removed no longer relevant NOTE comment related to `neurom.fst`

It was dating from when we had to maintain the compatibility between neurom.features and neurom.fst

### DIFF
--- a/neurom/apps/morph_stats.py
+++ b/neurom/apps/morph_stats.py
@@ -42,7 +42,7 @@ import pkg_resources
 
 import neurom as nm
 from neurom.exceptions import ConfigError
-from neurom.features import NEURITEFEATURES, NEURONFEATURES
+from neurom.features import NEURITEFEATURES, NEURONFEATURES, _find_feature_func
 from neurom.fst._core import FstNeuron
 
 L = logging.getLogger(__name__)
@@ -65,6 +65,8 @@ def eval_stats(values, mode):
         return values.tolist()
     if mode == 'total':
         mode = 'sum'
+    if mode in {'median', 'mean', 'std'} and len(values) == 0:
+        return None
 
     try:
         return getattr(np, mode)(values, axis=0)
@@ -164,17 +166,17 @@ def extract_stats(neurons, config):
     {config_path}
     """
 
-    def _fill_stats_dict(data, stat_name, stat):
+    def _fill_stats_dict(data, stat_name, stat, feature_name):
         """Insert the stat data in the dict.
 
         And if the stats is a 3D array, splits it into XYZ components.
         """
-        if stat is None or not isinstance(stat, np.ndarray) or stat.shape not in ((3, ), ):
-            data[stat_name] = stat
-        else:
+        if _find_feature_func(feature_name).shape[-1] == 3:
             for i, suffix in enumerate('XYZ'):
                 compound_stat_name = stat_name + '_' + suffix
-                data[compound_stat_name] = stat[i]
+                data[compound_stat_name] = stat[i] if stat is not None else None
+        else:
+            data[stat_name] = stat
 
     stats = defaultdict(dict)
 
@@ -186,14 +188,14 @@ def extract_stats(neurons, config):
         for mode in modes:
             stat_name = _stat_name(feature_name, mode)
             stat = eval_stats(feature, mode)
-            _fill_stats_dict(stats[neurite_type.name], stat_name, stat)
+            _fill_stats_dict(stats[neurite_type.name], stat_name, stat, feature_name)
 
     for feature_name, modes in config.get('neuron', {}).items():
         feature = nm.get(feature_name, neurons)
         for mode in modes:
             stat_name = _stat_name(feature_name, mode)
             stat = eval_stats(feature, mode)
-            _fill_stats_dict(stats, stat_name, stat)
+            _fill_stats_dict(stats, stat_name, stat, feature_name)
 
     return dict(stats)
 
--- a/neurom/apps/morph_stats.py
+++ b/neurom/apps/morph_stats.py
@@ -42,7 +42,7 @@ import pkg_resources
 
 import neurom as nm
 from neurom.exceptions import ConfigError
-from neurom.features import NEURITEFEATURES, NEURONFEATURES, _find_feature_func
+from neurom.features import NEURITEFEATURES, NEURONFEATURES, _get_feature_and_shape
 from neurom.fst._core import FstNeuron
 
 L = logging.getLogger(__name__)
@@ -59,21 +59,17 @@ def eval_stats(values, mode):
         mode: A summary stat to extract. One of:
             ['min', 'max', 'median', 'mean', 'std', 'raw', 'total']
 
-    Note: fails silently if values is empty, and None is returned
+    .. note:: If values is empty, mode `raw` returns `[]`, `total` returns `0.0`
+    and the other modes return `None`.
     """
     if mode == 'raw':
         return values.tolist()
     if mode == 'total':
         mode = 'sum'
-    if mode in {'median', 'mean', 'std'} and len(values) == 0:
+    if len(values) == 0 and mode not in {'raw', 'sum'}:
         return None
 
-    try:
-        return getattr(np, mode)(values, axis=0)
-    except ValueError:
-        pass
-
-    return None
+    return getattr(np, mode)(values, axis=0)
 
 
 def _stat_name(feat_name, stat_mode):
@@ -166,15 +162,15 @@ def extract_stats(neurons, config):
     {config_path}
     """
 
-    def _fill_stats_dict(data, stat_name, stat, feature_name):
+    def _fill_stats_dict(data, stat_name, stat, shape):
         """Insert the stat data in the dict.
 
-        And if the stats is a 3D array, splits it into XYZ components.
+        If the feature is 2-dimensional, the feature is flattened on its last axis
         """
-        if _find_feature_func(feature_name).shape[-1] == 3:
-            for i, suffix in enumerate('XYZ'):
-                compound_stat_name = stat_name + '_' + suffix
-                data[compound_stat_name] = stat[i] if stat is not None else None
+        assert len(shape) <= 2, f'Wrong shape: {shape}'
+        if len(shape) == 2:
+            for i in range(shape[1]):
+                data[f'{stat_name}_{i}'] = stat[i] if stat is not None else None
         else:
             data[stat_name] = stat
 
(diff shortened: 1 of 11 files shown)


neurom/apps/morph_stats.py
```diff
@@ -164,17 +166,17 @@ def extract_stats(neurons, config):
...
-    def _fill_stats_dict(data, stat_name, stat):
+    def _fill_stats_dict(data, stat_name, stat, feature_name):
         """Insert the stat data in the dict.
 
         And if the stats is a 3D array, splits it into XYZ components.
         """
-        if stat is None or not isinstance(stat, np.ndarray) or stat.shape not in ((3, ), ):
-            data[stat_name] = stat
-        else:
+        if _find_feature_func(feature_name).shape[-1] == 3:
```
GianlucaFicarelli, line 174: Minor, what about comparing shape with `(..., 3)` instead of `shape[-1] == 3`? They should give the same result with the features defined at the moment, but it looks more explicit to me (and in general shape can be a tuple with 0 or more elements).

neurom/apps/morph_stats.py
```diff
@@ -186,14 +188,14 @@ def _fill_stats_dict(data, stat_name, stat):
         for mode in modes:
             stat_name = _stat_name(feature_name, mode)
             stat = eval_stats(feature, mode)
-            _fill_stats_dict(stats[neurite_type.name], stat_name, stat)
+            _fill_stats_dict(stats[neurite_type.name], stat_name, stat, feature_name)
 
     for feature_name, modes in config.get('neuron', {}).items():
         feature = nm.get(feature_name, neurons)
```
GianlucaFicarelli, line 194: Thinking about it, we could define another function `get_feature_and_shape`, similar to `get`, that returns the feature and the shape? (it could be private if we don't want to expose it)
In this way we can pass the shape to `_fill_stats_dict`, and we don't need to call the private function `_find_feature_func` to get the function and the shape after we have already got the feature value.

neurom/apps/morph_stats.py
```diff
@@ -164,17 +166,17 @@ def extract_stats(neurons, config):
...
-        if stat is None or not isinstance(stat, np.ndarray) or stat.shape not in ((3, ), ):
-            data[stat_name] = stat
-        else:
+        if _find_feature_func(feature_name).shape[-1] == 3:
             for i, suffix in enumerate('XYZ'):
                 compound_stat_name = stat_name + '_' + suffix
-                data[compound_stat_name] = stat[i]
+                data[compound_stat_name] = stat[i] if stat is not None else None
+        else:
+            data[stat_name] = stat
```
GianlucaFicarelli, line 179: Just to be sure, do we need to do anything special also when shape is `(..., 2)` in the feature `partition_pairs`?

neurom/apps/morph_stats.py
```diff
@@ -65,6 +65,8 @@ def eval_stats(values, mode):
         return values.tolist()
     if mode == 'total':
         mode = 'sum'
+    if mode in {'median', 'mean', 'std'} and len(values) == 0:
+        return None
```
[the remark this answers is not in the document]
wizmer: Good point. Done.

neurom/apps/tests/test_morph_stats.py
```diff
@@ -121,6 +122,13 @@ def test_eval_stats_total_returns_sum():
     assert_equal(ms.eval_stats(np.array([1, 2, 3, 4]), 'total'), 10)
 
 
+def test_eval_stats_on_empty_stat():
+    assert_equal(ms.eval_stats(np.array([]), 'mean'), None)
+    assert_equal(ms.eval_stats(np.array([]), 'std'), None)
+    assert_equal(ms.eval_stats(np.array([]), 'median'), None)
```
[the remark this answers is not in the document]
wizmer: Done

neurom/apps/tests/test_morph_stats.py
```diff
@@ -267,3 +275,22 @@ def test_sanitize_config():
     }
     new_config = ms.sanitize_config(full_config)
     assert_equal(3, len(new_config))  # neurite, neurite_type & neuron
+
+def test_fix_issue_858():
```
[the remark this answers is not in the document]
wizmer: If have changed this test and renamed it because it will now also test the shape (..., 2).

neurom/features/neuritefunc.py
```diff
@@ -533,7 +537,13 @@ def neurite_volume_density(neurites, neurite_type=NeuriteType.all):
     """
     def vol_density(neurite):
         """Volume density of a single neurite."""
-        return neurite.volume / convex_hull(neurite).volume
+        try:
+            volume = convex_hull(neurite).volume
+        except scipy.spatial.qhull.QhullError as e:
+            L.warning(e)
+            return np.nan
```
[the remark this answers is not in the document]
wizmer: I'm afraid I don't have a very good enough reason except that otherwise I can't build the dataframe made of all the features.
I have improved the error message though.

neurom/apps/morph_stats.py
wizmer: Yes. Good point.

neurom/apps/morph_stats.py
wizmer: I have done what you have suggested. I haven't made the function public since we don't need to and so that we can still change the implementation in the future.

neurom/apps/morph_stats.py
```diff
@@ -164,36 +162,36 @@ def extract_stats(neurons, config):
...
-    def _fill_stats_dict(data, stat_name, stat):
+    def _fill_stats_dict(data, stat_name, stat, shape):
         """Insert the stat data in the dict.
 
-        And if the stats is a 3D array, splits it into XYZ components.
+        If the feature is 2-dimensional, the feature is flattened on its last axis
         """
-        if stat is None or not isinstance(stat, np.ndarray) or stat.shape not in ((3, ), ):
-            data[stat_name] = stat
+        assert len(shape) <= 2, f'Wrong shape: {shape}'
```
GianlucaFicarelli, line 170: We still need to agree on that, but I'd prefer to avoid the assert and add:
```
elif len(shape) > 2:
    raise ValueError(f'Feature with wrong shape: {shape}')
```

neurom/features/__init__.py
```diff
@@ -79,14 +103,9 @@ def get(feature_name, obj, **kwargs):
...
     """
-    for feature_dict in (NEURITEFEATURES, NEURONFEATURES):
-        if feature_name in feature_dict:
-            feat = feature_dict[feature_name]
-            break
-    else:
-        raise NeuroMError(f'Unable to find feature: {feature_name}')
+    feat = _find_feature_func(feature_name)
 
     return np.array(list(feat(obj, **kwargs)))
```
GianlucaFicarelli, line 110: I think that now this can be:
```
return _get_feature_and_shape((feature_name, obj, **kwargs))[0]
```

neurom/features/__init__.py
```diff
@@ -70,6 +70,30 @@ def _fun(neurites, neurite_type=_ntype.all):
...
 
+def _find_feature_func(feature_name):
+    """Returns the python function used when getting a feature with `neurom.get(feature_name)`."""
+    for feature_dict in (NEURITEFEATURES, NEURONFEATURES):
+        if feature_name in feature_dict:
+            return feature_dict[feature_name]
+    raise NeuroMError(f'Unable to find feature: {feature_name}')
+
+
+def _get_feature_and_shape(feature_name, obj, **kwargs):
```
GianlucaFicarelli, line 81: Considering that we return the feature value and function, we could call it `_get_feature_value_and_func` or `_get_feature_and_func`?

neurom/features/neuritefunc.py
```diff
@@ -530,10 +534,21 @@ def neurite_volume_density(neurites, neurite_type=NeuriteType.all):
...
     def vol_density(neurite):
         """Volume density of a single neurite."""
-        return neurite.volume / convex_hull(neurite).volume
+        try:
+            volume = convex_hull(neurite).volume
+        except scipy.spatial.qhull.QhullError as e:
+            L.exception('Failure to compute neurite volume using the convex hull. '
+                        'Feature `neurite_volume_density` will return `np.nan`.\n'
+                        'Caught stacktrace:\n %s',
+                        e)
```
GianlucaFicarelli, line 548: Minor, if we log with `L.exception` I think we could avoid logging `Caught stacktrace` and the exception itself.
Anyway I didn't see a real example in this case, so if you are happy with the message I'm fine.

GianlucaFicarelli (commented):
I'm wondering if we should add a changelog file. Any thought?

neurom/features/__init__.py
wizmer: Sorry, I must be tired.

wizmer:
I'm not a big fan of changelogs because I find them redundant with `git log` but yeah in theory we should, no doubt.

GianlucaFicarelli (approved):
Looks good to me.

wizmer:
Thank you !

PR #862, opened 2021-01-08
💥 Change the shape of the output DataFrame of `extract_dataframe`

wizmer:
`extract_dataframe` used to return `N` rows for each neuron where `N` is the
number of neurite types in the input config.

## Context

Since `extract_dataframe` input is a list of neuron, it is more natural to
return exactly one row per neuron.

## Proposed change

Use a MultiIndex on the columns to hold the metrics for each neurite type
instead of having the neurite type appear in the index (axis=0).

## Example: 

Before:

|   | name    | neurite_type    | metric A | metric B |
|---|---------|-----------------|----------|----------|
| 0 | no-axon | axon            | val A    | val B    |
| 1 | no-axon | apical_dendrite | val A'   | val B'   |

After:
<table>
  <col>
  <colgroup span="1"></colgroup>
  <colgroup span="2"></colgroup>
  <colgroup span="2"></colgroup>
  <tr>
    <td rowspan="2"></td>
    <th colspan="1" scope="colgroup">neuron</th>
    <th colspan="2" scope="colgroup">axon</th>
    <th colspan="2" scope="colgroup">apical_dendrite</th>
  </tr>
  <tr>
    <th scope="col">name</th>
    <th scope="col">metric A</th>
    <th scope="col">metric B</th>
    <th scope="col">metric A</th>
    <th scope="col">metric B</th>
  </tr>
  <tr>
    <th scope="row">0</th>
    <td>no-axon</td>
    <td>val A</td>
    <td>val B</td>
    <td>val A'</td>
    <td>val B'</td>
  </tr>
</table>

## :sparkles: Additional modification

The returned dataframe is now sorted as the input list of neuron and support multiple neuron with the same name

In practice

- `imap` is now used instead of `imap_unordered`. Maintaining the sorting is worth the slight performance decrease.
-  the intermediate results are no longer stored in a `dict` but in a `list`. This enables having multiple neuron with the same name. This is for example useful to compare releases.


wizmer:
Ping @arnaudon @adrien-berchet since you are the only actual users of `extract_dataframe`

jdcourcol:
I am using it too.

jdcourcol:
I was wondering why it was not a multiindex btw when using it.

neurom/apps/morph_stats.py
```diff
@@ -124,20 +125,25 @@ def extract_dataframe(neurons, config, n_workers=1):
...
-            for name, data in stats.items()
-            for neurite_type, features in data.items()]
-    return pd.DataFrame(columns=['name', 'neurite_type'] + columns,
-                        data=rows)
+    rows = [[name] + list(chain.from_iterable(features.values() for features in data.values()))
+            for name, data in stats]
+
+    columns = list(chain.from_iterable(
+        [[('neuron', 'name')],
+         product(map(lambda x: x.lower() ,config.get('neurite_type', _NEURITE_MAP.keys())),
```
arnaudon, line 142: format?

arnaudon:
Looks good! That will make our life a bit easier indeed!

lidakanari:
I agree, it looks a better format to me, should simplify things a bit.

wizmer:
Oh. I did not know that this function was so popular :blush:

adrien-berchet:
Looks good to me too :) 
And it seems that Travis CI was also not disabled in this repository.

wizmer:
Lol the CI is actually not running `tox`. I'll fix that first.

wizmer:
retest this please

GianlucaFicarelli (approved):

PR #863, opened 2021-01-08
🐛 Make the CI actually run tox

wizmer:
And add a test file required by the previous commit.


neurom/features/__init__.py
```diff
@@ -60,9 +60,6 @@ def register_neurite_feature(name, func):
         func: single parameter function of a neurite.
 
     """
-    if name in NEURITEFEATURES:
-        raise NeuroMError('Attempt to hide registered feature %s' % name)
-
```
GianlucaFicarelli, line 65: Why this is removed? I was thinking we want to avoid overwriting existing features (same behaviour as with the decorator).

wizmer: Because this function calls `_register_feature` which already performs this check.

GianlucaFicarelli: Right, I missed that call 👍

GianlucaFicarelli (approved):

PR #865, opened 2021-01-25
do not compute dangling check if no dendrites

arnaudon:
this crashes the error detection on morphologies with only axon, but should not

### DIFF
--- a/neurom/check/neuron_checks.py
+++ b/neurom/check/neuron_checks.py
@@ -304,6 +304,10 @@ def has_no_dangling_branch(neuron):
                                                          for n in iter_neurites(neuron)
                                                          if n.type != NeuriteType.axon)))
 
+    if not len(dendritic_points) > 0:
+        # returns True if no dendrites are present
+        return CheckResult(True)
+
     def is_dangling(neurite):
         """Is the neurite dangling ?."""
         starting_point = neurite.points[0][COLS.XYZ]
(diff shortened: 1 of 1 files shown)


arnaudon:
@wizmer , I can add missing test if you think this PR is interesting.

neurom/check/neuron_checks.py
```diff
@@ -304,6 +304,10 @@ def has_no_dangling_branch(neuron):
                                                          for n in iter_neurites(neuron)
                                                          if n.type != NeuriteType.axon)))
 
+    if len(dendritic_points) == 0:
+        # returns True if no dendrites are present
+        return CheckResult(True)
```
wizmer, line 309: The behavior is incorrect, a dangling axon would then go undetected. I'll have a look tomorrow.

arnaudon: yes, I'm not sure about what dangling is. It seems to be w.r.t to dendrite's first points, but then if there are no dendrites, it cannot be dangling...

lidakanari: I agree with @wizmer, I think the non-dendritic trees should be checked as well.

arnaudon:
In fact, I think I will close this PR, and just not try to detect dangling if a morphology just as an axon.

wizmer:
I think you can let it open since it still relates to a real issue. I'll have a look

arnaudon:
Why 12? It should be a function parameter, right?
```       
 if np.linalg.norm(starting_point - soma_center) - soma_max_radius <= 12.:
```

arnaudon:
I'll close this, as https://github.com/BlueBrain/NeuroR/pull/54 does the job implicetely by catching exceptions.

PR #755, opened 2019-05-08
Migrate to MorphIO

wizmer:
The NeuroM neuron class inherits from morphio.mut.Morphology

PR #869, opened 2021-02-16
Add plotting option to display NEURON secid on morphology

arnaudon:
Convenient for various manual NEURON related tasks.

It can be used as follow:
```
neurom view --plane xy --nrn-secid my_morph.asc
```
and will display the NRN section it (from `all `section list) near the center of each section. We may want to add the option to choose from which seclist to plot the ids, if needed.


arnaudon:
@wizmer, I'm getting these here as well ;)
```
  h5v2 is no longer supported, see: https://github.com/BlueBrain/MorphIO#H5v2
```

arnaudon:
I think I need mut_morphio to be able to save a tmp neuron, and use NEURON to get proper secids. Not sure how to do that best, but it seems to work!

wizmer:
Adding a morphio dependency breaks the Windows support. That's why we haven't done it yet. So we can't use morph-tool with the v1.
Feel free to re-open when the morphio integration has been merged: https://github.com/BlueBrain/NeuroM/pull/755

arnaudon:
Arg ok, I see. Can't we leave it open for the time being, so we can use it?

arnaudon:
- add option to select which section type (apical for example)
- add option to set the fontsize

arnaudon:
@asanin-epfl , what do you think of this PR? it is convenient tool to use when one want to see which section number NEURON uses on the morphology. @mariarv  and me are using it from times to times. I could be nice to have it somewhere safe.

arnaudon:
arg, the rebase looks like a real pain... @adrien-berchet ? :P

asanin-epfl:
76 files are affected for a plotting option? Are all those files truly related to the PR? Could it be faster to recreate the PR from the latest master?

arnaudon:
No, only a few... I'll make a fresh PR, as you did for sholl!

arnaudon:
Closing in favor of https://github.com/BlueBrain/NeuroM/pull/920 to bypass the rebase

PR #871, opened 2021-03-01
fix sholl feature and update docstring

arnaudon:
sholl_crossings is not a feature accessible from get, so we drop the feature decorator. It can be used with a list of sections as well, so we added more in the docstring.

### DIFF
--- a/neurom/features/neuronfunc.py
+++ b/neurom/features/neuronfunc.py
@@ -212,7 +212,7 @@ def trunk_angles(nrn, neurite_type=NeuriteType.all):
 
 
 @feature(shape=(...,))
-def sholl_crossings(neurites, center, radii):
+def sholl_crossings(neurites, center, radii, section_filter=None):
     """Calculate crossings of neurites.
 
     Args:
@@ -227,7 +227,7 @@ def sholl_crossings(neurites, center, radii):
         """Used to count_crossings of segments in neurite with radius."""
         r2 = radius ** 2
         count = 0
-        for start, end in iter_segments(neurite):
+        for start, end in iter_segments(neurite, section_filter=section_filter):
             start_dist2, end_dist2 = (morphmath.point_dist2(center, start),
                                       morphmath.point_dist2(center, end))
 
@@ -242,7 +242,7 @@ def sholl_crossings(neurites, center, radii):
 
 
 @feature(shape=(...,))
-def sholl_frequency(nrn, neurite_type=NeuriteType.all, step_size=10):
+def sholl_frequency(nrn, neurite_type=NeuriteType.all, step_size=10, section_filter=None):
     """Perform Sholl frequency calculations on a population of neurites.
 
     Args:
@@ -271,11 +271,13 @@ def sholl_frequency(nrn, neurite_type=NeuriteType.all, step_size=10):
                               if neurite_filter(neurites)))
 
         min_soma_edge = min(min_soma_edge, neuron.soma.radius)
-        max_radii = max(max_radii, np.max(np.abs(bounding_box(neuron))))
+        max_radii = max(max_radii,
+            max(np.max(np.abs(bounding_box(neurite))) for neurite in iter_neurites(neuron)
+            if neurite_filter(neurite)))
 
     radii = np.arange(min_soma_edge, max_radii + step_size, step_size)
     ret = np.zeros_like(radii)
     for neurites, center in neurites_list:
-        ret += sholl_crossings(neurites, center, radii)
+        ret += sholl_crossings(neurites, center, radii, section_filter=section_filter)
 
     return ret
--- a/neurom/features/neuronfunc.py
+++ b/neurom/features/neuronfunc.py
@@ -222,7 +222,8 @@ def sholl_crossings(neurites, center, radii):
                                                        radii=np.arange(0, 1000, 100))
 
     Args:
-        nrn(morph): morphology on which to perform Sholl analysis, or list of sections
+        neurites(list): morphology on which to perform Sholl analysis, or list of neurites
+        center(Point): center point
         radii(iterable of floats): radii for which crossings will be counted
 
     Returns:
@@ -241,13 +242,8 @@ def sholl_crossings(neurites, center, radii):
 
         return count
 
-    if hasattr(neurites, 'neurites') or hasattr(neurites, 'points'):
-        neurites = list(iter_neurites(neurites))
-    else:
-        neurites = list(neurites)
-
     return np.array([sum(_count_crossings(neurite, r)
-                         for neurite in neurites)
+                         for neurite in iter_neurites(neurites))
                      for r in radii])
 
 
--- a/neurom/features/neuronfunc.py
+++ b/neurom/features/neuronfunc.py
@@ -214,6 +214,13 @@ def trunk_angles(nrn, neurite_type=NeuriteType.all):
 def sholl_crossings(neurites, center, radii):
     """Calculate crossings of neurites.
 
+    This function can also be used with a list aa neurites, as follow:
+
+        secs = (sec for sec in nm.iter_sections(neuron) if complex_filter(sec))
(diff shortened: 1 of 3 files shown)


mgeplf:
What is the meaning of `sholl analysis` on a subset of sections?  Is the result different than the sholl analsys on the whole thing, and then just examining those sections?

mariarv:
yes, it's different. The meaning is number of crossing of the dendrites on some fixed distance from the soma, so depending which dendrites you consider (basal, apical oblique or tuft) the outcome of the function will vary

wizmer:
Adding a new argument to all features could be doable but I also wonder if we could not simply allow sequences of sections to be passed as the first argument of the features instead ?

Example:

```python
interesting_sections = (sec for sec in iter_section(neuron) if complicated_filter(section))
result = sholl_crossings(interesting_sections, center, radii)
```

arnaudon:
ah yes, this seems to work as well. Not sure what is easier to implement so it can be reused in other features. I can give it a try.

eleftherioszisis:
Given there is a very simple solution to this, isn't it a bit of an overkill to introduce a `section_filter`? imho it is cleaner filtering the way Benoit showed, rather than passing lambdas.

arnaudon:
yes, totally agree! But what benoit wrote does not work yet, right? I'll try that tomorrow.

arnaudon:
here you are!

arnaudon:
@mariarv are you satisfied with this? shall we merge?

mariarv:
Sure!

neurom/features/neuronfunc.py
```diff
@@ -216,7 +216,7 @@ def sholl_crossings(neurites, center, radii):
     """Calculate crossings of neurites.
```
arnaudon, line 216: this is not a feature with @feature

arnaudon: @asanin-epfl , there is a small bug here

arnaudon:
@asanin-epfl , can we merge this as well, as it has a small bug fix, in addition to more flexible use of sholl feature? thanks!

neurom/features/neuronfunc.py
mgeplf: > this is not a feature with @feature

Why isn't this a feature?

arnaudon: it does not have the correc arguments, and is only used for the real feature sholl_frequency below. It crashed neurom for me saying missing values for center and radii arguments. But weird it didn't crash before. We need to investigate that

neurom/features/neuronfunc.py
```diff
@@ -211,12 +211,11 @@ def _sort_angle(p1, p2):
             for i, _ in enumerate(ordered_vectors)]
 
 
-@feature(shape=(...,))
 def sholl_crossings(neurites, center, radii):
     """Calculate crossings of neurites.
 
     Args:
-        nrn(morph): morphology on which to perform Sholl analysis
+        nrn(morph): morphology on which to perform Sholl analysis, or list of sections
```
asanin-epfl, line 218: update docstring according to function arguments

neurom/features/neuronfunc.py
```diff
@@ -230,14 +229,18 @@ def _count_crossings(neurite, radius):
         for start, end in iter_segments(neurite):
             start_dist2, end_dist2 = (morphmath.point_dist2(center, start),
                                       morphmath.point_dist2(center, end))
-
             count += int(start_dist2 <= r2 <= end_dist2 or
                          end_dist2 <= r2 <= start_dist2)
 
         return count
 
+    if hasattr(neurites, 'neurites') or hasattr(neurites, 'points'):
```
asanin-epfl, line 237: This block does not make much sense. Why `neuron_population` function can't be reused. Why cast everything to `list`? There was a nice `iter_neurites` below which now is dropped. Why?

neurom/features/neuronfunc.py
arnaudon: it is not in the list in neuronfeatures that used to be here: https://github.com/BlueBrain/NeuroM/commit/ed3d9009e83cd3a742d86c82cfe23b905259e66c
In the test of the get function, it tests various features, but not this one (and not all possible features using `from neurom.apps.morph_stats import full_config` 
So I think this one could get pass undetected.

neurom/features/neuronfunc.py
arnaudon: that seem to be missing from before, I added it, good catch!

neurom/features/neuronfunc.py
arnaudon: this is the simplest I could do to accept a list of sections. Maria gets a list of section (mabye some in the apical tuft), and wants to compute sholl. The iter_neurites is still under the if l 237, which works if you give a neuron or a neurites, but not a list of sections. I had to convert to list fr some reason I don't recall

arnaudon:
I updated the docstring with an example of how to use this new option (copied from above in this PR).

neurom/features/neuronfunc.py
arnaudon: maybe there is a cleaner way to do this, but I couldn't figure it out. Maybe with v2?

asanin-epfl: Lets try this function. Can it be used instead? Also I've updated its docstring more.
```python
def sholl_crossings(neurites, center, radii):
    """Calculate crossings of neurites.

    This function can also be used with a list aa neurites, as follow:

        secs = (sec for sec in nm.iter_sections(neuron) if complex_filter(sec))
        sholl = nm.features.neuronfunc.sholl_crossings(secs,
                                                       center=neuron.soma.center,
                                                       radii=np.arange(0, 1000, 100))

    Args:
        neurites(list): morphology on which to perform Sholl analysis, or list of neurites
        center(Point): center point
        radii(iterable of floats): radii for which crossings will be counted

    Returns:
        Array of same length as radii, with a count of the number of crossings
        for the respective radius
    """
    def _count_crossings(neurite, radius):
        """Used to count_crossings of segments in neurite with radius."""
        r2 = radius ** 2
        count = 0
        for start, end in iter_segments(neurite):
            start_dist2, end_dist2 = (morphmath.point_dist2(center, start),
                                      morphmath.point_dist2(center, end))
            count += int(start_dist2 <= r2 <= end_dist2 or
                         end_dist2 <= r2 <= start_dist2)

        return count

    return np.array([sum(_count_crossings(neurite, r)
                         for neurite in iter_neurites(neurites))
                     for r in radii])
```

asanin-epfl: > this is the simplest I could do to accept a list of sections. Maria gets a list of section (mabye some in the apical tuft), and wants to compute sholl. The iter_neurites is still under the if l 237, which works if you give a neuron or a neurites, but not a list of sections. I had to convert to list fr some reason I don't recall

Can you add a test then for this case of list of sections?

neurom/features/tests/test_neuronfunc.py
```diff
@@ -176,6 +176,9 @@ def test_sholl_crossings_simple():
     nt.eq_([2, 4, 5],
            list(_nf.sholl_crossings(SIMPLE, center, radii=radii)))
 
+    nt.eq_([2, 2, 2],
```
arnaudon, line 179: @asanin-epfl I added the test here in fact! I'm surprising myself that I added a test!

asanin-epfl: Then I'm very surprised that it passed with the function from my comment above https://github.com/BlueBrain/NeuroM/pull/871#discussion_r599594919. Anyway. Check this new function, and if it works, lets use it.

arnaudon: yes, weird. I recall it was not working when I tried... something may have changed. Let's go like that then.

arnaudon:
Ok, so this PR is simpler than original thought! Thanks for pointing that, @asanin-epfl ! I let you merge then you want!

asanin-epfl (approved):

PR #876, opened 2021-03-05
Stop supporting h5v2 (#227)

wizmer:
If a h5v2 file is detected at loading time, the error message will point the user to the documentation for a way to convert it to h5v1.

https://github.com/BlueBrain/MorphIO#h5v2


tomdele (approved):
LGTM !

wizmer:
T'es au top topper !

PR #877, opened 2021-03-05
bounding box features

arnaudon:


eleftherioszisis:
The `lateral_dendrite_bbox` across the x-direction is equivalent to the `total_width` and the `vertical_dendrite_bbox` to the `total_height` features respectively. Given that these two features combined with the `total_depth` render the calculation of the `volume_dendrite_bbox` trivial by multiplying all three, isn't it redundant to add the latter as a new feature?

arnaudon:
yes, you're right, let's close this

PR #883, opened 2021-03-24
register all features without exception for those that start with 'n_'

asanin-epfl:


neurom/features/tests/test_get_features.py
```diff
@@ -490,7 +490,10 @@ def test_segment_meander_angles_single_section():
 def test_neurite_features_accept_single_tree():
     for f in NEURITEFEATURES:
         ret = get_feature(f, NRN.neurites[0])
-        nt.ok_(ret.dtype.kind in ('i', 'f'))
+        if isinstance(ret, np.ndarray):
+            nt.ok_(ret.dtype.kind in ('i', 'f'))
+        else:
+            nt.assert_true(np.isscalar(ret))
         if len(features._find_feature_func(f).shape) >= 1:
             nt.ok_(len(ret) > 0)
```
adrien-berchet, line 498: I propose the following simplification:
```suggestion
        if isinstance(ret, np.ndarray):
            nt.ok_(ret.dtype.kind in ('i', 'f'))
            nt.ok_(len(features._find_feature_func(f).shape) >= 1)
            nt.ok_(len(ret) > 0)
        else:
            nt.assert_true(np.isscalar(ret))
```

Also, shouldn't we either use `nt.assert_true`, either always `nt.ok_`?

asanin-epfl: Very good. Lets use it. I've adjusted your suggestion to replace `assert_true` with `ok_`

adrien-berchet (approved):
Looks good!

asanin-epfl:
very well then, thank you

asanin-epfl:
@tomdele the PR is reviewed already. Because Adrien does not have permission to approve, can you please do it instead?

tomdele (approved):
Legit !

asanin-epfl:
Thank you!

Issue #887, opened 2021-04-14
RFE: Move away from using deprecated nose for testing


sanjayankur31:
# Feature
nose is deprecated, so it'll be good to move to another testing tool, perhaps py.test.

# Motivation
nose is deprecated upstream (the last commit was in 2016.) and is also being deprecated in Fedora.

https://github.com/nose-devs/nose/commits/master

It is still in use here, though

https://github.com/BlueBrain/NeuroM/blob/master/requirements_dev.txt#L32

The nose website says:

https://nose.readthedocs.io/en/latest/

"Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2."

Deprecation of nose in Fedora: https://fedoraproject.org/wiki/Changes/DeprecateNose

Alternatives
py.test or nose2 (but nose2 doesn't seem to be at par with nose or py.test)

# Additional context

This came up while working on including NeuroM in Fedora as part of the [NeuroFedora](neuro.fedoraproject.org) initiative:

https://pagure.io/neuro-sig/NeuroFedora/issue/193

mgeplf:
@asanin-epfl I think Werner had good luck w/ https://github.com/pytest-dev/nose2pytest
after applying some transforms:

https://github.com/BlueBrain/BluePyOpt/blob/c3fe5c3c53f07d946031974794faf365243f9d7a/misc/pytest_migration/convert_pytest.sh

Hopefully something similar would make work here.

asanin-epfl:
Thank you! I think there is no need for the script, if using the latest master of nose2ytest. Some minor adjustments still have to be done.

PR #888, opened 2021-04-15
Code cleaning of apps package

asanin-epfl:
standalone `apps` package have been moved under `neurom` package, refactored and covered with tests.

PR #892, opened 2021-04-16
nose to pytest, move all tests to 'tests' folder

asanin-epfl:
Solution to #887


tox.ini
```diff
@@ -2,7 +2,7 @@
 name = neurom
 testdeps =
     mock
-    nose
```
mgeplf, line 5: holy smokes that was fast

asanin-epfl: There are still problem with tests locations, I have to move them to `tests` folder for simplicity

tests/view/conftest.py
```diff
@@ -4,15 +4,14 @@
...
-
 from neurom.view import common
 common._get_plt()
 
 from neurom.view.common import plt
+import pytest
 
 
-@contextmanager
+@pytest.yield_fixture
```
GianlucaFicarelli, line 14: With pytest >= 3.0 this should be just `@pytest.fixture`, because from https://docs.pytest.org/en/stable/yieldfixture.html:

> Marking functions as yield_fixture is still supported, but deprecated and should not be used in new code.

tests/view/conftest.py
```diff
@@ -22,7 +21,7 @@ def get_fig_2d():
         plt.close(fig)
 
 
-@contextmanager
+@pytest.yield_fixture
```
GianlucaFicarelli, line 24: See previous comment.

GianlucaFicarelli (commented):
Nice, this is a big refactoring of tests.
Just mentioning that I got a few skipped/warning messages:
```
472 passed, 2 skipped, 24 warnings
```

asanin-epfl:
> ```
> 472 passed, 2 skipped, 24 warnings
> ```

I kept them from NeuroM v1. I will deal with them in cleaning PRs like this one #888

tests/view/conftest.py
asanin-epfl: Thank you! I was unaware.

GianlucaFicarelli (approved):

asanin-epfl:
thank you!

PR #893, opened 2021-04-16
add version to the package

asanin-epfl:
@GianlucaFicarelli spotted the current invalid package versioning, so we published 2.0.0 as https://pypi.org/project/neurom/0.0.0/.


GianlucaFicarelli (approved):

setup.py
```diff
@@ -69,5 +70,6 @@
         'Programming Language :: Python :: 3.8',
         'Topic :: Scientific/Engineering :: Bio-Informatics',
     ],
+    use_scm_version={"local_scheme": "no-local-version"},
```
mgeplf, line 73: weird, why/how was it working before?

asanin-epfl: This PR brought back the way it worked before. During the PR of migration to v2, this line was lost.

PR #898, opened 2021-04-26
clean imports

asanin-epfl:
- drop relative imports
- delete unused PointType class
- don't use private modules, rename `_neuron` to `neuron`, `_soma` to `soma`


arnaudon (approved):

asanin-epfl:
@tomdele Alexis's approval does not allow me to merge. You will have to look here.

tomdele (approved):
look good

asanin-epfl:
Thank you!

adrien-berchet:
Hi @asanin-epfl 
No big deal but this PR introduces breaking changes due to file renamings. It would have been nice to increase the minor version number instead of just the bugfix number ;-)

asanin-epfl:
Yes! I totally forgot. Should not be critical as private classes become public.

adrien-berchet:
No problem, it is not critical indeed.

asanin-epfl:
Thank for the heads up!

arnaudon:
Wait, this could break a few other codes, no? No other complaints so far?

asanin-epfl:
It will break only if user imported the private module.

arnaudon:
ah ok ok , let see, it may just be me in diameter-synthesis then

adrien-berchet:
Not only since imports were removed from the `neurom.core.__init__`.
So all `from neurom.core import *` must be changed to `from neurom.core.neuron import *`. Are these considered as private?

asanin-epfl:
ok ok, 2.1.0

adrien-berchet:
Thanks :-)

adrien-berchet:
Now I am wondering if this change worth it. It actually breaks at least TNS, morph-tool and diameter-synthesis. Maybe reverting the imports in `neurom/core/__init__.py` would be simpler than updating everything else?

asanin-epfl:
ok

adrien-berchet:
Hard to check everything but I guess we shoud also revert `neurom/view/__init__.py` and `neurom/io/__init__.py`

asanin-epfl:
This one no `neurom/io/__init__.py`. It duplicates already `neurom/__init__.py`.

adrien-berchet:
But maybe another code was importing something like `neurom.io.NeuronLoader` instead of `neurom.NeuronLoader` so it would still be breaking for it. I don't think we have this kind of imports in TNS, diameter-synthesis and morph-tool so it should not bother me but maybe it's the case somewhere else so it's as you prefer.

asanin-epfl:
yes but in that case another code needs to change such import without changing the version of NeuroM in its `setup.py`

adrien-berchet:
Yeah true, it's ok for me then.

neurom/core/dataformat.py
```diff
@@ -43,15 +44,9 @@ class COLS:
     XYZR = slice(0, 4)
 
 
-class POINT_TYPE:
```
tomdele, line 46: [not for this review] This is so weird having this as caps .... Same for COLS

neurom/core/dataformat.py
```diff
@@ -43,15 +44,9 @@ class COLS:
...
-    """Point types.
+Point = namedtuple('Point', ('x', 'y', 'z', 'r'))
 
-    These follow SWC specification.
-    """
-    (UNDEFINED, SOMA, AXON, BASAL_DENDRITE, APICAL_DENDRITE,
-     FORK_POINT, END_POINT, CUSTOM) = range(8)
```
tomdele, line 52: [not for this review]  Same for a enum like object factorization like this are so misleading ...

tomdele (commented):
looks good

PR #902, opened 2021-04-29
Account for all custom neurite types in NeuriteType

asanin-epfl:


neurom/core/types.py
```diff
@@ -53,7 +53,13 @@ class NeuriteType(IntEnum):
     undefined = SectionType.undefined
     soma = 31
     all = 32
-    custom = SectionType.custom5
+    # custom = SectionType.custom5
```
adrien-berchet, line 56: dead code?

arnaudon:
ah, what is that? could you give some context to it? I'm thinking of adding custom neurite types in fact. For apical and axons, some branches have 'subtypes', could I use that?

neurom/core/types.py
```diff
@@ -53,7 +53,13 @@ class NeuriteType(IntEnum):
...
+    custom5 = SectionType.custom5
+    custom6 = SectionType.custom6
+    custom7 = SectionType.custom7
+    custom8 = SectionType.custom8
+    custom9 = SectionType.custom9
+    custom10 = SectionType.custom10
```
adrien-berchet, line 62: Why is it from 5 to 10? I never used this `custom*` thing.
Maybe we could do something dynamic that just returns the same 'custom*' as the `SectionType`'s one? So if new customs are added into morphio we will not need to update this.

neurom/core/types.py
asanin-epfl: done

neurom/core/types.py
asanin-epfl: It is due to https://github.com/BlueBrain/MorphIO/blob/bb8f1995e42d172d7007f4c637f7a399915f84ef/include/morphio/enums.h#L66.
Basically I just follow what MorphIO does.

arnaudon: ah ok, so it's just in case some swc files have higher numbers for section types than what we usualy expect?

asanin-epfl: yes

evancresswell: Non neuronal cells on Neuromorpho can have custom section types (ie custom 5). When plotting the dendrogram this crashes program

ex cell: http://neuromorpho.org/neuron_info.jsp?neuron_name=S10_1_1

adrien-berchet: ok

asanin-epfl: @adrien-berchet 5-10 is according to the spec http://neuromorpho.org/StdSwc1.21.jsp, see section **Tag Irregularities**

asanin-epfl:
@GianlucaFicarelli or @joni-herttuainen please approve the PR. Approvals from non-team members do not allow me to merge.

arnaudon:
yes, we are so useless haha

joni-herttuainen (approved):

asanin-epfl:
Thank you!

PR #904, opened 2021-04-30
bring back imports to __init__ of core and view for backward compatib…

asanin-epfl:
…ility

### DIFF
--- a/neurom/core/__init__.py
+++ b/neurom/core/__init__.py
@@ -27,3 +27,8 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 """Core functionality and data types of NeuroM."""
+
+# those imports here for backward compatibility
+from neurom.core.soma import Soma
+from neurom.core.neuron import Section, Neurite, Neuron
+from neurom.core.population import Population
--- a/neurom/view/__init__.py
+++ b/neurom/view/__init__.py
@@ -27,3 +27,7 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 """View tools to visualize morphologies."""
+from neurom.view.view import (plot_neuron, plot_neuron3d,
+                              plot_tree, plot_tree3d,
+                              plot_soma, plot_soma3d,
+                              plot_dendrogram)


neurom/core/__init__.py
```diff
@@ -27,3 +27,8 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 """Core functionality and data types of NeuroM."""
+
+# those imports here for backward compatibility
+from neurom.core.soma import Soma
+from neurom.core.neuron import Section, Neurite, Neuron
+from neurom.core.population import Population
```
adrien-berchet, line 34:
```suggestion
from neurom.core.population import Population
from neuron.core.types import NeuriteType, NeuriteIter
```

neurom/core/__init__.py
```diff
@@ -27,3 +27,8 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 """Core functionality and data types of NeuroM."""
+
+# those imports here for backward compatibility
+from neurom.core.soma import Soma
+from neurom.core.neuron import Section, Neurite, Neuron
```
adrien-berchet, line 33:
```suggestion
from neurom.core.neuron import (Section, Neurite, Neuron, iter_neurites,
                                iter_sections, iter_segments, graft_neuron)
```

adrien-berchet: shouldn't we keep the same previous API?

neurom/core/__init__.py
adrien-berchet: shouldn't we keep the same previous API?

adrien-berchet (commented):
Thanks for these changes!

Also, `neurom/geom/__init__.py` is missing `from .transform import translate, rotate` now, maybe it is better to revert this one as well?

neurom/core/__init__.py
asanin-epfl: `NeuriteType, NeuriteIter` are part of `neurom/__init__.py`

neurom/core/__init__.py
asanin-epfl: `iter_neurites, iter_sections, iter_segments, graft_neuron` are part of `neurom/__init__.py`

adrien-berchet: hmmm yeah. But it's still breaking then. I don't know what's the best but in morph-tool for example, you set the requirement as `'neurom>=2.0,<3'` so we could suppose that no change would be needed before `NeuroM==3`.

asanin-epfl: well, it will break but you don't need to change `neurom>=2.0,<3` to `neurom>=2.1,<3`. Just update imports and keep `neurom>=2.0,<3`

adrien-berchet: Yes. It does not comply with semantic versioning but I don't know the BBP policy about that (if there is any?).
Anyway, we can revert only these and update the other codes if you think it's better.

adrien-berchet: Sure, I agree. But also the less break the better :grin: 
But let's go for the less duplication.

adrien-berchet (approved):

adrien-berchet:
Will you overwrite the 2.1 tag or create a new one?

asanin-epfl:
create 2.1.1, delete 2.1
@joni-herttuainen  I will need your approval to merge

adrien-berchet:
Ah yes cool. Thanks!

joni-herttuainen (approved):

PR #907, opened 2021-05-06
Update AUTHORS.md

alex4200:
Update of format to follow new three-section format.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -1,17 +1,15 @@
-Main Contributors
-=================
+Maintainer
+----------
+* Michael Gevaert
 
-Main contributors to code and ideas in alphabetical order:
-----------------------------------------------------------
+Authors
+-------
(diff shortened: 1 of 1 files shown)


AUTHORS.md
```diff
@@ -1,17 +1,15 @@
...
+
+Authors
+-------
+* Benoit Coste
+* Mike Gevaert
+* Lida Kanari
+* Juan Palacios
+* Eleftherios Zisis
+
+Contributors
```
asanin-epfl, line 13: What's the point of empty contributors?

alex4200: Just to show that there is such a section (for presentation purposes). Akiko will use that file as an example on how the AUTHORS file should be formatted in the future.

asanin-epfl: ok

asanin-epfl (approved):

Issue #786, opened 2019-12-13
'morph_stats' output: just one output for morphology files equally named but with different extensions


pedroernesto:
Recent versions of NeuroM do not include the extension of the original morphology files in their _morph-stats_ output . Example:

When running _morph_stats_ on a directory with 2 files C230300D1-rep-cor.h5 and C230300D1-rep-cor.swc:
1) the output file contains information about just one instead of two cells
2) the extension of the source file is not included

{
   "C230300D1-rep-cor": {
      "axon": {
         "max_section_branch_order": 15, 
         "total_neurite_length": 14302.22999937718, 
         "total_number_of_neurite": 1
      }, 
      "basal_dendrite": {
         "max_section_branch_order": 3, 
         "total_neurite_length": 2398.5855940959173, 
         "total_number_of_neurite": 7
      }, 
      "total_soma_radius": 12.9966
   }
}

asanin-epfl:
Done via #909

PR #909, opened 2021-05-06
Include neuron filename extension into its name

asanin-epfl:
Solution to #786,

Useful for morph stats when morphologies differ only by file extension


adrien-berchet (approved):
LGTM
@arnaudon do you think it could break anything?

arnaudon:
it may not break anything, but I'm not sure this is a good option, different file format should have the same answer in morph_stat. If we want to test the possible differences, one can name 'my_morph_asc.asc' and 'my_morph_swc.swc' to compare the, no?

asanin-epfl:
> different file format should have the same answer in morph_stat.

Yes but it must not to overwrite one file extension with another. Besides there might be different morphologies under the same name with different extensions. Distinguishing among file extensions, will help user to detect this problem.


> If we want to test the possible differences, one can name 'my_morph_asc.asc' and 'my_morph_swc.swc' to compare the,

So, you manually added file extensions to names. How is that different from having file extensions?

arnaudon:
Yes, this is a difficult thing to settle. My approach to it is to only distinguish morphologies by their names. I only work with a set of morphologies with consistent extensions, to avoid these issues. What I meant is that, if, as in the issue that triggered this PR, one want to extract morph_stats on 2 morphs with only a different extension (which is a very specific usecase I never had myself), one can change the names of the morphs to do that, instead of this PR. 
What I like to have in a df for morphologies, is a column name with just Path('morph.asc').stem, and a column with the path to that morph on disk (where the extension is written). But this is just my way of doing things, it may not be the best!

But you can merge that, and we'll see in practice. It possibly won't affect anything at all.

asanin-epfl:
Thank you, I'm merging.

PR #911, opened 2021-05-07
explicitly describe unifurcation case in migration documentation

asanin-epfl:


arnaudon (approved):

asanin-epfl:
That was fast, thank you!

PR #905, opened 2021-05-03
Refactor Sholl analysis code, fix #663

asanin-epfl:
Neuron attributes soma center and soma radius were being computed for each neurite. Now they are computed only once.

PR #919, opened 2021-05-17
improve sholl

arnaudon:
This reverts part of https://github.com/BlueBrain/NeuroM/pull/905 (related to step_size argument, crucial for sholl) and fixes https://github.com/BlueBrain/NeuroM/issues/663 by replacing the box on the entire neurons, to max radial distance using the neurite filter. Now, we are guaranteed to have sholl shells every step_size, until the last sphere before there is no more crossing. Still we can provide a custom bins for more flexibility.

### DIFF
--- a/neurom/features/neuronfunc.py
+++ b/neurom/features/neuronfunc.py
@@ -40,8 +40,6 @@ from neurom.core.types import NeuriteType
 from neurom.core.types import tree_type_checker as is_type
 from neurom.features import feature
 
-from neurom.geom import bounding_box
-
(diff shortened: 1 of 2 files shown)


arnaudon:
I don't think we need more tests, no? @mgeplf  you were wondering about the crossing twice case? From docstring it seems these are counted twice, but I don't know what is the standard way to count these (@mariarv @lidakanari , do you know?)

mariarv:
@arnaudon it's acceptable to have double crossings in Sholl analysis

asanin-epfl (approved):
good to go then

PR #920, opened 2021-05-18
Option to plot nrn_secid labels

arnaudon:


neurom/view/view.py
```diff
@@ -188,13 +190,33 @@ def plot_soma(ax, soma, plane='xy',
                                    ignore=False)
 
 
+def _get_isec_map(nrn, nrn_secid):
+    """Get NEURON isec from nrn morphology."""
+    from morph_tool import nrnhines
+    import tempfile
+
+    with tempfile.TemporaryDirectory() as tmpdirname:
```
eleftherioszisis, line 198: Do you need the directory? A NamedTemporaryFile would not do as well?

arnaudon: Ah good point, I think I did it this way after a very quick googling ;) I'll update!

neurom/view/view.py
```diff
@@ -188,13 +190,32 @@ def plot_soma(ax, soma, plane='xy',
                                    ignore=False)
 
 
+def _get_isec_map(nrn, nrn_secid):
+    """Get NEURON isec from nrn morphology."""
+    from morph_tool import nrnhines
```
mgeplf, line 195: Having this dependency added makes things hard; can't it be in `morph-tool`?

arnaudon: Yes, you are right, I was looking to see If we could avoid pulling BluePyOpt, but for that, we'll need quite a few lines of code. I can move the plotting to morph-tool, but not the CLI, which is quite convenient. I can make another CLI in morph-tool, mimicking this one eventually, or just drop the CLI.

mgeplf: Yeah, the ecosystem is tough.

Perhaps there's some way to make the neurom cli slightly more pluggable/extendable, and then have morph-tool attach itself to that.

arnaudon: Why not, no idea how to do that, I'll look around.

mgeplf: I have no idea if it's possible either, but if it can be done cleanly, that's one option.

asanin-epfl: > Why not, no idea how to do that, I'll look around.

@arnaudon don't waste time on this. Bringing morph-tool to neurom definitely can't be done. Extending CLI of neurom by morph-tool seems magical. I've never seen anything like installing one library to extend CLI of another. Lets just move this to morph-tool. It seems appropriate. If you want, I can move it.

arnaudon: No worries, I wouldn't have wasting time on this ;) One thing we may want to discuss is the plotting of morphologies. I recall benoit being a bit against having them in neurom/morph-tool, and there is plotly-helper already on github. I wonder if it would make sense to have a neuron plotting library, in which we can do fancy stuff without thinking of pulling extra dependencies to the host code? I have in mind a few other plotting functions, which may be useful for some people. We can for now create a plotting.py in morph-tool if you think this is the best option.

arnaudon:
So I will move this to morph-tool asap, and close this one afterwards.

neurom/view/view.py
arnaudon: We notices with aleksei that morph-tool already has a plotting module for extension of dendrogram plot with synapses, so in this regard, it makes sense to add other plotting functions based on neurom ones for specific use cases, no?

asanin-epfl: morph-tool already has plotting, so it is safe to move it there. Building an extendable CLI for neurom is too much.

Issue #538, opened 2016-10-25
morph_stats should support a large number of morphologies


jdcourcol:
Currently, using a lot of morphologies (> 10000) uses a lot of memory and that makes morph_stats to hang for ever.
The reason being that all the statistics are being kept in memory and then written into a report (json or csv).

mgeplf:
I can probably fix this for the CSV version, but I would have to look at the best way to incrementally write json files.

juanchopanza:
Careful with the `--as-population` option. Also, is it due to the stats being kept in memory, or the morphologies (Neurn objects), or both?

jdcourcol:
@juanchopanza : It is true that I cannot assert that the memory was consumed by the statistics and not the morphologies.
 
> Careful with the --as-population option.
Can you elaborate on that ?

juanchopanza:
It collects summary stats for a whole population, not individual morpholigies. So you need to be clever about how that is done w/o keeping the statistics in memory.

mgeplf:
I would probably do that separately, but it's possible to do online aggregation instead of in batch, I think.  I will look at the code.

asanin-epfl:
Would need a total refactoring of `morph_stats`. With the current design it is impossible to patch.

arnaudon:
One can do it in batch upon calling the extract_dataframe method I think, instead of bruteforcing many morphs into it. It will in any case take for ever to compute many stats on such a large number of morphologies. For such usecase, I would use a distributed parallel computation calling only the neurom.get functions one is interested in. This morphs_stat is designed to be use on a a reasonable number of morphologies.

PR #922, opened 2021-05-25
Don't force loading of neurons into memory

asanin-epfl:
- Introduce the cache flag for `Population` and `load_neurons`
- move loading of neurons inside `Population`

Solution to #538

### DIFF
--- a/neurom/core/population.py
+++ b/neurom/core/population.py
@@ -50,8 +50,10 @@ class Population:
 
         Arguments:
             files (collections.abc.Sequence[str|Path|Neuron]): collection of neuron files or
-              paths to them.
-            name (str): Optional name for this Population.
+              paths to them
+            name (str): Optional name for this Population
+            ignored_exceptions (tuple): NeuroM and MorphIO exceptions that you want to ignore when
+              loading neurons.
         """
         self._files = files
         self._ignored_exceptions = ignored_exceptions
--- a/neurom/core/population.py
+++ b/neurom/core/population.py
@@ -61,9 +61,7 @@ class Population:
         self._ignored_exceptions = ignored_exceptions
         self.name = name
         if cache:
-            self._files = []
-            for f in files:
-                self._files.append(self._load_file(f))
+            self._files = [self._load_file(f) for f in files if f is not None]
         else:
             self._files = files
 
--- a/neurom/core/population.py
+++ b/neurom/core/population.py
@@ -27,42 +27,82 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 """Neuron Population Classes and Functions."""
+import logging
+from pathlib import Path
 
-from itertools import chain
+import neurom
+from morphio import MorphioError
+from neurom.exceptions import NeuroMError
+
+
+L = logging.getLogger(__name__)
 
 
 class Population:
     """Neuron Population Class.
 
-    Features:
-        - flattened collection of neurites.
-        - collection of somas, neurons.
-        - iterable-like iteration over neurons.
+    Offers an iterator over neurons within population, neurites of neurons, somas of neurons.
+    It does not store the loaded neuron in memory unless the neuron has been already passed
+    as loaded (instance of ``Neuron``).
     """
-    def __init__(self, neurons, name='Population'):
+    def __init__(self, files, name='Population', ignored_exceptions=()):
         """Construct a neuron population.
 
         Arguments:
-            neurons: iterable of neuron objects.
-            name: Optional name for this Population.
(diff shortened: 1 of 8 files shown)


asanin-epfl:
@tomdele or @mgeplf let me know if you want to review this one

neurom/core/population.py
```diff
@@ -27,42 +27,83 @@
...
+        return None
+
     def __iter__(self):
         """Iterator to populations's neurons."""
-        return iter(self.neurons)
+        for f in self._files:
+            nrn = self._load_file(f)
+            if nrn is None:
+                continue
+            yield nrn
```
adrien-berchet, line 94: Souldn't we use a cache mechanism here to avoid loading the neurons each time we request one? If I am correct, if we call `pop.neurons` then `pop.neurites`, all the files are loaded twice, which might be quite heavy.

adrien-berchet (commented):
I like the idea but if I understand properly, it is no more possible to just use a list of neurons to build a population which might be quite annoying IMO (I think we use this feature in `synthesis-workflow`).

asanin-epfl:
> ... it is no more possible to just use a list of neurons to build a population ...

It is possible to use a list of neurons.

neurom/core/population.py
asanin-epfl: The whole point of this PR is to avoid loading into memory, so it can handle huge amounts of data. Solution to your concern will be to use parallelization. Also it is still possible to keep everything in memory as `Population` accepts a list of loaded neurons as well.

adrien-berchet: I don't get your point on parallelization. Even if we parallelize the code to work on batches of neurons, each neuron file will still be loaded twice, isn't it? But if we want to avoid that it is indeed possible to load the neurons and then create the population from it, so I guess it's no big deal.

adrien-berchet:
> > ... it is no more possible to just use a list of neurons to build a population ...
> 
> It is possible to use a list of neurons.

Ah yes sorry, I did not read carefully the `_load_file` method since I expected it would just 'load' the file, but it actually checks if it is already a neuron object. Maybe this method could be renamed into something like `_get_neuron` or something to be more clear?

neurom/core/population.py
asanin-epfl: Well, as an alternative a flag `cache=True/False` can be introduced. When `True`, all loaded morphologies will be kept in memory.

asanin-epfl: > neuron file will still be loaded twice, isn't it?

Yes. A neuron will be loaded each time it is accessed within a separate iteration.

adrien-berchet: > Well, as an alternative a flag `cache=True/False` can be introduced. When `True`, all loaded morphologies will be kept in memory.

Indeed this would be very handy.

neurom/core/population.py
```diff
@@ -27,42 +27,92 @@
...
         """
-        self.neurons = tuple(neurons)
-        self.somata = tuple(neu.soma for neu in neurons)
-        self.neurites = tuple(chain.from_iterable(neu.neurites for neu in neurons))
+        self._ignored_exceptions = ignored_exceptions
         self.name = name
+        if cache:
+            self._files = []
+            for f in files:
+                self._files.append(self._load_file(f))
```
adrien-berchet, line 66: Is there a particular reason not to use list comprehension here?

adrien-berchet (commented):
Thanks for the cache!

arnaudon (commented):
Just out of curiosity, when is this an issue in term of number of neurons? Can it blow up the memory?

asanin-epfl:
> Just out of curiosity, when is this an issue in term of number of neurons? Can it blow up the memory?

See the linked issue in the first comment.

asanin-epfl:
> so is the issue the morphologies loaded, or the statistics in memory?

The issue is with morphologies being loaded into memory. Statistics don't take much memory. An average morphology I see is 2Mb. Now imagine 10000 of them into memory. It's 20Gb.

arnaudon:
Sure, but computing all the stats on 10k morphs on a computer with 20Gb of ram (aka a laptop) will burn it haha. Anyway, this looks good, I like it!

neurom/core/population.py
asanin-epfl: thanks, used it

asanin-epfl:
> Sure, but computing all the stats on 10k morphs on a computer with 20Gb of ram (aka a laptop) will burn it haha. Anyway, this looks good, I like it!

I see the places where this can be optimizied. I hope it will give boost in performance.

adrien-berchet (approved):

asanin-epfl:
Thank you!

Issue #804, opened 2020-02-03
tree_type_checker(*NEURITES) does not filter out soma


wizmer:
since NEURITES is a tuple that includes NeuriteType.all, it will never filter anything out.

Here is a modified version of the existing unit test to test for soma filtration.
```python
def test_tree_type_checker():
    tree_filter = tree_type_checker(NeuriteType.all)
    nt.ok_(tree_filter('fake_tree'))

    mock_tree = Mock()
    mock_tree.type = NeuriteType.axon

    tree_filter = tree_type_checker(*NEURITES)
    nt.ok_(tree_filter(mock_tree))

    mock_tree.type = NeuriteType.soma
    nt.ok_(not tree_filter(mock_tree))  # <- This assertion fails
```

asanin-epfl:
This is a topic for discussion rather. NeuroM followed the logic of NEURON simulation, and considered soma as section hence `NEURITES` were defined right in that paradigm. Now, since NeuroM moved to MorphIO, soma is not considered as section. Hence `.all` can't be part of `NEURITES`. I can drop it from there but I need confirmation from other interesting parties.

@arnaudon can you on behalf of scientists discuss the drop of `NeuriteType.all` from `NEURITES` in `neurom.core.types` module?

arnaudon:
I sent it around to discuss, I'll come back to you. I never use .all myself, but I've seen it in a few places. It may break some codes if we remove it.

wvangeit:
I guess it's a bit of semantic issue. I wouldn't consider the soma as a neurite. But that doesn't mean it's not a section though. 
So maybe the naming NEURITES was a bit unfortunate?

asanin-epfl:
> But that doesn't mean it's not a section though.

@wvangeit  In NeuroM v2 soma is not considered as section because of MorphIO.

lidakanari:
I agree that soma does not make sense as a neurite. However, we need to compute measurements, such as total length, for *all* neurites (including for example axon + dendrites). In addition, a lot of publications require the features for axons + dendrites.

asanin-epfl:
Removing of `.all` won't affect axon + dendrites. They stay. Only soma will be out of measurements. Currently when you use `NEURITES`, you get soma into your measurements as well.

lidakanari:
[...]

PR #931, opened 2021-06-10
remove NeuriteType.all from NEURITES

asanin-epfl:
Solution to #804


arnaudon (approved):
I don't use .all, so I'm good, but I'm not aware of all possible usecases.

asanin-epfl:
> I don't use .all, so I'm good, but I'm not aware of all possible usecases.

I understand. That's why I've added all concerning parties as reviewers here.

adrien-berchet (approved):
As far as I can see, only NeuroM is using `.all`, so we will just need a new version of NeuroM compatible with this PR.

arnaudon:
But this is NeuroM, stop working so late :P

adrien-berchet:
Indeed... 🤦😁
I don't know why I though I was on morph-tool :-)

asanin-epfl:
Thank you!

PR #933, opened 2021-06-11
Remove the cli command ``neurom features``

asanin-epfl:
Instead a proper documentation is provided on the list of available features.

### DIFF
--- a/neurom/features/neuronfunc.py
+++ b/neurom/features/neuronfunc.py
@@ -26,7 +26,19 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-"""Morphometrics functions for neurons or neuron populations."""
+"""Neuron features.
+
+Any public function from this namespace can be called via features mechanism on a neuron, a
+neuron population:
+
+>>> import neurom
+>>> from neurom import features
+>>> nrn = neurom.load_neuron('path/to/neuron')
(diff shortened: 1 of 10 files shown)


neurom/features/neuronfunc.py
```diff
@@ -209,24 +221,25 @@ def _sort_angle(p1, p2):
             for i, _ in enumerate(ordered_vectors)]
 
 
-def sholl_crossings(neurites, center, radii, neurite_filter=None):
+def _sholl_crossings(neurites, center, radii, neurite_type=NeuriteType.all):
```
asanin-epfl, line 224: @arnaudon I've made this function private. Is it ok for you? Did you use it?

arnaudon: ah could you leave it public, we could use it with specific argument for neurite type if recall, thanks!

neurom/features/neuronfunc.py
```diff
@@ -209,24 +221,25 @@ def _sort_angle(p1, p2):
...
         radii(iterable of floats): radii for which crossings will be counted
+        neurite_type(NeuriteType): Type of neurite to use. By default ``NeuriteType.all`` is used.
 
     Returns:
         Array of same length as radii, with a count of the number of crossings
         for the respective radius
+
+    This function can also be used with a list of sections, as follow::
+
+        secs = (sec for sec in nm.iter_sections(neuron) if complex_filter(sec))
```
arnaudon, line 239: @asanin-epfl , this use case in fact.

arnaudon (changes requested):
Just put back sholl to pubic, then all good!

arnaudon (approved):
Thank you!

asanin-epfl:
Thank you!

PR #934, opened 2021-06-11
Follow the same rules of morphology validation as MorphIO

asanin-epfl:
Merge only after https://github.com/BlueBrain/MorphIO/pull/327 is merged.


arnaudon (approved):

asanin-epfl:
Thank you!

Issue #593, opened 2017-02-15
Refactor plotting/viewing code


mgeplf:
The original comment by Mike:

A matplotlib upgrade broke the tests, so to get around that for the moment, matplotlib has been pinned to a working version: #592 

The plotting/viewing tests should be more robust against this, and should more carefully check that the output is expected, using a process similar to how matplotlib does its tests.

Once this is done, #592 should be rolled back.

-----------------
Update by asanin-epfl:

Better tests for such tangled code would be hard to implement. I propose to minimize the code first:
- Drop entirely `neurom.viewer`. This module with its dynamic loaders rather confuses and over-complicates the code. Users should directly call the required function like `plot_dendrogram`.
- See what else can be minimized/refactored in `neurom.view` module.
- Improve tests

PR #935, opened 2021-06-15
refactor viewer

asanin-epfl:
Solution to #593

- delete `neurom.viewer`
- rename `neurom.view.view` to `neurom.view.matplotlib_impl`
- rename `neurom.view.plotly` to `neurom.view.plotly_impl`
- rename `neurom.view.common` to `neurom.view.matplotlib_utils`
- swap arguments `ax` and `nrn` of all plot functions in `neurom.view.matplotlib_impl`


arnaudon:
Ah, I'm using this a lot, I hope it's not breaking to much! I'll take a proper look later.

adrien-berchet:
That's a quite big breaking change for us indeed. Why do you want to change this?

asanin-epfl:
Because `neurom.viewer` is an abstraction layer that complicates instead of simplifying:
- It introduces an additional layer of abstraction with magical `**kwargs` that are not documented, and can't be documented because it dynamically calls to different functions with different signatures. So, user has to go in the end to the underlying called function for documentation. This action completely kills the purpose of abstraction.
- it does not offer `plotly` option despite being an abstraction layer that tries to offer simpler API.

adrien-berchet:
I see and I completely agree.
Nevertheless, we had a lot of breaking changes in NeuroM recently. It would be nice to gather them in one big new release when they are not urgent (this one does not seem urgent?). Or maybe you could add the new API that you propose but also keep the current API and raise a warning for future deprecation?

arnaudon:
+1 for keeping both and put a deprecation warning, so we can have a smooth transition. I don't have time to look at this now, and I guess we could keep it this way and iterate over it on a few PRs/minor release to have a strong plotting module.

arnaudon:
Remember, there is also this code: https://github.com/BlueBrain/plotly-helper

asanin-epfl:
> Or maybe you could add the new API that you propose but also keep the current API and raise a warning for future deprecation?

I can keep `neurom.viewer` but other things get renamed. Will it be ok for you? Otherwise this PR can be put on hold for now.

adrien-berchet:
As far as I can see it would only break NeuroR and synthesis-workflow (just 1 import to update), so I think it would be a good compromise for us if it's ok for you.

jdcourcol:
The documentation contains examples with viewer. They have to be updated too.
Also the mybinder notebook has to be updated too.

asanin-epfl:
@adrien-berchet, the merge of this branch will go into `2.4.0` branch instead of `master`. I will be collecting all further changes in the branch `2.4.0` as well. In a couple of weeks, I will merge `2.4.0` into `master`.

@jdcourcol thank you. I've updated notebooks and examples.

adrien-berchet (approved):
> @adrien-berchet, the merge of this branch will go into `2.4.0` branch instead of `master`. I will be collecting all further changes in the branch `2.4.0` as well. In a couple of weeks, I will merge `2.4.0` into `master`.

Looks like a very good plan, thank you very much!

PR #936, opened 2021-06-17
rename ``Population.somata`` to ``Population.soma``

asanin-epfl:
to have name consistency among ``Population`` and ``Neuron``.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+Version 2.4.0
+-------------
+- rename ``Population.somata`` to ``Population.soma`` to have name consistency among ``Population``
+  and ``Neuron``.
+
 Version 2.3.0
 -------------
 - Introduce a new method to calculate partition asymmetry by Uylings. See docstring of
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,8 +3,8 @@ Changelog
 
 Version 2.4.0
 -------------
-- rename ``Population.somata`` to ``Population.soma`` to have name consistency among ``Population``
-  and ``Neuron``.
+- deprecate ``Population.somata`` in favor of ``Population.soma`` to have name consistency among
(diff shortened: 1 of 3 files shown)


adrien-berchet:
As for #935 , could we have a first version with the two properties and a deprecation warning please?
Also, could we group all these breaking changes in the same version? So we will have only one big update to do when it will be released.

adrien-berchet (approved):
Thanks!

eleftherioszisis:
May I ask why these two fundamentally different entities need to have common signatures? Population.soma is a confusing combination, isn't it?

asanin-epfl:
I saw that Population and Neuron has the same `neurites` property for features mechanism. So, I thought to make the similar approach for `soma`. Anyway, I can revert.
Can you clarify that `somata` aimed to empasize plural somas at Population? Should we rather call it `somas` then? Is `somata` an official name for plural somas?

eleftherioszisis:
`neurites` property is the same for both `Population` and `Neuron` because even a single neuron has a collection of neurites. However, a Neuron has only one soma, while the Population one soma per Neuron. This is why I find the combination of singular soma and plural Population misleading.

And yeah "soma" is a greek word and its plural is indeed "somata", not "somas". ^^

asanin-epfl:
good then, I'm reverting

eleftherioszisis:
I apologize for not noticing this PR earlier.. :(

asanin-epfl:
No worries. From now on I will put you as a reviewer. Thanks for clarifying.

adrien-berchet:
> And yeah "soma" is a greek word and its plural is indeed "somata", not "somas". ^^

Ah I did not know that! Thanks for the greek course :)

asanin-epfl:
> Ah I did not know that! Thanks for the greek course :)

That caught me as well. I always thought `somata` is more like official name for `soma` e.g. more biological, including all organelles and stuff.

PR #939, opened 2021-06-17
Refactor features.

asanin-epfl:
- Move ``neuritefunc`` functions that expect neurons to ``neuronfunc``. The functions are:
``max_radial_distances, number_of_sections, number_of_neurites, number_of_bifurcations, number_of_forking_points, number_of_terminations, number_of_segments``
- Make ``neuronfunc`` to work with list of neurons besides a neuron and a neuron population.
- Name consistency among private variables.

### DIFF
--- a/neurom/features/neuritefunc.py
+++ b/neurom/features/neuritefunc.py
@@ -49,7 +49,6 @@ from neurom.core.neuron import NeuriteType, Section, iter_neurites, iter_section
 from neurom.core.dataformat import COLS
 from neurom.core.types import tree_type_checker as is_type
 from neurom.features import _register_feature, bifurcationfunc, feature, sectionfunc
-from neurom.features.neuronfunc import _neuron_population
 from neurom.geom import convex_hull
 from neurom.morphmath import interval_lengths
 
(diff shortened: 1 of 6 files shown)


neurom/features/neuritefunc.py
```diff
@@ -179,57 +178,6 @@ def pl2(node):
     return _map_sections(pl2, neurites, neurite_type=neurite_type)
 
 
-################################################################################
-# Features returning one value per NEURON                                      #
-################################################################################
-
-def _map_neurons(fun, neurites, neurite_type):
```
asanin-epfl, line 186: @mgeplf FYI. I saw that you added this code in 2016, so just for safety I'm notifying you.

mgeplf: Thanks for the heads up.

Before than they were just `partial` wrapped functions; this way there could be documentation added to them that allowed for better docs/cli info, iirc. As long as the behavior is the same or better, I have no problems w/ the refactorings.

adrien-berchet (approved):
I think it's ok for us since we never call these features directly (as far I as know) but only with the `get()` function.
And it's cool to be able to use a list of Neurons!

arnaudon (approved):
moving of features from neurite to neuron would make a difference in the feature extraction part, as it classifies them according to neurite_types or neuron irrc, not sure of the implications, but it may break a few tests based on that

asanin-epfl:
> moving of features from neurite to neuron would make a difference in the feature extraction part, as it classifies them according to neurite_types or neuron irrc

Those feature functions expect neurons as input, so it is misleading to keep them in `neuritefunc`.

arnaudon:
yes, I agree with your change, I was just thinking of possible things that could break because of that, but it should be minor indeed, will you merge that in 2.4.0 branch as well?

asanin-epfl:
> will you merge that in 2.4.0 branch as well?

yes, you can see the merging branch at the top, right below the PR's title.

asanin-epfl:
2.4.0 will be released in a couple of weeks from now, no worries.

PR #942, opened 2021-06-21
Stats with args

asanin-epfl:
New Config format for morph_stats that allows to set features args, and drop `partition_asymmetry_length`.


arnaudon (approved):
I like that, thanks! We'll do a major test on our codes before the release of 2.4.0 to try all these new features, thanks!

asanin-epfl:
Thank you

Issue #918, opened 2021-05-12
NeuroM 3.0


asanin-epfl:
NeuroM seriously confuses concepts of Neuron and Morphology. Everything that currently has a name `neuron` or `Neuron` must be changed to `morphology` and `Morphology`. This way NeuroM can be clearly used for any topological entity like: Neuron, Glial cell, Vasculature cell, etc.

I suggest a huge renaming of `neuron` to `morphology` and release NeuroM 3.0 with this change.

eleftherioszisis:
Although it is a good idea to generalize the morphological entity, we need to keep in mind that each family of cells may have their own unique features that depend on extra datasets (e.g. perimeters in astrocytes). 

As far as the vasculature is concerned, I believe it is out of the scope of this package because it is a cyclic graph without a specific root, therefore it is quite different and its features may required specialized mechanics.

asanin-epfl:
Thank you so much for feedback @eleftherioszisis.

Unique features are not the problem. NeuroM still can have subclasses for Neuron, Astrocyte, etc. The head of these classes will be `Morphology`.

- Is the name `Morphology` ok for you? What other alternatives can you suggest?

- If NeuroM at some point decides to generalize also vasculature entities, what name can you suggest for such general entity instead of `Morphology`? `Graph`? `Topology`?

eleftherioszisis:
`Morphology` is a good generic name for any sort of entity that can be represented as cross-sections (points + diameters) linked together by segments. Therefore, neurons, glia, spines, and vasculature are all morphologies.

IMHO `Graph` does not imply an embedding in the euclidean space. A social network for instance can be a graph, but it does not fit the criteria for being a Morphology.

asanin-epfl:
Thank you for clarifying.

arnaudon:
This is going to break again several tools ;) But +1 for being consistent in naming these.

PR #947, opened 2021-06-23
2.4.0

asanin-epfl:
This is a PR for version 2.4.0. It includes the PRs:
- refactor viewer (#935)
- Refactor features. (#939)
- Delete deprecated neurom.features.register_neurite_feature (#940)
- Stats with args (#942) 

For details see the changes of `CHANGELOG.rst`.

I might include another PR here but anyway it is ready for your detailed review, and I'm open to your questions. Don't rush with the review. I plan to release this after the 5th of July.


adrien-berchet:
Thanks @asanin-epfl !
I am just wondering if such release with several breaking changes should not be a major release, so v3 instead of v2.4? Because in several codes we have the requirement as `neurom>=2,<3` so they would not break until we update them to work with this new release. WDYT?

asanin-epfl:
>  so v3 instead of v2.4? 

One thing that blocks me from that is #918 . I plan to make the very breaking change and rename all `neuron` names to `morph` names. I plan to use 3.0 for that. 2.4.0 should work because only plotting changes are breaking (I hope not much people use plotting). Other things only get deprecated.
Also I don't want to merge all changes including #918 in one big 3.0. Because I want to delete deprecated stuff in 3.0. There will be much of it.

adrien-berchet:
Ok I see
But does anybody needs these changes before what you want to introduce in v3? If not, why do you mind making a bigger v3 in which deprecated stuffs are removed and all these breaking changes are grouped?

asanin-epfl:
>  If not, why do you mind making a bigger v3 in which deprecated stuffs are removed and all these breaking changes are grouped?

If you are fine with it then I can make it. A huge v3 with all deprecated stuff removed and `neuron` renamed to `morphology`. Please confirm once more that you are ok with that.

adrien-berchet:
I think it would be easier for me to have only one big change to do, what do you think @arnaudon and @eleftherioszisis ?

arnaudon:
+1 for a big release, so we are safe on some codes with <3 m thank you!

eleftherioszisis:
I agree that we should not have breaking changes in a minor version release.

PR #956, opened 2021-08-27
Add method to Soma class to check wether it overlaps a point or not

adrien-berchet:


neurom/core/soma.py
```diff
@@ -149,6 +158,29 @@ def __str__(self):
...
                 (repr(self.points), self.center, self.radius))
 
+    def overlaps(self, points, exclude_boundary=False):
+        """Check that the given points are located inside the soma."""
+        points = np.atleast_2d(np.asarray(points, dtype=np.float64))
+        mask = np.ones(len(points)).astype(bool)
+        for p1, p2 in zip(self.points[:-1], self.points[1:]):
+            vec = p2[COLS.XYZ] - p1[COLS.XYZ]
+            vec_norm = np.linalg.norm(vec)
+            dot = np.einsum("i,ji", vec, points[mask] - p1[COLS.XYZ]) / vec_norm
```
arnaudon, line 168: isnt' this faster? (not sure) 

```(points[mask] - p1[COLS.XYZ]).dot(vec)```

File: neurom/core/soma.py [hunk 2]
```diff
@@ -149,6 +158,29 @@ def __str__(self):
         return ('SomaCylinders(%s) <center: %s, virtual radius: %s>' %
                 (repr(self.points), self.center, self.radius))
 
+    def overlaps(self, points, exclude_boundary=False):
```
arnaudon, line 161: hard to see what' going one quickly here, is it like checking if a point is in convex hull or the soma.points, or something like that?

arnaudon (approved):

asanin-epfl (commented):
I'm waiting for CHANGELOG, and resolved questions by Alexis.

neurom/core/soma.py
adrien-berchet: Indeed, and much more readable :)

neurom/core/soma.py
adrien-berchet: In this case each segment represents a cylinder, so we check if each point is inside at least one cylinder.

neurom/core/soma.py
arnaudon: not sure exactly what you did, but I just did this cylinder check a few days ago, you want to do this:
if cylinder is parametrized by pos_vec + x * dir_vec (2 vectors and parameter x) for the centerline and radius r, the distance from a point p to centerline of cylinder is `np.linalg.norm(np.cross(pos_vec-p, dir_vec))`, which you can just check against the radius. Maybe this is what you did ;)

neurom/core/soma.py
arnaudon: not sure exactly what you did, but I just did this cylinder check a few days ago, you want to do this:
if cylinder is parametrized by pos_vec + x * dir_vec (2 vectors and parameter x) for the centerline and radius r, the distance from a point p to centerline of cylinder is np.linalg.norm(np.cross(pos_vec-p, dir_vec)), which you can just check against the radius. Maybe this is what you did ;)

adrien-berchet: Yes, that basically what I did. I compute the longitudinal and orthogonal components and check if the orthogonal is less than radius and if the longitudinal is within the range of the cylinder axis.

adrien-berchet:
I added the case for the contour soma but I am not sure I understand how it works properly... As far as I could see in the doc, the contour should be in the XY plane and the Z is just ignored?

asanin-epfl:
> As far as I could see in the doc, the contour should be in the XY plane and the Z is just ignored?

Where this is mentioned? I don't know about this soma hierarchy. From what I see in the source code, Z is not ignored.

asanin-epfl (approved):

PR #964, opened 2021-09-14
fix raw option in morph_stat

arnaudon:
When `raw` option is used, it crashed because `np.raw` does not exist.


neurom/apps/morph_stats.py
```diff
@@ -131,8 +131,10 @@ def _get_feature_stats(feature_name, morphs, modes, kwargs):
 
         stat = value
         if isinstance(value, Sized):
-            if len(value) == 0 and mode not in {'raw', 'sum'}:
+            if len(value) == 0:
                 stat = None
+            elif mode == 'raw':
+                stat = value
```
adrien-berchet, line 137: Before NeuroMv3 it was `if mode == raw: return values.tolist()`, so maybe we should have `stat = value.to_list()` here to be consistent?
@asanin-epfl do you remember why you changed this in #950 ?

arnaudon: ah yes, I can add .tolist() indeed.

arnaudon:
I'm adding the raw key in test for coverage

arnaudon:
@asanin-epfl , why not .tolist?

asanin-epfl:
@arnaudon and why .tolist ?

neurom/apps/morph_stats.py
asanin-epfl: .tolist is not needed, `value` is a list already

arnaudon: ah ok!

adrien-berchet: Ah ok, it was changed elsewhere. Sorry :)

adrien-berchet (approved):

asanin-epfl:
We are waiting for @arnaudon coverage tests on 'raw' mode.

arnaudon:
here you are

neurom/apps/morph_stats.py
asanin-epfl: No, I changed I think. I was just bluntly optimizing. I fixed docstrings so now it's clear that feature values can't be numpy arrays. They are either a single number or a list of numbers. I didn't like joggling of values between numpy arrays and lists. So I choose one format for consistency. The simpler one. If user want he can always convert list to a numpy array.

tests/apps/test_morph_stats.py
```diff
@@ -68,14 +68,15 @@
...
-        'max_radial_distance': {'modes': ['mean']},
+        'max_radial_distance': {'modes': ['mean', 'raw']},
     }
 }
 
 REF_OUT = {
     'morphology': {
         'mean_soma_radius': 0.13065629648763766,
         'mean_max_radial_distance': 99.5894610648815,
+        'raw_max_radial_distance': 99.5894610648815,
```
asanin-epfl, line 79: so, `raw_max_radial_distance` and `mean_max_radial_distance` are the same? is it ok?

neurom/apps/morph_stats.py
adrien-berchet: Ok, it's simpler like this indeed. Thanks for the clarification.

tests/apps/test_morph_stats.py
arnaudon: there is just one number here

arnaudon:
I don't know how to fix coverage...

tests/apps/test_morph_stats.py
arnaudon: we can put raw in another entry if we want, but it may yield many numbers

asanin-epfl:
ok, I will fix the coverage

arnaudon:
Thank you!

tests/apps/test_morph_stats.py
```diff
@@ -233,12 +241,16 @@ def test_extract_dataframe_multiproc():
             for name in ['Neuron.swc', 'simple.swc']]
     with warnings.catch_warnings(record=True) as w:
         actual = ms.extract_dataframe(morphs, REF_CONFIG, n_workers=2)
+        # drop raw features as they require too much test data to mock
```
asanin-epfl, line 244: @adrien-berchet please confirm whether you are ok with such approach. I use in in 4 places here.

arnaudon: yes, it's fine, we don't care about the numbers, this is tested elsewhere

adrien-berchet: It's ok for me too

asanin-epfl:
do you need a release?x

arnaudon:
no, it's fine! I just noticed when I was trying something locally, don't bother!

PR #969, opened 2021-09-27
fix plot limits

arnaudon:
Automatic limit for matplotlib were done with soma size only, and do not work with cli.


arnaudon:
aaaah

arnaudon:
I give up...

adrien-berchet:
:rofl:

arnaudon:
solved in another PR

Issue #975, opened 2021-11-25
Axon metrics at zero for some files


jonathanlurie:
Hello,

I am using NeuroM to compute some morph metrics on the soma, axon, basal dendrites and apical dendrite. In some cases, the axon metrics are all zeros even though there is an axon. From what we could gather, it seems that this issue is occuring in about 25% of cases among the ~200 morphs we processed. Note that things are working well for basal dendrite, apical dendrite (when present) and general morphology metrics.

Here is a short version of the code we use to compute the metrics:

```python
METRIC_CONFIG = {
    'neurite': {
        'max_radial_distance': ['mean', 'std'],
        'number_of_sections': ['min', 'max', 'median', 'mean', 'std'],
        'number_of_bifurcations': ['min', 'max', 'median', 'mean', 'std'],
        'number_of_leaves': ['min', 'max', 'median', 'mean', 'std'],
        'total_length': ['min', 'max', 'median', 'mean', 'std'],
        'total_area': ['min', 'max', 'median', 'mean', 'std'],
        'total_volume': ['min', 'max', 'median', 'mean', 'std'],
        'section_lengths': ['min', 'max', 'median', 'mean', 'std'],
        'section_term_lengths': ['mean', 'std'],
        'section_bif_lengths': ['mean', 'std'],
        'section_branch_orders': ['min', 'max', 'mean', 'std'],
        'section_bif_branch_orders': ['mean', 'std'],
        'section_term_branch_orders': ['mean', 'std'],
        'section_path_distances': ['min', 'max', 'median', 'mean', 'std'],
        'section_taper_rates': ['min', 'max', 'median', 'mean', 'std'],
        'local_bifurcation_angles': ['min', 'max', 'median', 'mean', 'std'],
        'remote_bifurcation_angles': ['min', 'max', 'median', 'mean', 'std'],
        'partition_asymmetry': ['mean', 'std'],
        'partition_asymmetry_length': ['mean', 'std'],
        'sibling_ratios': ['min', 'max', 'median', 'mean', 'std'],
        'diameter_power_relations': ['min', 'max', 'median', 'mean', 'std'],
        'section_radial_distances': ['min', 'max', 'median', 'mean', 'std'],
        'section_term_radial_distances': ['mean', 'std'],
        'section_bif_radial_distances': ['mean', 'std'],
        'terminal_path_lengths': ['mean', 'std'],
        'section_volumes': ['min', 'max', 'median', 'mean', 'std'],
[...]

PR #977, opened 2021-12-08
allow for neurites to be attached to neurites

arnaudon:
Proposal for https://github.com/BlueBrain/NeuroM/issues/975

### DIFF
--- a/neurom/core/morphology.py
+++ b/neurom/core/morphology.py
@@ -438,20 +438,17 @@ class Morphology(morphio.mut.Morphology):
         self.name = name if name else 'Morphology'
         self.morphio_soma = super().soma
         self.neurom_soma = make_soma(self.morphio_soma)
-        self.initial_sections = self.get_initial_sections(only_root=only_root)
+        self.only_root = only_root  # set True for legacy
+        self.initial_sections = self.get_initial_sections()
 
     @property
     def soma(self):
         """Corresponding soma."""
         return self.neurom_soma
 
-    def get_initial_sections(self, only_root=False):
-        """Get sections to initiate neurites.
-
-        Args:
-            only_root (bool): is True, only considers root sections (legacy)
-        """
-        if only_root:
+    def get_initial_sections(self):
+        """Get sections to initiate neurites."""
+        if self.only_root:
             return self.root_sections
         initial_sections = []
         for root_section in self.root_sections:
@@ -463,6 +460,10 @@ class Morphology(morphio.mut.Morphology):
                     prev_type = section.type
         return initial_sections
 
+    def append_root_section(self, *args):
+        """Update initial sections if we append new root sections."""
+        super().append_root_section(*args)
+        self.initial_sections = self.get_initial_sections()
 
     @property
     def neurites(self):
--- a/neurom/core/morphology.py
+++ b/neurom/core/morphology.py
@@ -433,12 +433,13 @@ class Morphology(morphio.mut.Morphology):
         Args:
             filename (str|Path): a filename
             name (str): a option morphology name
+            only_root (bool): set to True for not considering neurites on neurites
         """
         super().__init__(filename)
         self.name = name if name else 'Morphology'
         self.morphio_soma = super().soma
         self.neurom_soma = make_soma(self.morphio_soma)
-        self.only_root = only_root  # set True for legacy
+        self.only_root = only_root
         self.initial_sections = self.get_initial_sections()
 
     @property
--- a/neurom/core/morphology.py
+++ b/neurom/core/morphology.py
@@ -45,9 +45,10 @@ from neurom.utils import warn_deprecated
 class Section:
     """Simple recursive tree class."""
 
-    def __init__(self, morphio_section):
+    def __init__(self, morphio_section, only_root=False):
         """The section constructor."""
         self.morphio_section = morphio_section
+        self.only_root = only_root  # for legacy
 
     @property
     def id(self):
@@ -58,6 +59,9 @@ class Section:
     def parent(self):
         """Returns the parent section if non root section else None."""
         if self.morphio_section.is_root:
+            return None
+
+        if not self.only_root and self.morphio_section.parent.type != self.type:
             return None
         return Section(self.morphio_section.parent)
 
@@ -423,7 +427,7 @@ class Neurite:
 class Morphology(morphio.mut.Morphology):
     """Class representing a simple morphology."""
 
-    def __init__(self, filename, name=None):
+    def __init__(self, filename, name=None, only_root=False):
         """Morphology constructor.
 
         Args:
@@ -434,16 +438,36 @@ class Morphology(morphio.mut.Morphology):
         self.name = name if name else 'Morphology'
         self.morphio_soma = super().soma
         self.neurom_soma = make_soma(self.morphio_soma)
+        self.initial_sections = self.get_initial_sections(only_root=only_root)
 
     @property
     def soma(self):
         """Corresponding soma."""
         return self.neurom_soma
 
+    def get_initial_sections(self, only_root=False):
+        """Get sections to initiate neurites.
+
+        Args:
+            only_root (bool): is True, only considers root sections (legacy)
+        """
+        if only_root:
+            return self.root_sections
+        initial_sections = []
+        for root_section in self.root_sections:
+            initial_sections.append(root_section)
+            prev_type = root_section.type
+            for section in root_section.iter():
+                if section.type != prev_type:
+                    initial_sections.append(section)
+                    prev_type = section.type
+        return initial_sections
+
+
     @property
     def neurites(self):
         """The list of neurites."""
-        return [Neurite(root_section) for root_section in self.root_sections]
+        return [Neurite(root_section) for root_section in self.initial_sections]
 
     @property
     def sections(self):


arnaudon:
I'm not sure why trunk angle stuff is breaking... I'm looking

eleftherioszisis:
The idea that is presented here of traversing the trees to specify the heterogeneous types is in the right direction. However, our goal is towards a solution that doesn't mutate the Morphology or add extra properties just to serve a very specific use case.

NeuroM Morphology should be seen as a "view" of the MorphIO Morphology, therefore it should be a very thin layer on top of it, adapting the data so that NeuroM can consume it without changing it significantly.

Your approach makes perfect sense, however, I believe it is implemented in the wrong place. Instead, it should be during the feature calculation, where the careful handling of the subtrees is performed without changing the Morphology.

mgeplf:
I agree w/ @eleftherioszisis; I wonder if adding something to the `Section` in `morphio` would help: `is_heterogenous(bool towards_root=false)`.  Thus, `s.is_heterogenous()` would say if any downstream sections are of a different type than the current one, and `s.is_heterogenous(true)` would check the same, but towards the root.

Then, we can augment NeuroM traversals to check this property, and depending on the behavior we want (from a scientific point of view) the correct behavior is performed.

To me, the code side of things is different from an agreement on *what* is required from a scientific point of view - once that is decided, we can hash out the details of the implementation.

arnaudon:
Imo it is still a view of MorphIO, and in fact, the current implementation is wrong. 
A neurite is defined as: 
```Neurite: collection of consecutive sections of the same SectionType . ```

If SectionType changes along a branch in MorphIO, this definition is not correct anymore (thus this https://github.com/BlueBrain/NeuroR/blob/master/neuror/sanitize.py#L36). This PR provides a correct implementation of a Neurite object, as a particular view of what MorphIO represents.

mgeplf:
@arnaudon: `Neurite: collection of consecutive sections of the same SectionType .`

Yeah, that's the definition we need relax, as @lidakanari mentioned in https://github.com/BlueBrain/NeuroM/issues/975#issuecomment-982713266:

> This is a real effect, it cannot be considered as an error. In some rodent brain regions this can be as frequent as 70% of the data! So imo this should not be removed / sanitized.

arnaudon:
So this is the correct definition, right? What do you mean relax? From what def to what def?

mgeplf:
*ahhhh*; I think I see where the confusion is coming in;  I see a `neurite` is as something that grows off the soma of a cell.  Thus, a neurite can start as a dendrite, coming off the soma, and then bifurcate into subtrees that have different types: one that continues as a dendrite, and part that is an axon.

Thus, the neurite is heterogenous, and the relaxation is that they are no longer all the same type.

Again, I'm not a scientist, just trying to understand.

eleftherioszisis:
As @mgeplf says we will need to bring the `Neurite` closer to what it represents in biology. That means we should allow different section types per neurite to cover the cases where the axon starts from the basal dendrite. We had the convention that a neurite has a single type, but as it is apparent it is not a correct one as it doesn't cover all cases in biology.

arnaudon:
So the real question is: can we call an axon emerging from a basal, a Neurite. If yes, my implementation is valid, if no, your heterogeneous proposal is valid. Let me enquire.

arnaudon:
To me: 
a) yes, if you are interested in subtypes, such as trunk/obliques, etc... not if you are considering 'main' types axon/basal/apical.
b) according to standard definition yes, but these ABD are not standard at all (let see what Jean-marc says, he is one expert on these types of neurons). A subtree has a pretty clear and unique definition, no? All downward sections from a starting section, downward defined as opposite direction to the root node (cell body).

mgeplf:
Clearing up `a)` about the main types is important, so that should be the question posed.  Also, I'm leery of  mixing top level types (`axon`/`dendrite`) w/ the concept of `trunk`/`oblique`.  The later are properties of sections, and there is ambiguity with them as different ppl have different definitions.

as for `b`: `but these ABD are not standard at all`, I interpret ` In some rodent brain regions this can be as frequent as 70%` from @lidakanari as suggesting this is quite common. Perhaps not the ABD, but the heterogeneous neurites (ie: tree rooted at the soma composed of `dendrite` and `axon`

lidakanari:
> Clearing up a) about the main types is important, so that should be the question posed. Also, I'm leery of mixing top level types (axon/dendrite) w/ the concept of trunk/oblique. The later are properties of sections, and there is ambiguity with them as different ppl have different definitions.
 
I totally agree with this. 

> as for b: but these ABD are not standard at all, I interpret In some rodent brain regions this can be as frequent as 70% from @lidakanari as suggesting this is quite common. Perhaps not the ABD, but the heterogeneous neurites (ie: tree rooted at the soma composed of dendrite and axon

Indeed, this is quite common depending on species, brain region. They are actually called ACD, Axon-Carrying Dendrites. Here is a small reference:
https://www.sciencedirect.com/science/article/pii/S0896627314006886

In addition, @wvangeit Mickael Zbili can comment on this property with respect to functional implication.

arnaudon:
Yes, here is another ref: https://www.frontiersin.org/articles/10.3389/fncel.2019.00570/full  in 'misbheaving neurons'

Mickael told me he does not know if we can call such axon a Neurite, and does not care, as long as it can be differentiated from the other neurites.

arnaudon:
Figure 4 of the paper cited by lida is good in fact, to know how to model these guys.

mgeplf:
Agreed that Fig 4 is key:
image

The question to me is; how many `neurites` are represented by the red and blue regions?
What do you think, @lidakanari, @wvangeit?

mzbili:
> > Clearing up a) about the main types is important, so that should be the question posed. Also, I'm leery of mixing top level types (axon/dendrite) w/ the concept of trunk/oblique. The later are properties of sections, and there is ambiguity with them as different ppl have different definitions.
> 
> I totally agree with this.
> 
> > as for b: but these ABD are not standard at all, I interpret In some rodent brain regions this can be as frequent as 70% from @lidakanari as suggesting this is quite common. Perhaps not the ABD, but the heterogeneous neurites (ie: tree rooted at the soma composed of dendrite and axon
> 
> Indeed, this is quite common depending on species, brain region. They are actually called ACD, Axon-Carrying Dendrites. Here is a small reference: https://www.sciencedirect.com/science/article/pii/S0896627314006886
> 
> In addition, @wvangeit, @mzbili can comment on this property with respect to functional implications.

Hi, I am not an expert on ABD but I know it is a growing field. If I understand well, the major functional impact to take into account ABD in the models in the fact the the ABD present different electrophysiological properties than non-ABD. To be clear, it means that to model ABD one has to differenciates dendrite and axon but also to differenciates dendrite that carries axon from all the other dendrites (and have different ephys properties in the dendrite that carries axon than in other dendrites). I am not sure that is the kind of precision the models from BBP wants to have...

eleftherioszisis:
If the `AcD dendrite` is fundamentally different than a `basal_dendrite`, then we could introduce a new section type `axon_carrying_dendrite` in MorphIO. NeuroM would deduce the respective neurite type as `axon_carrying_dendrite`, in which case we could easily target these neurites to perform separate morphometric analysis to their subtrees.

lidakanari:
I agree that these use cases should be handled different from regular dendrites and axons (so the `is_heterogenous` proposal is certainly of interest). 

So we have two points that need to be taken into account:
1. What is the impact on the axon (we already have some info on this from above reference). If axon starts from dendrite we need to consider a different AIS and the initial diameter is not the one we see on the soma surface! 
2. What is the impact on the dendrite (because we fully simulate the dendrites but only AIS this is even more important for simulations). We cannot compute morphometrics including the axon, so this part needs to be a "different" neurite. I assume the initial / trunk diameter that should be considered in the one on the soma and not the one on the first bifurcation. 

So to reply to @mgeplf we need to have two neurites: one axon (blue) starting from the dendrite, one dendrite (red) starting from the soma but excluding the axon. These are just suggestions from my side, I think the impact on morphometrics will be important but the impact on electrical models might be even more significant so let's see what they think.

mzbili:
> If the `AcD dendrite` is fundamentally different than a `basal_dendrite`, then we could introduce a new section type `axon_carrying_dendrite` in MorphIO (the axon subtree sections would still be of section type `axon`). NeuroM would deduce the respective neurite type as `axon_carrying_dendrite`, in which case we could easily target these neurites to perform separate morphometric analysis to their subtrees.

Hi, I think that scientifically it could be really interesting. In fact, some people have shown that AcD might present a bigger number of branches than non-AcD. Again, it depends on the precision we want for the models

arnaudon:
> So to reply to @mgeplf we need to have two neurites: one axon (blue) starting from the dendrite, one dendrite (red) starting from the soma but excluding the axon. These are just suggestions from my side, I think the impact on morphometrics will be important but the impact on electrical models might be even more significant so let's see what they think.

Isn't this exactly what this PR does? 

Remember, simulations don't need neurom, only morphological features depend on neurom. 
I agree, we can somehow flag the ABD, or add a special filter in neurom. We can just extract the list of basals which have an axon on them in the feature computation.

eleftherioszisis:
> So to reply to @mgeplf we need to have two neurites: one axon (blue) starting from the dendrite, one dendrite (red) starting from the soma but excluding the axon. These are just suggestions from my side, I think the impact on morphometrics will be important but the impact on electrical models might be even more significant so let's see what they think.

Just to clarify in case this is an issue of definitions: Given that by definition a cell neurite starts from the soma, imho we need a dedicated neurite type that would reflect these compound neurites as a whole and not make a Morphology have separate neurite objects for each subtree.

arnaudon:
The only issue I see for neurom is for trunk angles, because angles between normal neurites and this axon don't make sense anymore. We could add a neurite attribute `.at_soma` or something to filter these out in some feature computations (or only retain them). The path lengths will be the length from the basal/axon bif, but that's fine, no? Imho, we should just make the user aware he/she is dealing with this special case, so he/she needs to know how to interpret some features.

arnaudon:
https://github.com/BlueBrain/MorphIO/issues/361 so asc files don't allow for this axon bearing dendrites, right?

arnaudon:
As just noticed with @annakristinkaufmann on the SEU morphologies, we are a bit stuck with this: 
we have to run sanitize on the original .swc, but we get hit by: https://github.com/BlueBrain/NeuroR/blob/master/neuror/sanitize.py#L36 which should not raise but be handled properly with our codes.  It does not even seem we can pass this hurdle by converting go .asc (https://github.com/BlueBrain/MorphIO/issues/361). So these morphologies are not currently usable by us.

eleftherioszisis:
> As just noticed with @annakristinkaufmann on the SEU morphologies, we are a bit stuck with this:
we have to run sanitize on the original .swc, but we get hit by: https://github.com/BlueBrain/NeuroR/blob/master/neuror/sanitize.py#L36 which should not raise but be handled properly with our codes. It does not even seem we can pass this hurdle by converting go .asc (BlueBrain/MorphIO#361). So these morphologies are not currently usable by us.

Could you please move that to an issue in NeuroR so that we can keep track of it?

arnaudon:
sovled here for mcar: https://github.com/BlueBrain/morphology-workflows/pull/62

PR #979, opened 2022-02-09
Use readonly morphio Morphology

eleftherioszisis:
* Use exclusively morphio's readonly Morphology
* Morphology inheritance turned into composition
* Direct mutation removed
* Section booleaness removed as it is more confusing than useful


tests/features/test_morphology.py
```diff
@@ -59,12 +61,14 @@
...
 def _add_neurite_trunk(morph, elevation, azimuth, neurite_type=SectionType.basal_dendrite):
     """Add a neurite from the elevation and azimuth to a given morphology."""
+    mut = morph.to_morphio().as_mutable()
     new_pts = np.array(
         morphmath.vector_from_spherical(elevation, azimuth),
         ndmin=2
     )
     point_lvl = PointLevel(new_pts, [1])
-    morph.append_root_section(point_lvl, neurite_type)
+    mut.append_root_section(point_lvl, neurite_type)
```
mgeplf, line 70: ah, cool, this is exactly what I was thinking

tox.ini
```diff
@@ -14,7 +14,7 @@ envlist =
 [testenv]
 deps = {[base]testdeps}
 extras = plotly
-commands = pytest {posargs}
+commands = pytest {posargs} tests/
```
adrien-berchet, line 17: Why? Adding this will prevent us to use a command like `tox -e py38 -- tests/test_something.py`.
I think it would be better to add this in the `pyproject.toml` file:
```toml
[tool.pytest.ini_options]
testpaths = [
    "tests",
]
```
So by default it only uses the tests from the `tests` directory but it is still possible to specify which tests to run with tox.

neurom/core/morphology.py
```diff
@@ -523,37 +497,37 @@ def __hash__(self):
...
+class Morphology:
     """Class representing a simple morphology."""
 
     def __init__(self, filename, name=None):
         """Morphology constructor.
 
         Args:
-            filename (str|Path): a filename
-            name (str): a option morphology name
+            filename (str|Path): a filename or morphio.{,mut.}Morphology object
```
adrien-berchet, line 512: What does mean the `{,mut.}` here?

neurom/core/morphology.py
```diff
@@ -523,37 +497,37 @@ def __hash__(self):
...
+            name (str): a optional morphology name
```
adrien-berchet, line 513:
```suggestion
            name (str): an optional morphology name
```
I am not sure but I would use `an` in this case.

neurom/core/morphology.py
eleftherioszisis: It's  a typo, sorry. It was meant to be `morphio.{mut}.Morphology`, e.g. either readonly or mutable.

neurom/core/morphology.py
eleftherioszisis: Fixed

tox.ini
eleftherioszisis: Indeed. Thanks for the suggestion. I added it.

neurom/core/morphology.py
adrien-berchet: ah ok ok

adrien-berchet (approved):
LGTM, thanks!

neurom/core/morphology.py
```diff
@@ -511,10 +489,6 @@ def iter_sections(self, order=Section.ipreorder, neurite_order=NeuriteIter.FileO
         """
         return iter_sections(self, iterator_type=order, neurite_order=neurite_order)
 
-    def __nonzero__(self):
-        """If has root node."""
-        return bool(self.morphio_root_node)
-
```
eleftherioszisis, line 517: @mgeplf you mentioned you don't agree with removing the section booleaness. Is this still true?

mgeplf: no; I can see how it's confusing; one can check if there are children more explicitly.

mgeplf (approved):
LGTM

PR #981, opened 2022-02-16
Mixed subtree processing

mgeplf:


neurom/features/morphology.py
```diff
@@ -55,11 +55,60 @@
 from neurom.core.types import NeuriteType
 from neurom.features import feature, NameSpace, neurite as nf
 from neurom.utils import str_to_plane
-
+from neurom.core.morphology import Neurite
+from collections import OrderedDict
```
mgeplf, line 59: I think `dict`s are ordered by default fm 3.7 onward?

neurom/features/morphology.py
```diff
@@ -55,11 +55,60 @@
...
+
+    Note: Only two different mixed types are allowed
+    """
+    homogeneous_neurites = OrderedDict([(neurite.root_node.type, neurite)])
+    for section in neurite.root_node.ipreorder():
+        if section.type not in homogeneous_neurites:
+            homogeneous_neurites[section.type] = Neurite(section)
+    if len(homogeneous_neurites) != 2:
+        raise TypeError(
+            f"Subtree types must be exactly two. Found {len(homogeneous_neurites)} instead.")
```
mgeplf, line 77: is this an invariant we need? could it be more?

I've been asking myself if we should just handle the AoD case, or make a more general solution; I have no idea what biologically makes sense, though.

tests/features/test_get_features.py
```diff
@@ -627,23 +627,33 @@ def test_section_radial_distances_origin():
...
-    nsecs = features.get('number_of_sections_per_neurite', NEURON,
-                         neurite_type=NeuriteType.basal_dendrite)
-    assert len(nsecs) == 2
-    assert np.all(nsecs == [21, 21])
-
-    nsecs = features.get('number_of_sections_per_neurite', NEURON,
-                         neurite_type=NeuriteType.apical_dendrite)
-    assert len(nsecs) == 1
-    assert np.all(nsecs == [21])
+    for use_subtrees in (True, False):
```
mgeplf, line 630: FYI; I added this to see if we maintain backwards compat

neurom/features/morphology.py
```diff
@@ -55,11 +55,60 @@
...
+        # we don't know a priori the types of the subtrees in the neurites, thus we cannot
+        # filter by time at the neurite level yet
+        for neurite in iter_neurites(morph):
+
+            # map the function for each subtree with homogeneous type, propagating the type
+            # for a section level filtering
+            if neurite.morphio_root_node.is_heterogeneous():
+                yield from _map_homogeneous_subtrees(function, neurite, neurite_type)
+            else:
+                if is_type(neurite_type)(neurite):
```
mgeplf, line 106:
```suggestion
                yield from iter_neurites(neurite, mapfun=function, filt=is_type(neurite_type))
```

mgeplf: hmm, or not; I think we need a more complex mixed morph :)

neurom/features/morphology.py
eleftherioszisis: Indeed. I thought it was an implementation detail but I checked there is an insertion order guarantee from 3.7

neurom/features/__init__.py
```diff
@@ -90,6 +90,11 @@ def _get_feature_value_and_func(feature_name, obj, **kwargs):
         raise NeuroMError('Only Neurite, Morphology, Population or list, tuple of Neurite,'
                           ' Morphology can be used for feature calculation')
 
+    use_subtrees = False
```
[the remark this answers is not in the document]
eleftherioszisis: Done.

neurom/core/morphology.py
```diff
@@ -75,6 +75,10 @@ def append_section(self, section):
             return self.morphio_section.append_section(section.morphio_section)
         return self.morphio_section.append_section(section)
 
+    def is_homogeneous_point(self):
```
mgeplf, line 78: `is_homogeneous_section`?

perhaps note that it's not recursive

neurom/core/morphology.py
```diff
@@ -211,7 +215,24 @@ def __repr__(self):
              NeuriteType.undefined: 4}
 
 
-def iter_neurites(obj, mapfun=None, filt=None, neurite_order=NeuriteIter.FileOrder):
+def _homogeneous_subtrees(neurite):
+    """Returns a dictionary the keys of which are section types and the values are the sub-neurites.
```
mgeplf, line 219: seems to return a list

neurom/core/morphology.py
```diff
@@ -240,23 +261,45 @@ def iter_neurites(obj, mapfun=None, filt=None, neurite_order=NeuriteIter.FileOrd
...
         >>> n_points = [n for n in iter_neurites(pop, mapping, filter)]
     """
-    neurites = ((obj,) if isinstance(obj, Neurite) else
-                obj.neurites if hasattr(obj, 'neurites') else obj)
+    def extract_subneurites(neurite):
+        if neurite.is_heterogeneous():
+            return _homogeneous_subtrees(neurite)
+        return [neurite]
+
+    neurites = (
```
mgeplf, line 269: is this `black` formatted or something else? I found the old one more readable (just my opinion)

neurom/core/morphology.py
```diff
@@ -240,23 +261,45 @@ def iter_neurites(obj, mapfun=None, filt=None, neurite_order=NeuriteIter.FileOrd
...
+
+    return map(mapfun, neurite_iter)
 
 
 def iter_sections(neurites,
                   iterator_type=Section.ipreorder,
                   neurite_filter=None,
-                  neurite_order=NeuriteIter.FileOrder):
+                  neurite_order=NeuriteIter.FileOrder,
+                  section_filter=None):
```
mgeplf, line 302: add to docstring

neurom/features/morphology.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
+# Copyright (c) 2019, Ecole Polytechnique Federale de Lausanne, Blue Brain Project
```
mgeplf, line 1: back in time?

neurom/features/morphology.py
```diff
@@ -428,36 +459,47 @@ def _mean_radius(neurite):
                     "values excluded all the points of the section so the radius of the first "
                     "point after the 'min_length_filter' path distance is returned."
                 )
-                # pylint: disable=invalid-unary-operand-type
+                # pylint: disable = invalid-unary-operand-type
```
mgeplf, line 462: pylint handles spaces in the disable? I'd prefer to keep it the old way, if that's ok

mgeplf (commented):
first pass; now I need to digest it more

neurom/core/morphology.py
eleftherioszisis: Fixed docstring.

neurom/core/morphology.py
eleftherioszisis: Added.

neurom/features/morphology.py
eleftherioszisis: oops

neurom/features/morphology.py
eleftherioszisis: lol, fixed

neurom/core/morphology.py
mgeplf: Your name makes more sense, now that I see the symmetry

neurom/core/morphology.py
mgeplf: "when"; ok, no problem, I guess I'll have to get used to it.

neurom/core/morphology.py
```diff
@@ -240,23 +259,45 @@ def iter_neurites(obj, mapfun=None, filt=None, neurite_order=NeuriteIter.FileOrd
         >>> mapping = lambda n : len(n.points)
         >>> n_points = [n for n in iter_neurites(pop, mapping, filter)]
     """
-    neurites = ((obj,) if isinstance(obj, Neurite) else
-                obj.neurites if hasattr(obj, 'neurites') else obj)
+    def extract_subneurites(neurite):
```
mgeplf, line 262: can we move this to where it's used

neurom/core/morphology.py
```diff
@@ -240,23 +259,45 @@ def iter_neurites(obj, mapfun=None, filt=None, neurite_order=NeuriteIter.FileOrd
...
+        neurites = flatten(map(extract_subneurites, neurites))
+
     neurite_iter = iter(neurites) if filt is None else filter(filt, neurites)
-    return neurite_iter if mapfun is None else map(mapfun, neurite_iter)
+
+    if mapfun is None:
+        return neurite_iter
+
+    if use_subtrees:
+        return (mapfun(neurite, section_type=neurite.type) for neurite in neurite_iter)
```
mgeplf, line 291:
```suggestion
        mapfun = functools.partial(mapfun, section_type=neurite.type)
```

mgeplf: I'd also move that into the `if use_subtrees` (w/ a guard on `mapfun is not None`)

neurom/features/__init__.py
```diff
@@ -64,15 +66,25 @@ def _flatten_feature(feature_shape, feature_value):
     return reduce(operator.concat, feature_value, [])
 
 
-def _get_neurites_feature_value(feature_, obj, neurite_filter, kwargs):
```
mgeplf, line 67: I'm really weirded out by `kwargs` not being `**kwargs`, if it were, then `use_subtrees` could go in the right location

neurom/features/__init__.py
```diff
@@ -87,43 +99,72 @@ def _get_feature_value_and_func(feature_name, obj, **kwargs):
...
                 res = [feature_(s, **kwargs) for s in obj]
+
     elif isinstance(obj, Morphology):
+
         # input is a morphology
         if feature_name in _MORPHOLOGY_FEATURES:
+
             feature_ = _MORPHOLOGY_FEATURES[feature_name]
+
+            if "use_subtrees" in inspect.signature(feature_).parameters:
```
mgeplf, line 133: sneaky, not super excited about it, but I suppose it works, and I can't think of something better

Can we at least factor it out, so the other places it's used can all be updated at once if we think of something better?

neurom/features/population.py
```diff
@@ -45,15 +45,18 @@
...
 from neurom.features import feature, NameSpace
-from neurom.features.morphology import sholl_crossings
+from neurom.features import morphology as mf
 
 feature = partial(feature, namespace=NameSpace.POPULATION)
 
 
 @feature(shape=(...,))
-def sholl_frequency(morphs, neurite_type=NeuriteType.all, step_size=10, bins=None):
+def sholl_frequency(
```
mgeplf, line 57: aside: why does population need a custom sholl, and no where else?

tests/test_mixed.py
```diff
@@ -0,0 +1,1952 @@
+import sys
```
mgeplf, line 1: I haven't read these; should I?

One of the concerns I have is that when ppl implement new features, they will forget to deal w/ the subtree problem; I'm wondering if we could add a test here that takes *all* the features, and compares running it against a very simple homogenous and mixed tree, and checks if the results are different?  That way the test would fail if someone forgets (in theory)

neurom/features/morphology.py
```diff
@@ -45,27 +45,44 @@
...
+from itertools import chain
 from functools import partial
+from collections.abc import Iterable
 import math
 import numpy as np
 
 from neurom import morphmath
-from neurom.core.morphology import iter_neurites, iter_segments, Morphology
+from neurom.core.morphology import (
+    iter_neurites, iter_sections, iter_segments, iter_points, Morphology
```
mgeplf, line 56: black doesn't leave a trailing comma?

neurom/features/morphology.py
```diff
@@ -499,13 +541,28 @@ def _count_crossings(neurite, radius):
...
             radii = [morph.soma.radius]
-    return [sum(_count_crossings(neurite, r)
-                for neurite in iter_neurites(morph, filt=is_type(neurite_type)))
-            for r in radii]
+
+    if isinstance(morph, Iterable):
+        sections = filter(is_type(neurite_type), morph)
+    else:
+        if use_subtrees:
+            sections = iter_sections(morph, section_filter=is_type(neurite_type))
```
mgeplf, line 549: It's only these kinds of calls that make me wonder if the `section_filter` can be folded into the `neurite_type` filter, which is what I've been waffling about.  There's something about this if/else...

neurom/core/morphology.py
eleftherioszisis: Yeah, removed the function and used it directly where it is needed.

neurom/core/morphology.py
eleftherioszisis: The main issue with this is that it passes the type of each neurite, not a fixed one (neurite.type vs. neurite_type).

neurom/features/__init__.py
eleftherioszisis: Changed it into using `**kwargs`.

tests/test_mixed.py
eleftherioszisis: The tests look into the features dictionaries and raise if there are new tests that are not included.
```python
    features_not_tested = list(
        set(_NEURITE_FEATURES) - set(features.keys()) - set(_MORPHOLOGY_FEATURES)
    )

    assert not features_not_tested, (
        "The following morphology tests need to be included in the tests:\n\n" +
        "\n".join(sorted(features_not_tested)) + "\n"
    )
```

neurom/features/morphology.py
eleftherioszisis: The module is not formatted with black in general. I was thinking of making a pass on the entire package when v4 is merged so that we don't have a lot of irrelevant changes here.

neurom/features/morphology.py
eleftherioszisis: The module is not formatted with black in general. I was thinking of making a pass on the entire package when v4 is merged so that we don't have a lot of irrelevant changes here.

neurom/features/morphology.py
eleftherioszisis: Well, for some reason this specific feature should also work with a list of sections...

from the docstring:
```
    This function can also be used with a list of sections, as follow::
        secs = (sec for sec in nm.iter_sections(morph) if complex_filter(sec))
        sholl = nm.features.neuritefunc.sholl_crossings(secs,
                                                        center=morph.soma.center,
                                                        radii=np.arange(0, 1000, 100))
```
I don't know why it is like this, but I didn't want to break backward compatibility even though it feels wrong to me to support sections at the morphology level of this feature.

doc/source/heterogeneous.rst
```diff
@@ -0,0 +1,222 @@
...
+
+    basal, axon_carrying_dendrite, apical = list(iter_neurites(m))
+
+    print(basal.type, axon_carrying_dendrite.type, apical.type)
+
+would print the types ``(basal_dendrite, basal_dendrite, apical_dendrite)``.
+I.E. the axon-carrying dendrite would be treated as a basal dendrite.
+For feature extraction and checks, the axon-carrying dendrite is treated as a basal dendrite.
+Features, for which an axon neurite type is passed, do not have access to the axonal part of the neurite.
+For instance, the number of basal and axon neurites will be two and zero respectively.
```
mgeplf, line 87: Perhaps add what a feature would do in this case; ie, would the `volume` include the ACD or just the the basal portion?

neurom/core/morphology.py
```diff
@@ -211,7 +227,22 @@ def __repr__(self):
              NeuriteType.undefined: 4}
 
 
-def iter_neurites(obj, mapfun=None, filt=None, neurite_order=NeuriteIter.FileOrder):
+def _homogeneous_subtrees(neurite):
+    """Returns a list of the root nodes of the sub-neurites.
+
+    A sub-neurite can be either the entire tree or a homogeneous downstream
+    sub-tree.
+    """
```
mgeplf, line 235: do we want to check/error/differenciate/document what happens if there is a `dendrite - axon - dendrite - axon` situation?

neurom/core/morphology.py
```diff
@@ -240,23 +271,44 @@ def iter_neurites(obj, mapfun=None, filt=None, neurite_order=NeuriteIter.FileOrd
         >>> mapping = lambda n : len(n.points)
         >>> n_points = [n for n in iter_neurites(pop, mapping, filter)]
     """
-    neurites = ((obj,) if isinstance(obj, Neurite) else
-                obj.neurites if hasattr(obj, 'neurites') else obj)
+    neurites = (
```
mgeplf, line 274:
```suggestion
    neurites = (obj, )
    if isinstance(obj, Neurite):
        neurites = obj.neurites
    elif hasaattr(obj, "neurites"):
        neurites = obj
```
shorter and clearer?

neurom/core/morphology.py
```diff
@@ -240,23 +271,44 @@ def iter_neurites(obj, mapfun=None, filt=None, neurite_order=NeuriteIter.FileOrd
...
+
     if neurite_order == NeuriteIter.NRN:
         if isinstance(obj, Population):
             warnings.warn('`iter_neurites` with `neurite_order` over Population orders neurites'
                           'within the whole population, not within each morphology separately.')
         last_position = max(NRN_ORDER.values()) + 1
         neurites = sorted(neurites, key=lambda neurite: NRN_ORDER.get(neurite.type, last_position))
 
+    if use_subtrees:
+
```
mgeplf, line 290: nit: let's start dropping the extra line between `if` and the next statement, to keep it the same as other code

neurom/features/__init__.py
```diff
@@ -64,15 +66,30 @@ def _flatten_feature(feature_shape, feature_value):
...
-def _get_neurites_feature_value(feature_, obj, neurite_filter, kwargs):
+def _get_neurites_feature_value(feature_, obj, neurite_filter, use_subtrees, **kwargs):
     """Collects neurite feature values appropriately to feature's shape."""
     kwargs.pop('neurite_type', None)  # there is no 'neurite_type' arg in _NEURITE_FEATURES
-    return reduce(operator.add,
-                  (feature_(n, **kwargs) for n in iter_neurites(obj, filt=neurite_filter)),
-                  0 if feature_.shape == () else [])
+
+    return reduce(
+        operator.add,
```
mgeplf, line 74: equivalent to `sum()`, no?

neurom/features/__init__.py
```diff
@@ -87,43 +104,72 @@ def _get_feature_value_and_func(feature_name, obj, **kwargs):
...
+        raise NeuroMError(
+            "Only Neurite, Morphology, Population or list, tuple of Neurite, Morphology"
+            f"can be used for feature calculation. Got: {obj}"
+        )
 
     neurite_filter = is_type(kwargs.get('neurite_type', NeuriteType.all))
     res, feature_ = None, None
 
     if isinstance(obj, Neurite) or (is_obj_list and isinstance(obj[0], Neurite)):
+
```
mgeplf, line 116: this gets cleaned up in the other PR, so I'm skipping it

neurom/features/bifurcation.py
```diff
@@ -33,7 +33,7 @@
 from neurom.exceptions import NeuroMError
 from neurom.core.dataformat import COLS
 from neurom.core.morphology import Section
-from neurom.features.section import section_mean_radius
+from neurom.features import section as sf
```
mgeplf, line 36: for the few places it gets used, let's just use the full path?

neurom/features/morphology.py
mgeplf: before v4 is released, yeah?

neurom/features/morphology.py
mgeplf: if we drop that, do any of the tests fail?  I'd consider removing it if not?

mgeplf (commented):
I think one decision we should make is whether to throw a warning (w/ a way of turning it off) if it's a mixed subtree, but ppl haven't specified it.

neurom/features/morphology.py
eleftherioszisis: Yeah, there is a test that checks the result of passing sections..

neurom/features/morphology.py
eleftherioszisis: I don't have a preference. We can do it before the release, yeah.

neurom/core/morphology.py
eleftherioszisis: I think it should be:
```python
if isinstance(obj, Neurite):
    neurites = (obj,)
elif hasattr(obj, "neurites"):
    neurites = obj.neurites
else:
    neurites = obj
```
right?

neurom/core/morphology.py
eleftherioszisis: I will try :D

neurom/features/bifurcation.py
eleftherioszisis: Sure.

neurom/features/morphology.py
eleftherioszisis: nope, just the single test where a list of sections is passed.

neurom/core/morphology.py
mgeplf: yeah, whatever is correct :)

neurom/features/__init__.py
mgeplf: I guess I'm used to using `sum()` for all sorts of application of the `+` operator.  I find `reduce` requires more thinking; whatever you prefer

neurom/core/morphology.py
eleftherioszisis: Added a warning if the neurite is not an AcD

eleftherioszisis: Also, I changed the way that identifies the subtrees so that it creates a subtree for each different type it encounters downstream, regardless if it has been added to the list. This allows to have a `dendrite-axon-dendrite-axon` situation, which was not possible before due to the set I was using.

neurom/features/__init__.py
eleftherioszisis: As we discussed, let's revisit this in the other PR for the refactoring of this module.

eleftherioszisis (approved):

eleftherioszisis (changes requested):
rebase yo

eleftherioszisis (approved):

PR #986, opened 2022-02-24
Fix the trunk_origin_radii feature

adrien-berchet:


arnaudon (approved):

neurom/features/morphology.py
```diff
@@ -380,11 +383,18 @@ def _mean_radius(points):
...
+                    "point is returned."
+                )
+                valid_pts[-1] = True
         if max_length_filter is not None:
             valid_pts = (valid_pts & (path_lengths <= max_length_filter))
         return points[valid_pts, COLS.R].mean()
 
-    return [_mean_radius(n.points)
+    return [_mean_radius(next(n.iter_sections()).points)
             for n in iter_neurites(morph, filt=is_type(neurite_type))]
```
eleftherioszisis, line 398:
```python
return [mean_radius(n) for n in  iter_neurites(morph, filt=is_type(neurite_type))]
```

eleftherioszisis: Then make the `mean_radius` closure to take a neurite:

```python
def mean_radius(neurite):
    interval_lengths = morphmath.interval_lengths(neurite.root_node.points)
    ...
```
Make it explicit. This way will also make it easier to extend it for mixed subtrees (#981).

eleftherioszisis (commented):
I am wondering if using the first section is a sufficient solution or it needs to go deeper?
@lidakanari , thoughts?

neurom/features/morphology.py
adrien-berchet: Ah yeah good point, I forgot about `root_node`.

neurom/features/morphology.py
```diff
@@ -374,18 +377,25 @@ def trunk_origin_radii(
...
         valid_pts = np.ones(len(path_lengths), dtype=bool)
         if min_length_filter is not None:
             valid_pts = (valid_pts & (path_lengths >= min_length_filter))
+            if not valid_pts.any():
+                warnings.warn(
+                    "In 'trunk_origin_radii': the 'min_length_filter' value is greater than the "
+                    "path distance of the last point of the last section so the radius of this "
+                    "point is returned."
+                )
+                valid_pts[-1] = True
```
eleftherioszisis, line 393: Would it be better if you returned early here? If the `min_length_filter` is greater than the path distance, the `max_length_filter` should be as well if specified. 

Therefore, I would:
```python
return points[-1, COLS.R]
```
What do you think?

adrien-berchet: You made me think more about edge cases and indeed there are several weird possibilities. In the last commit I tried to handle them, WDYT?

eleftherioszisis: I think it's great now! You still can return early though, can't you?

neurom/features/morphology.py
```diff
@@ -374,18 +377,56 @@ def trunk_origin_radii(
         return [n.root_node.points[0][COLS.R]
                 for n in iter_neurites(morph, filt=is_type(neurite_type))]
 
-    def _mean_radius(points):
+    if min_length_filter is not None and min_length_filter <= 0:
+        raise ValueError(
```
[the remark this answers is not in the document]
adrien-berchet: Good idea

neurom/features/morphology.py
adrien-berchet: Ah yes I forgot sorry

eleftherioszisis:
lgtm!

adrien-berchet:
Ok, cool!
I am not authorized to merge so I let you do.

neurom/features/morphology.py
```diff
@@ -374,18 +378,56 @@ def trunk_origin_radii(
...
+            valid_max = (path_lengths <= max_length_filter)
+            valid_pts = (valid_pts & valid_max)
+            if not valid_pts.any():
+                warnings.warn(
+                    "In 'trunk_origin_radii': the 'min_length_filter' and 'max_length_filter' "
+                    "values excluded all the points of the section so the radius of the first "
+                    "point after the 'min_length_filter' path distance is returned."
+                )
+                # pylint: disable=invalid-unary-operand-type
+                return points[np.argwhere(~valid_max)[0][0], COLS.R]
```
eleftherioszisis, line 427: An alternative way to do this simpler would be:
return `points[~valid_max, COLS.R][0]`, right?

I sit guaranteed here that `np.argwhere(~valid_max)[0][0]` will return at least one value?

adrien-berchet: Ah yes indeed we can just use a mask here.
The first element of `~valid_max` is always `True` because `max_length_filter > 0` so we are sure that at least the first point of the section is `True`.

eleftherioszisis: Cool. I don't have any other comments, so I am going to merge this. Thanks for your time!

adrien-berchet: Ok, cool. Thanks for your review!

PR #989, opened 2022-02-25
Use soma center in sholl_frequency distance calculation

### LINKED ISSUE #988: `sholl_frequency` does not use soma center to calculate max radius
The soma center is not taken into account when calculating the distance to the points to find the max radius (L481):
https://github.com/BlueBrain/NeuroM/blob/e31eb63b522a3558233d821fcfbb178e6754aa56/neurom/features/morphology.py#L479-L485

Of course, for morphologies that are centered at (0, 0, 0) this does not make a difference, but it will mess up the rest.

eleftherioszisis:
Fixes #988

### DIFF
--- a/neurom/features/morphology.py
+++ b/neurom/features/morphology.py
@@ -478,8 +478,10 @@ def sholl_frequency(morph, neurite_type=NeuriteType.all, step_size=10, bins=None
 
     if bins is None:
         min_soma_edge = morph.soma.radius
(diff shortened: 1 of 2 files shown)


mgeplf (approved):

mgeplf:
Nice catch

mgeplf:
Would it be worth adding a test where the soma isn't at the origin?

eleftherioszisis:
> Would it be worth adding a test where the soma isn't at the origin?

Yes, good point. I added one.

PR #1002, opened 2022-03-11
Make deprecation warnings respect pre-existing warnings configuration

### LINKED ISSUE #997: Inappropriate `warnings.simplefilter`
in `neurom.utils.warn_deprecated` `warnings.simplefilter` is used to enforce that a deprecation warning is shown:
https://github.com/BlueBrain/NeuroM/blob/c570a2faa18f4e417fcdb743f6487beee0c60c3e/neurom/utils.py#L40-L44

This is problematic for several reasons
* It's not the business of a library to configure how the warnings are shown
* There is no way for the library user to easily disable those warnings
* By calling `warnings.simplefilter` `neurom` modifies the global state and has the side-effect of permanently setting ` warnings.simplefilter('default', DeprecationWarning)`. E.g
  ```py
  import warnings
  import neurom as nm
  
  print("before:", warnings.filters)
  nm.load_neuron(some_path)  # this triggers the warning `neurom.io.utils.load_neuron` is deprecated
  print("after:", warnings.filters)
  ```
  output:
  ```
  before: []
  after: [('default', None, <class 'DeprecationWarning'>, None, 0), ('always', None, <class 'DeprecationWarning'>, None, 0)]
  ```

I would strongly advocate for removing the two `warnings.simplefilter` lines.

P.S.: since `warnings.warn` is wrapped into `warn_deprecated` setting `stacklevel=3` should be considered.

eleftherioszisis:
Fixes #997

### DIFF
--- a/neurom/utils.py
+++ b/neurom/utils.py
@@ -40,7 +40,7 @@ from neurom.core.dataformat import COLS
 def warn_deprecated(msg):
     """Issue a deprecation warning."""
     with warnings.catch_warnings():
-        warnings.simplefilter("always", DeprecationWarning)
+        warnings.simplefilter("once", DeprecationWarning)
         warnings.warn(msg, category=DeprecationWarning, stacklevel=3)
 
 
--- a/neurom/utils.py
+++ b/neurom/utils.py
@@ -35,13 +35,14 @@ from functools import wraps
 import numpy as np
 
 from neurom.core.dataformat import COLS
+from neurom.exceptions import NeuroMDeprecationWarning
+
+import warnings
 
 
 def warn_deprecated(msg):
     """Issue a deprecation warning."""
-    with warnings.catch_warnings():
-        warnings.simplefilter("always", DeprecationWarning)
-        warnings.warn(msg, category=DeprecationWarning, stacklevel=3)
+    warnings.warn(msg, category=NeuroMDeprecationWarning, stacklevel=2)
 
 
 def deprecated(fun_name=None, msg=""):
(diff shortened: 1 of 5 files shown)


neurom/utils.py
```diff
@@ -39,9 +39,9 @@
 
 def warn_deprecated(msg):
     """Issue a deprecation warning."""
-    warnings.simplefilter('always', DeprecationWarning)
-    warnings.warn(msg, category=DeprecationWarning, stacklevel=2)
-    warnings.simplefilter('default', DeprecationWarning)
+    with warnings.catch_warnings():
+        warnings.simplefilter("always", DeprecationWarning)
```
adrien-berchet, line 43: Isn't `"once"` enough? `"always"` might spam the logs a bit too much, isn't it?

eleftherioszisis: Yeah, good point.

eleftherioszisis: I changed it but it doesn't seem to have an effect. I believe it's because of the context manager resetting the configuration at each call.

Stannislav (approved):
Thanks for fixing!

I left a comment about `UserWarning` vs. `DeprecationWarning`, where I'm not sure about some details.

Otherwise I think the main issue that the warning was not possible to disable and that the global configuration was changed as side-effect was addressed. So if you're happy with the rest, then I'm happy too. :)

eleftherioszisis:
Thank you @Stannislav for testing things out. Indeed increasing the stacklevel to 3 seems to be the best solution.

What was surprising to me was that the visibility of the warning changes depending on the stack level. So even though for a stack level of 2 it is hidden by default, stacklevel=3 brings the warning level to `__main__` (where the load_neuron function is called from), which has deprecation warnings as visible by default.

Stannislav (approved):
Thanks @eleftherioszisis, LGTM.

PR #1005, opened 2022-03-16
Refactor features

### LINKED ISSUE #1004: Mismatching behavior between segment and section taper rates
Segment taper rates calculate the absolute slope of the segment diameters:

https://github.com/BlueBrain/NeuroM/blob/a9e3020f242b3ff04cdc244e068afdfc7137dfcb/neurom/features/neurite.py#L230-L243

```
np.abs(diff[:, COLS.R])
```
whereas the section taper rates calculate the signed slope from the linear fit to the diameters:

https://github.com/BlueBrain/NeuroM/blob/a9e3020f242b3ff04cdc244e068afdfc7137dfcb/neurom/features/neurite.py#L247-L259

Is this to be expected? I would expect that in both cases the signed slope would be desired.

```python
from io import StringIO

import numpy as np
from numpy import testing as npt

from neurom.features import neurite
from neurom import load_morphology

neu = load_morphology(StringIO(u"""((CellBody) (0 0 0 2))
                                    ((Dendrite)
                                     (0 0 0 2)
                                     (1 0 0 1)
                                     (2 0 0 0))"""), reader='asc')


npt.assert_almost_equal(neurite.section_taper_rates(neu.neurites[0]), [-1])
npt.assert_almost_equal(neurite.segment_taper_rates(neu.neurites[0]), [-1., -1])

```
The second test fails with `[1, 1]`

@lidakanari @arnaudon @adrien-berchet

eleftherioszisis:
* Reduce feature logic by using the neurite/section/segment iterators and mappers
* Move nested functions in `neurom.features.neurite.py` into `neurom.features.section.py` and add tests.
* Fixed #1004

### DIFF
--- a/neurom/features/morphology.py
+++ b/neurom/features/morphology.py
@@ -63,6 +63,12 @@ from neurom.geom import convex_hull
 feature = partial(feature, namespace=NameSpace.NEURON)
 
 
+def _map_neurites(function, morph, neurite_type):
+    return list(
+        iter_neurites(morph, mapfun=function, filt=is_type(neurite_type))
+    )
+
+
 @feature(shape=())
 def soma_volume(morph):
     """Get the volume of a morphology's soma."""
@@ -76,7 +82,7 @@ def soma_surface_area(morph):
     Note:
         The surface area is calculated by assuming the soma is spherical.
     """
-    return 4 * math.pi * morph.soma.radius ** 2
+    return 4.0 * math.pi * morph.soma.radius ** 2
 
 
 @feature(shape=())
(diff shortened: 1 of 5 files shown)


neurom/features/morphology.py
```diff
@@ -428,27 +432,28 @@ def _mean_radius(neurite):
...
+        return morphmath.section_length(neurite.root_node.points)
+
+    return _map_neurites(trunk_section_length, morph, neurite_type)
 
 
 @feature(shape=())
 def number_of_neurites(morph, neurite_type=NeuriteType.all):
     """Number of neurites in a morph."""
-    return sum(1 for _ in iter_neurites(morph, filt=is_type(neurite_type)))
+    return sum(_map_neurites(lambda _: 1, morph, neurite_type))
```
mgeplf, line 450: FYI, I think the old way was done to reduce the list creation, so only an iterable is needed.  I don't think that perf is critical here, so for uniformity sake, I think it's ok; but something to keep in mind.

eleftherioszisis: The reason behind offloading almost everything to `_map_neurites` is mainly because it will be necessary for the mixed type subtree processing if we want to avoid adding extra logic in each feature.

It might be worth rethinking however at which stage of the feature calculation the type of the iterable (e.g. list) is applied.

eleftherioszisis: I could however replace `sum` with `len` so that we avoid the extra iteration.

adrien-berchet (approved):

PR #1008, opened 2022-03-21
Fix principal direction extent

### LINKED ISSUE #1007: Bug in principal direction extent
There is a bug in the calculation of the principal direction extents:

https://github.com/BlueBrain/NeuroM/blob/fdf9f78be79508e34c0349936e233c184279c955/neurom/morphmath.py#L462-L496

At the point where the most negative value is subtracted from the entire extent:
```python
         if scalar_projs[0] < 0.: 
             extent -= scalar_projs[0] 
```
It shouldn't subtract from the entire extent, rather from the i-th component `extent[i]`.

Failing test example with points on a circle with radius `0.5`, and center at `0.0`:
```python
from neurom.morphmath import principal_direction_extent
from numpy import testing as npt

def test_principal_direction_extent():

    circle_points = np.array([
        [ 5.0e-01,  0.0e+00,  0.0e+00],
        [ 4.7e-01,  1.6e-01,  0.0e+00],
        [ 3.9e-01,  3.1e-01,  0.0e+00],
        [ 2.7e-01,  4.2e-01,  0.0e+00],
        [ 1.2e-01,  4.8e-01,  0.0e+00],
        [-4.1e-02,  5.0e-01,  0.0e+00],
        [-2.0e-01,  4.6e-01,  0.0e+00],
        [-3.4e-01,  3.7e-01,  0.0e+00],
        [-4.4e-01,  2.4e-01,  0.0e+00],
        [-4.9e-01,  8.2e-02,  0.0e+00],
        [-4.9e-01, -8.2e-02,  0.0e+00],
        [-4.4e-01, -2.4e-01,  0.0e+00],
        [-3.4e-01, -3.7e-01,  0.0e+00],
        [-2.0e-01, -4.6e-01,  0.0e+00],
        [-4.1e-02, -5.0e-01,  0.0e+00],
        [ 1.2e-01, -4.8e-01,  0.0e+00],
        [ 2.7e-01, -4.2e-01,  0.0e+00],
        [ 3.9e-01, -3.1e-01,  0.0e+00],
        [ 4.7e-01, -1.6e-01,  0.0e+00],
        [ 5.0e-01, -1.2e-16,  0.0e+00]
    ])

    npt.assert_allclose(
        principal_direction_extent(circle_points),
        [1., 1., 0.],
    )
```
Which returns instead `[1.49, 1.0, 0.0]`.

Thanks, @lidakanari for finding that there is something wrong with the principal direction extent function.

@arnaudon @adrien-berchet

eleftherioszisis:
Fixes #1007

### DIFF
--- a/neurom/morphmath.py
+++ b/neurom/morphmath.py
@@ -471,8 +471,6 @@ def principal_direction_extent(points):
 
     Returns:
         extents : the extents for each of the eigenvectors of the cov matrix
-        eigs : eigenvalues of the covariance matrix
-        eigv : respective eigenvectors of the covariance matrix
     """
     # pca can be biased by duplicate points
     points = np.unique(points, axis=0)
@@ -483,14 +481,8 @@ def principal_direction_extent(points):
     # principal components
     _, eigv = pca(points)
 
-    extent = np.zeros(3)
-
-    for i in range(eigv.shape[1]):
-        # orthogonal projection onto the direction of the v component
-        scalar_projs = np.sort(np.array([np.dot(p, eigv[:, i]) for p in points]))
-        extent[i] = scalar_projs[-1]
-
-        if scalar_projs[0] < 0.:
-            extent -= scalar_projs[0]
-
-    return extent
+    # for each eigenvector calculate the range of the scalar projections of all points on it
+    return np.fromiter(
+        (np.ptp(np.inner(points, eigv[:, i])) for i in range(eigv.shape[1])),
+        dtype=float,
+    )
--- a/neurom/morphmath.py
+++ b/neurom/morphmath.py
@@ -481,8 +481,8 @@ def principal_direction_extent(points):
     # principal components
     _, eigv = pca(points)
 
-    # for each eigenvector calculate the range of the scalar projections of all points on it
-    return np.fromiter(
-        (np.ptp(np.inner(points, eigv[:, i])) for i in range(eigv.shape[1])),
-        dtype=float,
-    )
+    # for each eigenvector calculate the scalar projection of the points on it (n_points, n_eigv)
+    scalar_projections = points.dot(eigv)
+
+    # and return the range of the projections (abs(max - min)) along each column (eigenvector)
+    return np.ptp(scalar_projections, axis=0)
--- a/neurom/morphmath.py
+++ b/neurom/morphmath.py
@@ -462,15 +462,17 @@ section_length = path_distance
 def principal_direction_extent(points):
     """Calculate the extent of a set of 3D points.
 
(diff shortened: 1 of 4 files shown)


eleftherioszisis:
@lidakanari , could you please check that with these changes you get the expected results?

neurom/morphmath.py
```diff
@@ -483,14 +481,8 @@ def principal_direction_extent(points):
...
-        scalar_projs = np.sort(np.array([np.dot(p, eigv[:, i]) for p in points]))
-        extent[i] = scalar_projs[-1]
-
-        if scalar_projs[0] < 0.:
-            extent -= scalar_projs[0]
-
-    return extent
+    # for each eigenvector calculate the range of the scalar projections of all points on it
+    return np.fromiter(
+        (np.ptp(np.inner(points, eigv[:, i])) for i in range(eigv.shape[1])),
```
arnaudon, line 486: would this work?
`return np.ptp(point.dot(eigv.T))`

eleftherioszisis: yeah, but it would be unreadable :)

eleftherioszisis: However, it's a good point that `points.dot` performs matrix mult correctly wrt eigv.

lidakanari (commented):
The function seems to return correct values for the tested cases. However, the signature of what it returns changes with respect to the previous one, as it now returns a vector of the same dimension as the input data. i.e. 2d point cloud returned 3d vector before, now it returns 2d.

eleftherioszisis:
> The function seems to return correct values for the tested cases. However, the signature of what it returns changes with respect to the previous one, as it now returns a vector of the same dimension as the input data. i.e. 2d point cloud returned 3d vector before, now it returns 2d.

Indeed, the dimension of the extents depends on the dimension of the points. Therefore, for 2D points it would return a 2D vector of extents, and for 3D a 3D respectively. For NeuroM purposes, only 3D points are used, so this is not really an issue.

arnaudon (approved):

lidakanari (approved):
Ok, that should be good. Also something to keep in mind, currently the extents are **not** sorted (i.e. largest to smallest). If this is by choice, it would be good to mention in documentation because it might be confusing.

adrien-berchet (approved):

eleftherioszisis:
> Ok, that should be good. Also something to keep in mind, currently the extents are **not** sorted (i.e. largest to smallest). If this is by choice, it would be good to mention in documentation because it might be confusing.

That's true indeed. There was no requirement that they are sorted from max to min, which imho renders the `direction` argument a bit useless in the respective feature. 

Should I order them?

Issue #1012, opened 2022-03-23
New feature: shape factor


lidakanari:
This feature needs to be implemented on the 2D projection of a morphology and it will describe the shape factor in the projected plane i.e. x-y. We will therefore need to compute this in all three projections at the morphology level.

Shape factor is taken from https://ijcsi.org/papers/IJCSI-10-6-1-162-166.pdf and it is defined as `area / diameter^2`

Sample code for this feature:

``` python
def shape_factor(pts):
    """Computes the area over diameter square."""
    
    def diameter_max(pts):
        """Maximum distance between points.
            This feature can be similar to maximum extent so it could be checked if this can be used instead.
        """
        dist = scipy.spatial.distance.cdist(pts, pts)
        return np.max(dist)

    hull = nm.geom.ConvexHull(pts)
    area = hull.volume
    diam = diameter_max(pts)
    return area / np.power(diam, 2)
```

lidakanari:
For the same dataset:
[TestData.zip](https://github.com/BlueBrain/NeuroM/files/8333925/TestData.zip)

The expected shape factor is:
``` python
shape_factor(points_circle) == 0.7524316090295856
shape_factor(points_square) == 0.5138955480337417
shape_factor(points_rectangle) == 0.305260262688911
shape_factor(points_oval) == 0.3569094941740931
```

Issue #1011, opened 2022-03-23
New feature: circularity


lidakanari:
This feature needs to be implemented on the 2D projection of a morphology and it will describe the circularity in the projected plane i.e. x-y. We will therefore need to compute this in all three projections at the morphology level.

Circularity is defined as `4 * pi  * area / (perimeter)^2
`
``` python
def circularity(pts):
    """Computes circularity as 4 pi A / P^2"""
    hull = nm.geom.ConvexHull(pts)
    area = hull.volume
    perimeter = hull.area
    return (4 * np.pi * area) / (perimeter * perimeter)
```

lidakanari:
For the same dataset:
[TestData.zip](https://github.com/BlueBrain/NeuroM/files/8333925/TestData.zip)

The expected circularity is:

```python 
circularity(circle) == 0.9878344238311306
circularity(square) == 0.8166489559832344
circularity(rectangle) == 0.6109332086629591
circularity(oval) == 0.8102300985964894
```

Issue #1010, opened 2022-03-23
New feature: aspect ratio


lidakanari:
This feature needs to be implemented on the 2D projection of a morphology and it will describe the minimum over the maximum dimension in the projected plane i.e. x-y. We will therefore need to compute this in all three projections at the morphology level. 

To extract the projection points from a morphology:

``` python 
def get_points_projection(neurite, projection1='X', projection2='Y'):
    proj1 = neurite.points[:, getattr(nm.COLS, projection1)]
    proj2 = neurite.points[:, getattr(nm.COLS, projection2)]
    return np.transpose([proj1, proj2])
```

Please find some sample code for this feature:

``` python 
def aspect_ratio(pts):
    """Computes min dimension / max dimension"""
    max_dim = np.max(nm.morphmath.principal_direction_extent(pts))
    min_dim = np.min(nm.morphmath.principal_direction_extent(pts))
    return min_dim / max_dim
```

lidakanari:
For testing purposes here are a few shapes:

[TestData.zip](https://github.com/BlueBrain/NeuroM/files/8332749/TestData.zip)

The expected results for these shapes:
``` python
aspect_ratio(circle) == 0.9838348383830106
aspect_ratio(square) == 0.9799193845949096
aspect_ratio(rectangle) == 0.33658158126782983
aspect_ratio(oval) == 0.49083100011838016

```

eleftherioszisis:
Using the convex hull points to avoid the point distribution to affect the principal directions, I get the following results:
```python
In [4]: aspect_ratio(circle)
Out[4]: 0.9975026440270615

In [5]: aspect_ratio(square)
Out[5]: 0.9829836870510568

In [6]: aspect_ratio(rectangle)
Out[6]: 0.4130280313071019

In [7]: aspect_ratio(oval)
Out[7]: 0.493502615108013
```

PR #1013, opened 2022-03-24
Shape related features

eleftherioszisis:
* Moved convex_hull to morphmath and reverted to the old way of taking an obj in neurom.geom
* Added features `aspect_ratio`, `circularity`, `shape_factor`

Implements #1010 #1011 #1012

### DIFF
--- a/neurom/features/morphology.py
+++ b/neurom/features/morphology.py
@@ -650,7 +650,7 @@ def aspect_ratio(morph, neurite_type=NeuriteType.all, projection_plane="xy"):
 def circularity(morph, neurite_type=NeuriteType.all, projection_plane="xy"):
     """Calculates the circularity of the morphology points along the plane.
 
-    The circularity is defined as the 4 * pi * area of the convex hull area over its
+    The circularity is defined as the 4 * pi * area of the convex hull over its
     perimeter.
 
     Args:
--- a/neurom/features/morphology.py
+++ b/neurom/features/morphology.py
@@ -634,8 +634,7 @@ def _unique_projected_points(morph, projection_plane,  neurite_type):
 
 @feature(shape=())
 def aspect_ratio(morph, neurite_type=NeuriteType.all, projection_plane="xy"):
-    """Calculates the min/max ratio of the principal direction extents along
-    the selected plane.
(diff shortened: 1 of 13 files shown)


neurom/features/morphology.py
```diff
@@ -589,14 +589,10 @@ def volume_density(morph, neurite_type=NeuriteType.all):
...
-    def get_points(neurite):
-        return neurite.points[:, COLS.XYZ]
-
     # note: duplicate points are present but do not affect convex hull calculation
     points = [
         point
-        for point_list in iter_neurites(morph, mapfun=get_points, filt=is_type(neurite_type))
+        for point_list in iter_neurites(morph, mapfun=sf.section_points, filt=is_type(neurite_type))
         for point in point_list
     ]
```
adrien-berchet, line 597: I think this could be:
```python
points = [
    point_list.tolist()
    for point_list in iter_neurites(morph, mapfun=sf.section_points, filt=is_type(neurite_type))
]
```
Isn't it better?

neurom/features/morphology.py
```diff
@@ -608,3 +604,83 @@ def get_points(neurite):
...
+    """
+    projected_points = _unique_projected_points(morph, projection_plane, neurite_type)
+    return [] if len(projected_points) == 0 else morphmath.aspect_ratio(projected_points)
+
+
+@feature(shape=())
+def circularity(morph, neurite_type=NeuriteType.all, projection_plane="xy"):
+    """Calculates the circularity of the morphology points along the plane.
+
+    The circularity is defined as the 4 * pi * area of the convex hull area over its
```
adrien-berchet, line 653:
```suggestion
    The circularity is defined as the 4 * pi * area of the convex hull over its
```
Shouldn't this word be removed?

eleftherioszisis: yeah

neurom/features/morphology.py
eleftherioszisis: It wouldn't be a flat list of single points without the second loop.

adrien-berchet: Ah yeah sorry you want a flat list here.

adrien-berchet (approved):

PR #1018, opened 2022-04-03
Return nan if no points found in shape features

eleftherioszisis:
Given that morphology shape features return single numbers, it's more meaningful to return NaN when no points have been found for the specific type, instead of returning an empty list.


adrien-berchet (approved):
Makes more sense indeed

PR #1021, opened 2022-04-08
Allow using lists of kwargs in morph_stats features

### LINKED ISSUE #1015: Is there a way to define optional arguments from morph_stats application?
I would like to use `extract_dataframe` from `morph_stats` with the shape features, such as `aspect_ratio`. 

This function works with a config as follows:

``` python
from neurom.apps import morph_stats
full_cfg = morph_stats.full_config()
df = morph_stats.extract_dataframe(neuron, config=full_cfg)
```
How can I select the options such as `projection_plane` from the input `config`? 

I am not sure if this is possible with the current code, if not would it be possible to add this?

eleftherioszisis:
Summary:
* Added new config support for adding the same feature multiple times with different arguments (see below)
* If kwargs are populated in feature they are appended to its name (`f'{mode}_{feature_name}__{suffix}'`)
* The suffix includes each feature's kwarg present in the config, apart from `neurite_type`, e.g. `variant:branch-order__method:uylings`
* Config sanitization has become the point where all different versions are converted to the single valid one for the rest of the program.
* Made protected all the functions apart from `extract_stats`, `extract_dataframe`, and `main`.
* Sanitization has moved from main to `extract_stats` to guarantee that regardless of the entry point in the module, i.e. using any of the three functions in the previous point, the config is sanitized the same way.

A list of features looks like this:
```yaml
morphology:
  soma_radius:
    modes:
      - mean
  max_radial_distance:
    modes:
      - mean
neurite:
  section_lengths:
    modes:
      - max
      - sum
  section_volumes:
    modes:
      - sum
  section_branch_orders:
    modes:
      - max
      - raw
  segment_midpoints:
    modes:
      - max
  max_radial_distance:
    modes:
      - mean
  principal_direction_extents:
    kwargs:
      - direction: 1
      - direction: 0
    modes:
      - min
      - sum
  partition_asymmetry:
    kwargs:
        - variant: branch-order
          method: petilla
        - variant: length
          method: uylings
    modes:
        - min
        - max
neurite_type:
- AXON
- APICAL_DENDRITE
- BASAL_DENDRITE
- ALL
```
Using that `extract_stats` results to:
```python
{'all': {'max_partition_asymmetry__variant:branch-order__method:petilla': 0.9,
         'max_partition_asymmetry__variant:length__method:uylings': 0.8795344229855895,
         'max_section_branch_orders': 10,
         'max_section_lengths': 11.758282,
         'max_segment_midpoints_0': 64.40167236328125,
         'max_segment_midpoints_1': 48.48197937011719,
         'max_segment_midpoints_2': 53.750946044921875,
         'mean_max_radial_distance': 99.58946,
         'min_partition_asymmetry__variant:branch-order__method:petilla': 0.0,
         'min_partition_asymmetry__variant:length__method:uylings': 0.00030289197373727377,
         'min_principal_direction_extents__direction:0': 80.01260723801528,
         'min_principal_direction_extents__direction:1': 5.274489134421081,
         'raw_section_branch_orders': [0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7,
                                       7, 8, 8, 9, 9, 10, 10, 0, 1, 1, 2, 2, 3,
                                       3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 8, 9, 9,
                                       10, 10, 0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5,
                                       6, 6, 7, 7, 8, 8, 9, 9, 10, 10, 0, 1, 1,
                                       2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 8,
                                       9, 9, 10, 10],
         'sum_principal_direction_extents__direction:0': 356.3872399244001,
         'sum_principal_direction_extents__direction:1': 25.26438016226154,
         'sum_section_lengths': 840.68524,
         'sum_section_volumes': 1104.9077745403258},
...
 'morphology': {'mean_max_radial_distance': 99.58946,
                'mean_soma_radius': 0.13065629977308288}}


```

Notice how the multiple feature entries with different kwargs result into different name:
```
 min_principal_direction_extents__direction:1
```
Closes #1015

### DIFF
--- a/neurom/apps/morph_stats.py
+++ b/neurom/apps/morph_stats.py
@@ -194,27 +194,34 @@ def extract_stats(morphs, config):
 
     stats = defaultdict(dict)
     for category in ("neurite", "morphology", "population"):
-        for feature_name, opts in config[category]:
+        for feature_name, opts in config[category].items():
 
-            kwargs = deepcopy(opts["kwargs"])
+            list_of_kwargs = opts["kwargs"]
             modes = opts["modes"]
 
-            if category == 'neurite':
+            for feature_kwargs in list_of_kwargs:
 
-                neurite_types = (
-                    neurite_types
-                    if 'neurite_type' not in kwargs and neurite_types
-                    else [_NEURITE_MAP[kwargs.get('neurite_type', 'ALL')]]
-                )
+                if category == 'neurite':
 
-                for neurite_type in neurite_types:
-                    kwargs["neurite_type"] = neurite_type
-                    stats[neurite_type.name].update(
-                        _get_feature_stats(feature_name, morphs, modes, kwargs)
+                    # mutated below, need a copy
+                    feature_kwargs = deepcopy(feature_kwargs)
+
+                    types = (
+                        neurite_types
+                        if 'neurite_type' not in feature_kwargs and neurite_types
+                        else [_NEURITE_MAP[feature_kwargs.get('neurite_type', 'ALL')]]
                     )
 
-            else:
-                stats[category].update(_get_feature_stats(feature_name, morphs, modes, kwargs))
+                    for neurite_type in types:
+                        feature_kwargs["neurite_type"] = neurite_type
+                        stats[neurite_type.name].update(
+                            _get_feature_stats(feature_name, morphs, modes, feature_kwargs)
+                        )
+
+                else:
+                    stats[category].update(
+                        _get_feature_stats(feature_name, morphs, modes, feature_kwargs)
+                    )
 
     return dict(stats)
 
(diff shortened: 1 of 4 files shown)


neurom/apps/morph_stats.py
```diff
@@ -76,9 +74,11 @@ def extract_dataframe(morphs, config, n_workers=1):
...
             - neurite_type: a list of neurite types for which features are extracted
               If not provided, all neurite_type will be used
-            - neurite: a dictionary {{neurite_feature: mode}} where:
-                - neurite_feature is a string from NEURITEFEATURES or NEURONFEATURES
-                - mode is an aggregation operation provided as a string such as:
+            - neurite:
+                Either a list of features: [feature_name, {kwargs: {}, modes: []}] or
+                a dictionary of features {feature_name: {kwargs: {}, modes: []}}.
+                - kwargs is an optional entry allowing to pass kwargs to the feature function
+                - modes is an aggregation operation provided as a string such as:
```
adrien-berchet, line 81: Do we want to keep both formats? Shouldn't we raise a deprecation warning for the dict format?

neurom/apps/morph_stats.py
```diff
@@ -241,23 +260,38 @@ def full_config():
...
+
+        return {"kwargs": options.get("kwargs", {}), "modes": options.get("modes", [])}
+
+    # either a dict or an iterable of feature_name, options pairs
+    pairs = category_features.items() if isinstance(category_features, dict) else category_features
+
+    return [[name, standardize_options(options)] for name, options in pairs]
+
+
+def _sanitize_config(config):
```
adrien-berchet, line 286: I am just wondering if we shouldn't perform a validation step on the config, so we can report issues accurately about why the config is invalid instead of just failing later in the code with a KeyError or something like that. It can come in another PR though.

adrien-berchet (commented):
Nice, thanks for this!

neurom/apps/morph_stats.py
eleftherioszisis: Do you mean using a schema? That's a good idea indeed, but better do it on a follow up PR yeah.

neurom/apps/morph_stats.py
eleftherioszisis: Not sure where we stand wrt to deprecating the old formats. @mgeplf , should I add warnings for all previous config layouts?

neurom/apps/morph_stats.py
adrien-berchet: Yes that's what I meant but let's do this in another PR

neurom/apps/morph_stats.py
mgeplf: Hmm, the way I was thinking about adding this was to make, the `kwargs` (which was poorly named), a list:
```
neurite:
    partition_asymmetry:
        kwargs:
           - {}
           - { variant: 'length',  method: 'petilla' }
```
and then `isinstance` kwargs to see if it was a dict (and then use the old behavior), or a list(and then use the new behavior.  I think the config could be more minimal this way; however, the modes would be the same for both `kwargs` variants.

eleftherioszisis: I like this. I will change it accordingly!

eleftherioszisis: I have implemented @mgeplf suggestion for multiple kwargs. @lidakanari , @adrien-berchet could you please try it and let me know if it works for you?Thanks.

mgeplf: LGTM; but let's wait for the feedback on usage.

lidakanari:
Yep, that works well for me, thanks!

lidakanari (approved):

adrien-berchet:
Looks perfect, thank you very much!

Issue #1022, opened 2022-04-11
New feature: percentage of neuronal mass with respect to soma


lidakanari:
We need a feature to measure the percentage of mass "above" and "below" the soma. The direction is determined based on the selected axis (default='Y'). 

A draft for this feature can be found here: 
``` python
from neurom.core.morphology import iter_neurites
from neurom.core.types import tree_type_checker as is_type
from neurom.features import section as sf


def get_mass_above_soma(neuron, neurite_type, axis='Y'):
    """Computes the percentage of points above / below the soma
       with respect to the selected axis.
    """
    col = getattr(neurom.COLS, axis)

    points = list(
            iter_neurites(neuron, mapfun=sf.section_points, filt=is_type(neurite_type))
        )
    point_sel = np.vstack(points)[:, col]
    soma_center = neuron.soma.center
    return 100 * np.count_nonzero(point_sel > soma_center[col]) / len(point_sel)

```

Example values for a test neuron

``` python
neuron = neurom.load_morphology('./tests/data/neurolucida/bio_neuron-000.asc')

get_mass_above_soma(neuron, neurite_type=neurom.AXON, axis='Y') == 73.25581395348837
get_mass_above_soma(neuron, neurite_type=neurom.AXON, axis='X') == 56.60377358490566
get_mass_above_soma(neuron, neurite_type=neurom.AXON, axis='Z') == 26.39315489249671

get_mass_above_soma(neuron, neurite_type=neurom.BASAL_DENDRITE, axis='Z') == 48.285198555956676
get_mass_above_soma(neuron, neurite_type=neurom.BASAL_DENDRITE, axis='X') == 50.90252707581227
get_mass_above_soma(neuron, neurite_type=neurom.BASAL_DENDRITE, axis='Y') == 62.99638989169675
```

adrien-berchet:
If you want to estimate the mass, shouldn't you sum up the lengths (volumes would be even better) of the segments above or below the soma? The number of points is a good approximation only if the morphology was resampled before calling this function.

lidakanari:
Yes, you are right @adrien-berchet, computing the length percentages would actually be better. @eleftherioszisis can you easily implement this with lengths?

lidakanari:
One more comment, I'm not sure if this should be 100% maybe just leave the value without the 100. Thanks!

PR #1023, opened 2022-04-13
Implement new feature of length above soma

eleftherioszisis:
New feature for calculating the fraction of neurite length above the soma, where the up direction is determined by the `up` argument.

Implements #1022

### DIFF
--- a/neurom/features/morphology.py
+++ b/neurom/features/morphology.py
@@ -684,3 +684,41 @@ def shape_factor(morph, neurite_type=NeuriteType.all, projection_plane="xy"):
     """
     projected_points = _unique_projected_points(morph, projection_plane, neurite_type)
     return np.nan if len(projected_points) == 0 else morphmath.shape_factor(projected_points)
+
+
+@feature(shape=())
+def length_fraction_above_soma(morph, neurite_type=NeuriteType.all, axis="Y"):
+    """Returns the length fraction of the segments that have their midpoints higher than the soma.
+
+    Args:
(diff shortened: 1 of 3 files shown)


neurom/features/morphology.py
```diff
@@ -684,3 +684,41 @@ def shape_factor(morph, neurite_type=NeuriteType.all, projection_plane="xy"):
     """
     projected_points = _unique_projected_points(morph, projection_plane, neurite_type)
     return np.nan if len(projected_points) == 0 else morphmath.shape_factor(projected_points)
+
+
+@feature(shape=())
+def length_fraction_above_soma(morph, neurite_type=NeuriteType.all, up="Y"):
+    """Returns the length fraction of the segments that have their midpoints higher than the soma.
```
adrien-berchet, line 691: What is the expected accuracy of this feature? Is it relevant to split the segments that cross the 0 of the considered axis to improve accuracy?

adrien-berchet (approved):
LGTM, just a minor question

adrien-berchet:
Ah and just another thing, could a user need this feature using an arbitrary axis?

neurom/features/morphology.py
adrien-berchet: Ok! :-)

lidakanari (approved):
LGTM, spliting segments is not required and an arbitrary axis can be added if needed in the future. For the current needs this should be sufficient imo.

adrien-berchet (approved):

PR #1025, opened 2022-04-25
Make fork iterators homogeneous

eleftherioszisis:


mgeplf:
only the `tests/test_mixed.py` tests fail, does that mean this is backwards compat?

eleftherioszisis:
> only the `tests/test_mixed.py` tests fail, does that mean this is backwards compat?

Nope. The test_mixed.py tests both subtree enabled and disabled feature calculation. 

However, it is true that for a morphology with homogeneous neurites, there would be no backward break because of these iterators. Only in the case of artifacts, e.g. a mistyped section, the homogeneous fork iterator would skip the forking point.

eleftherioszisis:
Closing for now. We may revisit it in the future.

PR #1026, opened 2022-04-26
Remove deprecations

eleftherioszisis:


eleftherioszisis:
@mgeplf I think the only think left here is to decide if I should remove the viewer module. I am hesitant because the old way was more intuitive for the users compared to using the backend modules directly.

mgeplf:
>  I am hesitant because the old way was more intuitive for the users compared to using the backend modules directly.

Yeah, I think I agree about intuitive-ness.
However, in some cases, the plotly backend is nice; I wonder if we can just make it cleaner to select a different one?  I don't have a proposal, at the moment, though.

eleftherioszisis:
> > I am hesitant because the old way was more intuitive for the users compared to using the backend modules directly.
> 
> Yeah, I think I agree about intuitive-ness. However, in some cases, the plotly backend is nice; I wonder if we can just make it cleaner to select a different one? I don't have a proposal, at the moment, though.

ok, so let's not include the viewer related stuff in this PR and I will open a follow up one for the viewer specific stuff. How about that? I will take a look at the viewer module and check if we can have a unified api with multiple backed. I am afraid that it might not be trivial given that it has not been done already.

mgeplf (approved):

PR #1032, opened 2022-04-29
Remove deprecated viewer module

eleftherioszisis:


neurom/view/__init__.py
```diff
@@ -28,6 +28,4 @@
 
 """View tools to visualize morphologies."""
 from neurom.view.matplotlib_impl import (plot_morph, plot_morph3d,
-                                         plot_tree, plot_tree3d,
```
eleftherioszisis, line 31: @mgeplf , just to verify after our discussion, we do want to hide these. right?

mgeplf: I don't really see their utility, but on the other hand, they're part of the API, and someone may have used them.

I guess we have 3 choices:
1) Leave things how they are
2) Mark them as deprecated, and remove them later
3) Remove them now, explain in Changelog that they just aren't as visible, and to give feedback if they are needed.

I think 3 is a fine compromise, what do you think?

eleftherioszisis: yeah, I agree

eleftherioszisis: I believe this is good to go now, right?

mgeplf (approved):

PR #1034, opened 2022-05-09
Expose subtree processing from the morph_stats api

eleftherioszisis:
Allow to process heterogeneous subtrees for morph_stats analysis.

Examples usage:
```python
neurom.apps.morph_stats.extract_dataframe(moprh, config, use_subtrees=True)
neurom.apps.morph_stats.extract_stats(moprh, config, use_subtrees=True)
```
#975

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,7 @@ Changelog
 Version 4.0.0
 -------------
 
+- Mixed subtree processing can be used in morph_stats app via the use_subtrees flag.
 - ``neurom.view.[plot_tree|plot_tree3d|plot_soma|plot_soma3D]`` were hidden from the
   neurom.view module. They can still be imported from neurom.view.matplotlib_impl.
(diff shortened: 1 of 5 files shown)


adrien-berchet:
LGTM but why is the default set to `False`? I think the results from the old behavior are a bit hard to understand for a user who probably expects the results given by `use_subtrees=True` by default, WDT?

eleftherioszisis:
> LGTM but why is the default set to `False`? I think the results from the old behavior are a bit hard to understand for a user who probably expects the results given by `use_subtrees=True` by default, WDT?

Default is True because this is the normal behavior as one would expect using entire trees of single types, i.e. the legacy behavior. I don't think we should break backward compatibility given that it is not necessary here.

eleftherioszisis:
@adrien-berchet may I merge this?

adrien-berchet:
Yep, let's keep this default value

PR #1035, opened 2022-05-10
Make documentation/docstrings testable

eleftherioszisis:
* Turn most of code snippets into testable code.
* sphinx doctest checks docstrings and documentation for testable code


doc/source/features.rst
```diff
@@ -48,18 +48,25 @@ only to a morphology or a morphology population.
...
+    m = load_morphology(morphology_path)
+
+    # valid input
+    rd = max_radial_distance(m.neurites[0])
+
+    # invalid input
+    try:
+        rd = max_radial_distance(m)
+    except Exception as e:
+        pass
```
eleftherioszisis, line 65: Ugly, but I couldn't find a nice way to make this work.

doc/source/migration.rst
```diff
@@ -56,12 +56,13 @@ Migration to v3 version
       fig, ax = matplotlib_utils.get_figure()
       matplotlib_impl.plot_dendrogram(m, ax)
       matplotlib_utils.plot_style(fig=fig, ax=ax)
-      matplotlib_utils.save_plot(fig=fig, output_path=output_path)
+      matplotlib_utils.save_plot(fig=fig, output_path="output-directory-path")
```
eleftherioszisis, line 59: Any ideas on how to nicely prevent this directory creation on the root level after doctest runs?

Using tempfile here would make this example more obscure, wouldn't it?

doc/source/features.rst
GianlucaFicarelli: Why do we need to show an example with invalid input?
If we really need to keep it, we could comment it out with an explanation, or use the `contextlib.suppress` context manager

tox.ini
```diff
@@ -14,7 +15,9 @@ envlist =
 [testenv]
 deps = {[base]testdeps}
 extras = plotly
-commands = pytest {posargs}
+commands =
+    sphinx-build -b doctest -d {toxinidir}/doc/build/doctrees {toxinidir}/doc/source {toxinidir}/doc/build/doctest
```
GianlucaFicarelli, line 19: I would keep the sphinx related stuff in pyXX-docs, using `make doctest`.
See for example https://github.com/BlueBrain/atlas-alignment/blob/307aa84ba7ae21be8b9b28bc6f3341226c12867e/tox.ini#L51-L53 or other projects with a similar approach.

eleftherioszisis: The issue with that is that it would test the code in the docs only for python3.8 rather than for all supported pythons. This is the main reason for which I think docs and doctest should not be in the same job.

doc/source/features.rst
eleftherioszisis: I like your thinking. I will comment it out!
The why I guess is to show what you can and cannot do with the features imported from different levels (neurite, morphology, population).

doc/source/migration.rst
eleftherioszisis: I could comment this out as well.

tox.ini
GianlucaFicarelli: I still think that proper unit tests should go under tests/ and tested as usual, while code in the docs should not be part of this job.
It's good to execute the code in the docs to be sure that the documentation is up to date / not broken, but testing with only one version of python is enough IMHO.
If there is code in the docs that's not part of the unit tests, it should be added (and in this way it's considered for the coverage as well).

GianlucaFicarelli: I don't know the original reason, but I guess it can be to keep tox.ini simpler, keep the sphinx logic separated, or for some historical reasons. In general, putting complex commands in tox.ini doesn't work very well.

tox.ini
```diff
@@ -42,15 +41,13 @@ commands =
...
 extras = docs
 commands =
-    # remove autosummary output
-    rm -rf {toxinidir}/doc/source/_neurom_build
-    make clean
-    make html SPHINXOPTS=-W
+    # remove autosummary output and cleanup
+    rm -rf {toxinidir}/doc/source/_neurom_build {toxinidir}/doc/build/*
+    sphinx-build -W -b doctest {toxinidir}/doc/source {toxinidir}/doc/build/doctest
+    sphinx-build -W -b html {toxinidir}/doc/source {toxinidir}/doc/build/html
```
eleftherioszisis, line 49: @GianlucaFicarelli , using the cli directly solves all the path problems and removes all the make files. What do you think?

eleftherioszisis: @mgeplf, I would like your opinion too on this one. :)

mgeplf: Will do.

mgeplf: Ok, I'm happy with removing the makefile and going this way.

tox.ini
```diff
@@ -15,7 +16,8 @@ envlist =
 [testenv]
 deps = {[base]testdeps}
 extras = plotly
-commands = pytest {posargs}
+commands =
+    pytest {posargs}
 
 [testenv:lint]
 basepython=python3.7
```
GianlucaFicarelli, line 23: Unrelated to this PR, but specifying `basepython=python3.7` in `lint` and `format` seems outdated, since the CI is running `py38-lint`.

eleftherioszisis: Rebased wrt the latest changes. It executes the correct testenvs/python combos now.

tox.ini
```diff
@@ -3,6 +3,7 @@ name = neurom
 testdeps =
     mock
     pytest>3.0
+    sphinx
```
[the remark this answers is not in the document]
eleftherioszisis: Indeed. Removed

eleftherioszisis:
Are there any further comments on this, or may I merge it?

mgeplf:
LGTM

doc/source/developer.rst
```diff
@@ -59,9 +59,10 @@ individually:
 
 .. code-block:: bash
 
-        $ tox -e py36-lint       # runs only pylint
-        $ tox -e py36-docs       # run only documentation check
-        $ tox -e py36            # run only the tests
+        $ tox -e py38-lint       # runs only pylint
+        $ tox -e py38-docs       # run only documentation check
```
adrien-berchet, line 63:
```suggestion
        $ tox -e py38-docs       # run only documentation check and build the doc
```

doc/source/examples.rst
```diff
@@ -35,34 +35,32 @@ Examples
     started *with the virtualenv activated*. That gives access to the ``neurom``
     installation.
 
-Fast analysis with :py:mod:`neurom`
+Analysis with :py:mod:`neurom`
 ***********************************
```
adrien-berchet, line 39:
```suggestion
******************************
```

doc/source/examples.rst
```diff
@@ -35,34 +35,32 @@ Examples
...
     >>> import neurom as nm
-    >>> m = nm.load_morphology('some/data/path/morph_file.swc')
+    >>> m = nm.load_morphology("tests/data/swc/Neuron.swc")
     >>> ap_seg_len = nm.get('segment_lengths', m, neurite_type=nm.APICAL_DENDRITE)
     >>> ax_sec_len = nm.get('section_lengths', m, neurite_type=nm.AXON)
 
 
-Morphology visualization with the :py:mod:`neurom.viewer` module
+Morphology visualization with the :py:mod:`neurom.view` module
 ****************************************************************
```
adrien-berchet, line 52:
```suggestion
**************************************************************
```

doc/source/examples.rst
```diff
@@ -35,34 +35,32 @@ Examples
...
-    >>> from neurom import viewer
-    >>> fig, ax = viewer.draw(m)
-    >>> fig.show()
-    >>>
-    >>> fig, ax = viewer.draw(m, mode='3d') # valid modes '2d', '3d', 'dendrogram'
-    >>> fig.show()
+    >>> from neurom.view import plot_morph, plot_morph3d, plot_dendrogram
+    >>> plot_morph(m)
+    >>> plot_morph3d(m)
+    >>> plot_dendrogram(m)
```
adrien-berchet, line 63: Just to understand: how this does snippet load the `m` object? Is it because all the tests of a document are executed in the same session?

doc/source/features.rst
```diff
@@ -48,18 +48,22 @@ only to a morphology or a morphology population.
...
+    from neurom import load_morphology, features
+    from neurom.features.neurite import max_radial_distance
+
+    m = load_morphology("tests/data/swc/Neuron.swc")
+
+    # valid input
+    rd = max_radial_distance(m.neurites[0])
+
+    # invalid input
+    # rd = max_radial_distance(m)
```
adrien-berchet, line 62: As far as I can see [here](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testoutput) and [here](https://stackoverflow.com/a/12609/3868743), it is possible to check and show the exception that is raised by the invalid input.

doc/source/examples.rst
eleftherioszisis: Tests are grouped by the 'group' label, in this case, it's the `[examples]` group that forces all the snippets with that label to be joined together.

doc/source/features.rst
eleftherioszisis: I would need to separate this line into a different snippet to capture that cleanly, and I don't think it is worth it tbh.

doc/source/examples.rst
adrien-berchet: Ah ok, nice.
Thanks

doc/source/features.rst
adrien-berchet: I can understand, as you prefer 😁

eleftherioszisis: Let's leave it as it is for the time being. :D

adrien-berchet (approved):

PR #1037, opened 2022-05-16
Add build test for mybinder

eleftherioszisis:


.github/workflows/mybinder-build.yml
```diff
@@ -0,0 +1,21 @@
+name: Trigger MyBinder build
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+    #tags:
+    #  - '[0-9]+.[0-9]+.[0-9]+'
```
eleftherioszisis, line 9: Once the review is done, I will change this into executing only on merging so that we don't spam binder servers.

adrien-berchet (approved):
LGTM

eleftherioszisis:
Well I am not sure I should merge this given that the job always passes regardless if the build fails or not. :(

PR #1038, opened 2022-05-18
Remove pyXX prefix for lint, docs, and coverage

eleftherioszisis:


tox.ini
```diff
@@ -7,16 +7,19 @@ testdeps =
...
 envlist =
     py{37,38,39,310}
-    py38-lint
-    py38-coverage
-    py38-docs
+    lint
+    coverage
+    docs
+
+ignore_basepython_conflict = true
```
adrien-berchet, line 14: Is it really needed?

eleftherioszisis: Yeah, otherwise when you run it locally you get warnings such as:
```python
/home/zisis/.local/lib/python3.6/site-packages/tox/config/__init__.py:593: UserWarning: conflicting basepython version (set 3.8, should be 3.10) for env 'py310';resolve conflict or set ignore_basepython_conflict
  proposed_version, implied_version, testenv_config.envname
/home/zisis/.local/lib/python3.6/site-packages/tox/config/__init__.py:593: UserWarning: conflicting basepython version (set 3.8, should be 3.7) for env 'py37';resolve conflict or set ignore_basepython_conflict
  proposed_version, implied_version, testenv_config.envname
/home/zisis/.local/lib/python3.6/site-packages/tox/config/__init__.py:593: UserWarning: conflicting basepython version (set 3.8, should be 3.9) for env 'py39';resolve conflict or set ignore_basepython_conflict

```

eleftherioszisis: I can specify however basepython for each lint, docs, coverage instead of putting it in the parent testenv.

adrien-berchet: Yeah it is more explicit like this.

tox.ini
```diff
@@ -7,16 +7,17 @@ testdeps =
...
+    docs
 
 [testenv]
 deps = {[base]testdeps}
 extras = plotly
 commands = pytest {posargs}
 
-[testenv:py38-lint]
+[testenv:lint]
+basepython=python3.8
```
adrien-berchet, line 20: I think it would be better to use py37 for linting, so we are sure it will catch everything (if we use a py38 feature it will not complain but it will fail with py37) and the CI will be faster since all the jobs will not execute in the same environment.

eleftherioszisis: I thought about that but one would need most interpreters to run tox locally.

adrien-berchet: It's easy to install ;-)

eleftherioszisis: Not on mac :P

eleftherioszisis: Nah, you can install all of them, it is just they are not on your path out-of-the-box. I will change lint to py37 and I will deal with the mac side.

eleftherioszisis: Done

adrien-berchet: Cool, thanks!

adrien-berchet (approved):

PR #1039, opened 2022-05-19
Fix tutorials and add check-tutorial testenv

eleftherioszisis:


adrien-berchet (approved):
Just to be sure I understand what happens: it just executes all the cells of the notebook to check they all succeed but it does not check anything on the results, right?

eleftherioszisis:
> Just to be sure I understand what happens: it just executes all the cells of the notebook to check they all succeed but it does not check anything on the results, right?

Yep, exactly. Just runs them.

adrien-berchet:
> Yep, exactly. Just runs them.

Ok, thanks!

PR #1040, opened 2022-05-20
Add isort for formatting/linting

eleftherioszisis:


pyproject.toml
```diff
@@ -9,3 +9,7 @@ build-backend = "setuptools.build_meta"
 testpaths = [
     "tests",
 ]
+
+[isort]
```
adrien-berchet, line 13: You could consider adding `force_single_line = true` here. As far as I am concerned I think it looks better and makes imports easier to read. But as you prefer.

eleftherioszisis: Sure, added it.

adrien-berchet (approved):

PR #1042, opened 2022-05-24
Add black to neurom, format everything, and add to lint

eleftherioszisis:


tox.ini
```diff
@@ -35,8 +37,10 @@ commands =
 basepython=python3.7
```
adrien-berchet, line 37: You could add `skip_install = true` in this section.

mgeplf:
GIthub turned on handling of `.git-blame-ignore-revs` (https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/), so we'll add the commit there, once it has a stable hash

mgeplf (approved):

mgeplf (changes requested):
Let's allow single quotes w/ `--skip-string-normalization`, and rerun the format.

eleftherioszisis:
@mgeplf is it good now?

mgeplf (approved):

PR #1049, opened 2022-07-04
allow Morphology objects to be either mut or immut

mgeplf:


neurom/core/morphology.py
```diff
@@ -559,13 +564,24 @@ def points(self):
...
-        mut.soma.points = trans(mut.soma.points)
+        morph = self._morphio_morph
+
+        is_immutable = hasattr(morph, 'as_mutable')
+
+        # make copy or convert to mutable if immutable
+        if is_immutable:
+            morph = morph.as_mutable()
+        else:
+            morph = morphio.mut.Morphology(morph)
```
adrien-berchet, line 575:
```suggestion
        # make copy or convert to mutable if immutable
        try:
            morph = morph.as_mutable()
            is_immutable = True
        except AttributeError:
            morph = morphio.mut.Morphology(morph)
            is_immutable = False
```
What about something like this?

adrien-berchet: As you prefer, it was just a suggestion. Generally speaking, try/except blocks are more Pythonic and thus more efficient because we don't have to check for anything, we just have to 'try'. In this case the `transform` function is not supposed to be called many times so it this kind of optimization does not matter.

eleftherioszisis (approved):

eleftherioszisis:
@mgeplf , may I merge it?

PR #1050, opened 2022-08-18
Feature cache

adrien-berchet:
# Issue
Sometimes we want to compute features on different sets of neurites or morphologies. While it is possible to compute all the features we need, store them and then compute the combinations we need, it can be more convenient for the user to just compute the features on each set and don't bother we any cache mechanism.

# Proposed solution
Provide a cache mechanism in NeuroM.

### DIFF
--- a/neurom/features/__init__.py
+++ b/neurom/features/__init__.py
@@ -144,7 +144,7 @@ def get(feature_name, obj, cache=False, **kwargs):
     Arguments:
         feature_name(string): feature to extract
         obj: a morphology, a morphology population or a neurite tree
-        clear_cache (bool): If True, the cache of this feature is cleared before computation.
+        cache (bool): If True, the cache of this feature is used for computation.
         kwargs: parameters to forward to underlying worker functions
 
     Returns:
--- a/neurom/features/__init__.py
+++ b/neurom/features/__init__.py
@@ -74,7 +74,7 @@ def _get_neurites_feature_value(feature_, obj, neurite_filter, kwargs):
                   0 if feature_.shape == () else [])
 
 
-def _get_feature_value_and_func(feature_name, obj, **kwargs):
+def _get_feature_value_and_func(feature_name, obj, cache=False, **kwargs):
     """Obtain a feature from a set of morphology objects.
 
     Arguments:
@@ -100,7 +100,7 @@ def _get_feature_value_and_func(feature_name, obj, **kwargs):
         if feature_name in _NEURITE_FEATURES:
             assert 'neurite_type' not in kwargs, 'Cant apply "neurite_type" arg to a neurite with' \
                                                  ' a neurite feature'
-            feature_ = _NEURITE_FEATURES[feature_name]
+            feature_ = _NEURITE_FEATURES[feature_name][cache]
             if isinstance(obj, Neurite):
                 res = feature_(obj, **kwargs)
             else:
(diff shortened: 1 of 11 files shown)


neurom/features/__init__.py
```diff
@@ -142,12 +144,13 @@ def get(feature_name, obj, **kwargs):
     Arguments:
         feature_name(string): feature to extract
         obj: a morphology, a morphology population or a neurite tree
+        clear_cache (bool): If True, the cache of this feature is cleared before computation.
```
arnaudon, line 147: arg is `cache`, not `clear_cache` ?

adrien-berchet: Ups yes indeed, I changed the code but not the docstring...

neurom/features/section.py
```diff
@@ -41,26 +42,31 @@ def section_points(section):
...
     return sum(s.length for s in section.iupstream())
 
 
+@cached_func()
 def section_length(section):
     """Length of a section."""
     return section.length
 
 
+@cached_func()
```
adrien-berchet, line 57: I'm not sure the functions of this module should be cached since it is not possible to not use the cache, so it might lead to heavy memory consumption. Or maybe in this case we should set `maxsize` to a relevant value isntead of `maxsize=None` (i.e. 'infinite' cache).

eleftherioszisis:
Could you please describe the issue, how this feature solves it, and a summary of what is going on? Thanks :)

arnaudon:
Adrien is having fun!

adrien-berchet:
> Adrien is having fun!

Exactly! :laughing: 
This PR is a bit exploratory to see what we can do.

adrien-berchet:
Yeah probably, that's why I made it optional.

But I think that if it's possible to provide a cache mechanism that helps users 'most of the time', it can be interesting. And users with specific environments can still create their own cache using the features without the NeuroM cache.

For now I am going to let it in the current state and I will try to test it in several situations.

mgeplf:
> Yeah probably, that's why I made it optional.

Even then, it raises the question of why it is optional, which makes things more confusing to the user.

> But I think that if it's possible to provide a cache mechanism that helps users 'most of the time', it can be interesting.

"Most of the time" is exactly why it can be dangerous - it's the edge cases that will catch people.  'Caching' is labeled the one of the hardest problems in computer science for a reason.

Can you provide more examples where it would help people 'most of the time' and what the time savings are?  Most of the features are computed rather quickly, so I'd like to see the numbers on where a cache is making a large difference to the run time of a system.

The thing is that it makes things more complicated to maintain, trace through the code, etc.

eleftherioszisis:
Mike raises a very good point here that first the bottlenecks have to be determined via a thorough analysis and then seek a solution to any performance issues.

For instance, maybe caching is not the best way to go in favor of composing features to be applied to morphologies via a single pass per morphology. Or maybe the complexity of some features needs to be reworked so that it doesn't drag the rest.

arnaudon:
Just for some more context, this PR is from adrien seeing my code here: https://github.com/BlueBrain/diameter-synthesis/blob/main/diameter_synthesis/simpler_diametrizer.py where I had to to some recursive pass through neurons and compute many times section path lengths. A cache made a big speed up for me, as I had to recompute many times the same path lengths via the numpy.linalg.norm thing. This is one use case of using cache in neurom, which I did 'manually' here. Of course, there may be a smarter way to implement my algo without using a cache, but I was too lazy ;)

mgeplf:
> where I had to to some recursive pass through neurons and compute many times section path lengths. A cache made a big speed up for me, as I had to recompute many times the same path lengths via the numpy.linalg.norm thing.

That's exactly the behavior I'd expect - you have a use case where you know how to cache something, and adding a cache isn't bad for that particular path.

WRT caching, since neurom is using a mutable morphio object underneath (which is something that is/will be changed), the cache has to also check for invalidation if something has been changed - which is something that could catch someone, and why I think advanced users can add a cache to however they're calling the library, if they need one.

PR #1051, opened 2022-08-30
Fix view cli

arnaudon:
Upon using `neurom view neuron.asc` with matplotlib, we get a massive zoom at the origin. This ensures we see the entire morphology with equal axis scales.

### DIFF
--- a/neurom/apps/cli.py
+++ b/neurom/apps/cli.py
@@ -76,6 +76,7 @@ def view(input_file, is_3d, plane, backend, realistic_diameters):
 
     plot(load_morphology(input_file))
     if is_matplotlib:
+        plt.axis('equal')
         plt.show()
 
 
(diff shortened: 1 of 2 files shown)


neurom/apps/cli.py
```diff
@@ -76,6 +76,10 @@ def view(input_file, is_3d, plane, backend, realistic_diameters):
 
     plot(load_morphology(input_file))
     if is_matplotlib:
+        try:
+            plt.axis('equal')
+        except NotImplementedError:  # for python 3.7 it will not have equal scales
+            plt.axis('auto')
```
adrien-berchet, line 82: Isn't `auto` the default behavior? If it is you can pass py37 case.

arnaudon: no it's not, auto makes the plot with entire neuron, but not equal x/y scales, something weird is happening in the use of matplotlib, but in a script, I can always add this 'equal' line after a plot

adrien-berchet: Ah ok, I see.

adrien-berchet (approved):

PR #1053, opened 2022-09-19
Add path length option to sholl analysis

arnaudon:
By default the Sholl analysis is done wrt radial euclidean distance, but having path distance as an optional metric for Sholl analysis is sometimes useful. 
This PR provides such functionality via an additional parameters `distance_type`.


neurom/apps/cli.py
```diff
@@ -75,8 +75,9 @@ def view(input_file, is_3d, plane, backend, realistic_diameters):
             plot = partial(plotly_impl.plot_morph, plane=plane)
 
     plot(load_morphology(input_file))
-    if is_matplotlib and not is_3d:
-        plt.axis('equal')
+    if is_matplotlib:
+        if not is_3d:
+            plt.axis('equal')
```
eleftherioszisis, line 80: What's different here?

arnaudon: coverage was complaining (but it wasn't in the previous PR...)

eleftherioszisis: You had introduce this change to fix the view cli in a previous commit. How does this change affect that issue you had before?

arnaudon: remember we discussed on slack that the issue in other PR was not about python3.7 as I naively assumed, but because there was not such option for 3d plots. So this is the correct way to implement the fix on the previous PR. (we discussed after it was merged I think). It re-appeared here for some weird reason that the try: except: was not tested (but it was in previous PR), so coverage was not at 100%. The end result is exactly the same.

eleftherioszisis: ok cool. I wanted to make sure all is good.

eleftherioszisis:
LGTM. Maybe @andrisecker have some comments?

arnaudon:
He is trying it, once he's happy with the result, we can merge!

neurom/features/morphology.py
```diff
@@ -480,14 +483,25 @@ def sholl_crossings(morph, neurite_type=NeuriteType.all, center=None, radii=None
...
+                count += int(start_dist2 <= r2 <= end_dist2 or
+                             end_dist2 <= r2 <= start_dist2)
+
+        if distance_type == 'path':
+            for section in iter_sections(neurite):
+                base_length = sf.section_path_length(section) - section.length
+                path_dists = base_length + sf.segment_lengths(section, prepend_zero=True)
+
+                forward = np.logical_and(path_dists[:-1] <= radius, path_dists[1:] >= radius)
+                backward = np.logical_and(path_dists[:-1] >= radius, path_dists[1:] <= radius)
```
eleftherioszisis, line 503: On second thought, this doesn't make much sense, does it?

Radial distance and path distance are two incompatible metrics. However, here are compared for the Sholl crossings, because you are using the radial distance of the shell against the path distance of the segments.

You may want to refer the the sholl descriptors in the following paper:
https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1010229

at the 2.2 section, point 7

> Total Wiring. Take a tree being a connected component of a neuron that contain soma, restricted to a ball of a radius r. This construction measures the total path length of this tree. This is the sum of the path distances of all dendritic segments and it is related to the notion of Arbor Density or Arbor Territory [[43](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1010229#pcbi.1010229.ref043)]. Greater total wiring dendritic length means greater invaded space which render the dendrites available to more incoming connections and greater connectivity.

eleftherioszisis (changes requested):

neurom/features/morphology.py
arnaudon: I'm not 100% sure I follow you, but now the 'shell' with radius `radius` is in path length, so it's not a spherical shell in 3d. I think this does what is in the paper you mention, no?

arnaudon:
@andrisecker , are you good with this? Is it doing what you wanted?

neurom/features/morphology.py
eleftherioszisis: `radius` is not reflecting path length because the bins are calculated from the radial distance in the sholl frequency function:
```python
bins = np.arange(min_soma_edge, min_soma_edge + max(max_radius_per_neurite), step_size)
```
In the paper, it states indeed that you can have path distance descriptors with an independent var on the path length span L(N), and then they proceed to list the most common descriptors along with their physical interpretation. All of them use radial distance as the independent variable, including the total wiring I mentioned above that gives insight into arborization density.

Therefore, I wonder what the physical meaning of this descriptor is.

arnaudon: ah yes, good catch, I forgot to update the frequency bins, now it's consistent

arnaudon:
@andrisecker are you still using this? do you think its worth merging or not?

andrisecker:
@arnaudon I'm not using it any more, but I think if I had this use case and it's already implemented then you should merge it.

neurom/features/morphology.py
```diff
@@ -528,19 +546,34 @@ def sholl_frequency(morph, neurite_type=NeuriteType.all, step_size=10, bins=None
...
+                np.max(np.linalg.norm(n.points[:, COLS.XYZ] - morph.soma.center, axis=1))
+                for n in morph.neurites
+                if neurite_filter(n)
+            ]
+            if not max_radius_per_neurite:
+                return []
+            min_distance = morph.soma.radius
+            max_distance = min_distance + max(max_radius_per_neurite)
+
+        if distance_type == "path":
```
adrien-berchet, line 561:
```suggestion
        elif distance_type == "path":
```

arnaudon:
@eleftherioszisis  I let you double check and approve this when you have time, as I cannot merge

neurom/features/morphology.py
```diff
@@ -457,7 +458,9 @@ def neurite_volume_density(morph, neurite_type=NeuriteType.all):
 
 
 @feature(shape=(...,))
-def sholl_crossings(morph, neurite_type=NeuriteType.all, center=None, radii=None):
+def sholl_crossings(
+    morph, neurite_type=NeuriteType.all, center=None, radii=None, distance_type="euclidean"
```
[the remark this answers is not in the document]
arnaudon: done

neurom/features/morphology.py
```diff
@@ -480,14 +484,25 @@ def sholl_crossings(morph, neurite_type=NeuriteType.all, center=None, radii=None
     """
     def _count_crossings(neurite, radius):
```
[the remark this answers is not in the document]
arnaudon: I think its ok to use radius as a generic name, as it implies an underlying notion of distance, which can be whatever

arnaudon:
rebase is a pain, and this feature is not used, I discard..

PR #1054, opened 2022-09-20
Replace iter_* methods by properties in core objects and improve iter_segments

adrien-berchet:


adrien-berchet:
The coverage error will be fixed in #1053 so I will rebase afterwards

neurom/core/morphology.py
```diff
@@ -200,6 +200,10 @@ def __repr__(self):
         return (f'Section(id={self.id}, type={self.type}, n_points={len(self.points)})'
                 f'<parent: Section(id={parent_id}), nchildren: {len(self.children)}>')
 
+    def iter_segments(self):
+        """Iterator to the segments of the section."""
+        return iter_segments(self)
+
```
eleftherioszisis, line 206: Does the method add anything other than convenience? On top of that is a limited version of the function `iter_segments` as it uses the defaults. Why not just use the function instead?

neurom/core/morphology.py
```diff
@@ -386,7 +400,7 @@ def volume(self):
         """
         return sum(s.volume for s in self.iter_sections())
 
-    def iter_sections(self, order=Section.ipreorder, neurite_order=NeuriteIter.FileOrder):
```
eleftherioszisis, line 389: Please don't break the public api :)

neurom/core/morphology.py
adrien-berchet: This one is just for convenience and consistency (since for other objects we can have methods to iterate over the lower scale objects, e.g. `iter_sections()` in the `Neurite` class).

neurom/core/morphology.py
adrien-berchet: But this API makes no sense :laughing: 
`self` is a `Neurite` object so the `neurite_order` parameter is just never considered.
If this change breaks something it means that the user tries to do something very weird and the result is probably not what he has in mind. But this change can go to the v4 branch if you prefer.

neurom/core/morphology.py
eleftherioszisis: I wish I could remove all the iter methods from the objects. :D 

ok fair enough

neurom/core/morphology.py
```diff
@@ -288,7 +292,10 @@ def iter_sections(neurites,
     )
 
 
-def iter_segments(obj, neurite_filter=None, neurite_order=NeuriteIter.FileOrder):
+def iter_segments(obj,
```
eleftherioszisis, line 295: Could you please add a section_filter now that you are at it?

See the implementation on the v4 branch: 
https://github.com/BlueBrain/NeuroM/blob/41ec17f3cb93166a19294117927ee1d10e75075e/neurom/core/morphology.py#L354

Or you could add the changes directly to v4 if they are not urgent.

neurom/core/morphology.py
eleftherioszisis: Indeed, I agree it makes no sense. Let's add it on v4 yeah.

neurom/core/morphology.py
adrien-berchet: Ahah yeah, that could be another solution :)
I don't really care, I just noticed it was not consistent so I added it but we could indeed remove all the `iter_*` methods.

neurom/core/morphology.py
adrien-berchet: Ok, I am going to open a new PR on the v4 branch

eleftherioszisis: Could you keep the iterato_type instead of order for consistency with the function, please?

adrien-berchet: Ah no I can just change the target here

adrien-berchet: Btw, why is it called `iterator_type`? Wouldn't `section_order` be more relevant?

eleftherioszisis: I like both, but the reason I suggested the change is to be as consistent as possible with the existing terminology.

adrien-berchet: I agree it must be consistent, I just think that since it's a new major version we could take the opportunity to improve the API. And in this case, `iterator_type` looks a bit obscure to me while `section_order` looks more clear for the user.

eleftherioszisis:
Looks good. Would it be possible before merging to update the documentation of the changelog and migration for v4?

adrien-berchet:
Is it good like this?

eleftherioszisis (approved):

PR #1057, opened 2022-09-20
Format long lists in tests

adrien-berchet:


adrien-berchet:
Hi @eleftherioszisis 
There is nothing relevant in this PR, could we merge it please? So I can remove it from my list :smile:

PR #1059, opened 2022-09-20
Draft of composite type

eleftherioszisis:
**Issue:**
When subtrees are considered, the neurites are broken into their homogeneous subtrees, increasing the tree cardinality. This can be problematic if this data is used for synthesis for instance because an entire tree is considered a subtree that can be as small as a single section.

**Proposal**
This is a draft of adding a composite type without breaking the existing integer types of NeuriteType.

The following will hold true for neurites with composite types:
1. The `CompositeType` behaves like an integer, plus an extra '_types' attribute, and is a `NeuriteType`.
2. Equality with a single section type is determined by equality with one of its subtypes. For example, an `axon_carrying_dendrite` will pass from `axon` or `basal_dendrite` filters.
3. Features that return values per neurite will maintain the neurite cardinality.
4. Features that specify filters other than `All` will apply to the composite neurites according to (1). For example, if `number_of_sections` is called for an `axon_carrying_dendrite`, it will return a positive value for both `neurite_type=NeuriteType.axon` and `neurite_type=NeuriteType.basal_dendrite`
5. Can be assigned only if `use_subtrees` is enabled. Otherwise, the simple type of the root node will be used.

**Note**
There are sane alternatives to this implementation, but they will probably require sacrificing something, such as NeuriteType deriving from IntEnum, etc.

### DIFF
--- a/neurom/core/types.py
+++ b/neurom/core/types.py
@@ -53,9 +53,12 @@ class CompositeType(int):
     Each subtype corresponds to a homogeneous subtree in the morphology.
 
     Args:
-        value (int): The integer value 
+        value (int): The integer value.
+        types: (List[SectionType]): The list of subtypes for given type.
     """
+
     def __new__(cls, value, types):
+        """Create an int object and assign to it the `subtypes` attribute."""
         obj = super().__new__(cls, value)
         obj.subtypes = types
         return obj
@@ -84,6 +87,7 @@ class NeuriteType(IntEnum):
     )
 
     def __eq__(self, other):
+        """Logic for checking single and composite types."""
         is_self_composite = hasattr(self, "value") and isinstance(self.value, CompositeType)
         is_other_composite = hasattr(other, "value") and isinstance(other.value, CompositeType)
 
@@ -101,6 +105,10 @@ class NeuriteType(IntEnum):
         return self.value
 
     def __new__(cls, value):
+        """Create a new integer enum entry and replace its `_value_` entry with the value object.
+
+        This allows passing the CompositeType instance to self.value.
+        """
         obj = int.__new__(cls, value)
         if isinstance(value, CompositeType):
             obj._value_ = value
--- a/neurom/core/types.py
+++ b/neurom/core/types.py
@@ -28,7 +28,7 @@
 
 """Type enumerations."""
 
-from enum import IntEnum, unique
+from enum import IntEnum, unique, Enum
 
 from morphio import SectionType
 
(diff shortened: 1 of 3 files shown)


neurom/core/types.py
```diff
@@ -64,6 +81,39 @@ class NeuriteType(IntEnum):
...
+        return self.value
+
+    def __new__(cls, value):
+        """Create a new integer enum entry and replace its `_value_` entry with the value object.
+
+        This allows passing the CompositeType instance to self.value.
+        """
+        obj = int.__new__(cls, value)
+        if isinstance(value, CompositeType):
+            obj._value_ = value
```
adrien-berchet, line 114: Isn't it dangerous? Can `obj` still perform all `int` operations after this change?

eleftherioszisis: I wouldn't say it's dangerous because it's part of the documentation:
https://docs.python.org/3/library/enum.html#finer-points

And yes value is either a regular integer or a CompositeType, which is also of integer type. If `_value_` is not assigned, then it keeps only the `int(value)` instead of the CompositeType and its subtypes.

neurom/core/types.py
```diff
@@ -47,25 +47,35 @@ class NeuriteIter(OrderedEnum):
...
+                    return other.subtypes[0] in self.subtypes
+                if other.is_composite():
+                    return self.subtypes[0] in other.subtypes
+                return self.subtypes[0] == other.subtypes[0]
+            return other in self.subtypes
+
+        if isinstance(other, Subtypes):
+            return self in other.subtypes
+
+        return self == other
```
eleftherioszisis, line 74: Not super excited about this mountain range here, but neurite type is also compared with bare SectionType and sometimes with ints.

neurom/core/types.py
```diff
@@ -47,25 +47,29 @@ class NeuriteIter(OrderedEnum):
     NRN = 2
 
 
-class CompositeType(int):
-    """A composite type is a list of section types.
+class Subtypes:
 
-    Each subtype corresponds to a homogeneous subtree in the morphology.
+    def __init__(self, *subtypes):
+        self.subtypes = subtypes
```
adrien-berchet, line 53: Do we care about the subtype order? If not you could use a `set()` and simplify the `__eq__` method.

eleftherioszisis: I believe so. An axon-carrying-dendrite has to have a basal before an axon in the traversal order.

adrien-berchet: Ah ok
But then in the `__eq__` method we don't care about the order? From a biological point of view, `axon == [basal, axon]` should be considered the same as `axon == [axon, basal]`?

PR #1060, opened 2022-09-22
Fix: extract_stats could not work on single neurite

adrien-berchet:
Context: A `neurite_type` entry is automatically added in `extract_stats` while `neurom.features._get_feature_value_and_func` requires that there is no `neurite_type` entry in kwargs when the object type is `Neurite`. So these two functions are not compatible together in the case of a single `Neurite` object, this PR fixes it.

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -23,7 +23,7 @@ max-locals=15
 # Maximum number of return / yield for function / method body
 max-returns=6
 # Maximum number of branch for function / method body
-max-branchs=12
+max-branches=12
 # Maximum number of statements in function / method body
 max-statements=50
 # Maximum number of parents for a class (see R0901).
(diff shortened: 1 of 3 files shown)


pylintrc
```diff
@@ -23,7 +23,7 @@ max-locals=15
 # Maximum number of return / yield for function / method body
 max-returns=6
 # Maximum number of branch for function / method body
-max-branchs=12
+max-branches=12
```
arnaudon, line 26: this one I've seen so many times already...

adrien-berchet: Yeah, I'm was surprised it has not been fixed yet in NeuroM :thinking:

eleftherioszisis: It's fixed on v4 :)

adrien-berchet: Ahhh that's it, that's why I thought it was already fixed :)

arnaudon (approved):

adrien-berchet:
Thanks

PR #1067, opened 2023-02-20
Proposal

adrien-berchet:
This PR is just a proposal to create a new type that behaves like an integer but that can handle composite types. Let's discuss it here and then we can see how we can use it in morph filters.

### DIFF
--- a/neurom/core/morphology.py
+++ b/neurom/core/morphology.py
@@ -234,12 +234,11 @@ def _homogeneous_subtrees(neurite):
 
     homogeneous_types = [neurite.type for neurite in homogeneous_neurites]
 
-    if len(homogeneous_neurites) >= 2 and homogeneous_types != [
-        NeuriteType.basal_dendrite,
-        NeuriteType.axon,
-    ]:
+    try:
+        NeuriteType(homogeneous_types)
+    except ValueError as exc:
         warnings.warn(
-            f"{neurite} is not an axon-carrying dendrite. "
+            f"{neurite} has not a registered NeuriteType. "
             f"Subtree types found {homogeneous_types}",
             stacklevel=2,
         )
@@ -296,8 +295,12 @@ def iter_neurites(
         neurites = sorted(neurites, key=lambda neurite: NRN_ORDER.get(neurite.type, last_position))
 
     if use_subtrees:
-        for neurite in neurites:
-            neurite.enable_mixed_type()
+        neurites = flatten(
+            _homogeneous_subtrees(neurite) if neurite.is_heterogeneous() else [neurite]
+            for neurite in neurites
+        )
+        # for neurite in neurites:
+        #     neurite.enable_mixed_type()
 
     neurite_iter = iter(neurites) if filt is None else filter(filt, neurites)
 
@@ -432,7 +435,7 @@ class _NeuriteTypeMode:
         return mode
 
     def _root_node_type(self):
-        return self._neurite.morphio_root_node.type
+        return NeuriteType(self._neurite.morphio_root_node.type)
 
     def _subtrees_type(self):
         it = self._neurite.root_node.ipreorder()
@@ -442,13 +445,14 @@ class _NeuriteTypeMode:
             if section.type != section.parent.type:
                 subtree_types.append(section.to_morphio().type)
 
+        if len(subtree_types) == 1:
+            return subtree_types[0]
+        else:
+            subtree_types = list(reversed(subtree_types))
+
         return NeuriteType(subtree_types)
-        # return NeuriteType.from_list(subtree_types)
-
-        # if len(subtree_types) == 1:
-        #     return subtree_types[0]
-
-        # if subtree_types == [NeuriteType.basal_dendrite, NeuriteType.axon]:
+        # # if subtree_types == [NeuriteType.basal_dendrite, NeuriteType.axon]:
+        # if subtree_types == [NeuriteType.axon, NeuriteType.basal_dendrite]:
         #     return NeuriteType.axon_carrying_dendrite
         # else:
         #     raise TypeError()
--- a/neurom/core/morphology.py
+++ b/neurom/core/morphology.py
@@ -30,6 +30,7 @@
 
 import warnings
 from collections import deque
+from cached_property import cached_property
 from pathlib import Path
 
 import morphio
@@ -246,7 +247,7 @@ def _homogeneous_subtrees(neurite):
 
 
 def iter_neurites(
-    obj, mapfun=None, filt=None, neurite_order=NeuriteIter.FileOrder, use_subtrees=False
+    obj, mapfun=None, filt=None, neurite_order=NeuriteIter.FileOrder
 ):
     """Iterator to a neurite, morphology or morphology population.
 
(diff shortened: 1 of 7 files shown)


adrien-berchet:
So don't look much the changes in `neurom/core/morphology.py`, they are not relevant for now.

arnaudon:
what is the usecase for this? how would it work?

adrien-berchet:
Ah sorry, I didn't give any context here :innocent: 
This is just a proposal to complete #1059 in order to handle neurite trees with several section types.

arnaudon:
ah this, ok ok, I didn't really follow it, I let you guys sort it out!

neurom/core/types.py
```diff
@@ -47,48 +48,107 @@ class NeuriteIter(OrderedEnum):
     NRN = 2
 
 
-class Subtypes:
+class NeuriteSubtype(int):
+    _ALL = 32
+    _BASE = 100
+    _COMPOSITE = 99
```
eleftherioszisis, line 54: Is this needed?

neurom/core/types.py
```diff
@@ -47,48 +48,107 @@ class NeuriteIter(OrderedEnum):
...
     def is_composite(self):
-        return len(self.subtypes) > 1
+        return self._value_ >= self._BASE
 
     def __eq__(self, other):
-
-        if isinstance(other, Subtypes):
+        if not isinstance(other, NeuriteSubtype):
+            other = NeuriteSubtype(other)
+        if self._ALL in self._subtypes or self._ALL in other._subtypes:
```
eleftherioszisis, line 79: _ALL should not be a subtype.

neurom/core/types.py
```diff
@@ -47,48 +48,107 @@ class NeuriteIter(OrderedEnum):
...
+class NeuriteType(NeuriteSubtype, Enum):
     """Type of neurite."""
 
     axon = SectionType.axon
     apical_dendrite = SectionType.apical_dendrite
     basal_dendrite = SectionType.basal_dendrite
     undefined = SectionType.undefined
     soma = 31
-    all = 32
+    all = NeuriteSubtype._ALL
```
eleftherioszisis, line 130: imo these two should be globals at the start of the file.

neurom/core/types.py
```diff
@@ -47,48 +48,107 @@ class NeuriteIter(OrderedEnum):
...
+        return is_eq
+
+    def __ne__(self, other):
+        return not self == other
+
+    @classmethod
+    def from_list(cls, values):
+        """Create a NeuriteType from a list of sub types."""
+        value = 0
+        for num, i in enumerate(values[::-1]):
```
eleftherioszisis, line 101: Why is the linear combination backward?

neurom/core/types.py
adrien-berchet: Could be indeed

neurom/core/types.py
adrien-berchet: To make the integer easier to read and more consistent with the list representation. So for example, `[axon, basal, apical] == [2, 3, 4] == 20304` or `[axon, soma, apical] == [2, 31, 4] == 23104`. This way we can easily represent any combination as an integer and we don't need to define all combinations (like the `axon_carrying_dendrite` you defined).

neurom/core/types.py
adrien-berchet: Ah no you're right, `_COMPOSITE` is no more used, it was just for a test.

neurom/core/types.py
adrien-berchet: I was not sure about this but I think it can be interesting because then we don't have to write special cases everywhere in the code to handle the `all` case. With this comparison it will just always be considered as equal to all other subtypes, so I think it can simplify the code of `NeuroM`. But this was just an idea, we can remove it.

neurom/core/morphology.py
```diff
@@ -439,13 +442,16 @@ def _subtrees_type(self):
             if section.type != section.parent.type:
                 subtree_types.append(section.to_morphio().type)
 
-        if len(subtree_types) == 1:
-            return subtree_types[0]
+        return NeuriteType(subtree_types)
+        # return NeuriteType.from_list(subtree_types)
```
eleftherioszisis, line 446: So the issue with this is that we want to allow specific heterogeneous types, otherwise we will have heterogeneous types for neurites that have section type artifacts.

neurom/core/types.py
```diff
@@ -47,48 +48,107 @@ class NeuriteIter(OrderedEnum):
     NRN = 2
 
 
-class Subtypes:
+class NeuriteSubtype(int):
```
eleftherioszisis, line 51: Maybe SubtypeCollection?

adrien-berchet: Can be indeed :)

neurom/core/morphology.py
adrien-berchet: Well, I think the user should be responsible of the quality of the morphologies he is analyzing. IMO we should just provide some functions to check and fix heterogeneous types, but then the analysis functions should just consider that the morphology is clean.
Nevertheless, we could issue warnings if the heterogeneous type looks weird (e.g. `[axon, soma, apical, basal]` :sweat_smile: ).

eleftherioszisis: The goal of adding heterogemenous types in NeuroM is not to expose impurities at the section level as neurite subtypes. MorphIO is much better equipped to handle that given that it is purely section-centered.

NeuroM should support valid mixed neurite types that have been found in the literature. I am not against allowing the user to specify a custom mixed type, but I believe it's not a good idea for neurom to return mixed subtypes due to impurities.

neurom/core/types.py
eleftherioszisis: I would expect to pass an ALL as a type but not as a subtype (e.g. [Axon, All]).
So instead sth like:
```python
if self._ALL in {self._value_, other._value_}:
```

adrien-berchet: So you would also forbid to create subtypes containing the `_ALL` value (e.g. `[2, 32] == [axon, all]`), right?

neurom/core/morphology.py
adrien-berchet: We can restrict the supported combinations if you wish. And thus NeuroM would just raise an exception when an unknown subtype is found?

eleftherioszisis: yeah

neurom/core/types.py
eleftherioszisis: Yes, they don't make much sense tbh because they define a group, not a single mixed type. wdyt?

adrien-berchet: Btw, the current implementation is very flexible:
```python
SubtypeCollection(3, 2, 1) == _ALL_SUBTYPE
SubtypeCollection(3, 2, 1) == SubtypeCollection(_ALL_SUBTYPE)
SubtypeCollection(3, 2, 1) == SubtypeCollection([axon, _ALL_SUBTYPE])
```

adrien-berchet: Well, I agree that defining a subtype containing the `_ALL_SUBTYPE` subtype does not make sense. But I think that forbidding such subtypes is enough, what happens in the `__eq__` operator is just an implementation detail (but the values in the `_subtypes` attributes are always integers, so it's easy to compare, while the values in the `_value_` attribute can be a `SubtypeCollection`, which makes the comparison a bit more cumbersome, though I can change this).

neurom/core/morphology.py
```diff
@@ -426,7 +426,10 @@ def set(self, mode):
         }[mode]
 
     def __call__(self):
-        return self._mode()
+        mode = self._mode()
+        # if isinstance(mode, (list, tuple)):
+        #     mode = NeuriteType.from_list(mode)
+        return mode
```
eleftherioszisis, line 432:
```suggestion
        return self._mode()
```

adrien-berchet:
Superseded by #1071

PR #1071, opened 2023-04-03
Proposal for v4 composite types

adrien-berchet:
This PR is the continuation of #1067


neurom/features/morphology.py
```diff
@@ -71,6 +71,11 @@
 feature = partial(feature, namespace=NameSpace.NEURON)
 
 
+def _return_1(*args, **kwargs):
```
eleftherioszisis, line 74: Not in favor of this. A lambda is much more readable for these trivial cases.

neurom/features/neurite.py
```diff
@@ -65,18 +65,23 @@
...
+    return 1
+
+
 def _map_sections(fun, neurite, iterator_type=Section.ipreorder, section_type=NeuriteType.all):
     """Map `fun` to all the sections."""
     check_type = is_type(section_type)
 
     def homogeneous_filter(section):
-        return check_type(section) and Section.is_homogeneous_point(section)
+        return check_type(section) and all(check_type(c) for c in section.children)
```
eleftherioszisis, line 78: `Section.is_homogeneous_point` is not sufficient here?

neurom/core/morphology.py
```diff
@@ -302,7 +300,12 @@ def iter_neurites(
 
     return (
         (
-            mapfun(neurite, section_type=filt.type if (filt is not None or filt.type is NeuriteType.all) else None)
+            mapfun(
+                neurite,
+                section_type=filt.type
+                if (filt is not None or filt.type is NeuriteType.all)
+                else None,
```
eleftherioszisis, line 307: black is not really helpful here :/

neurom/core/population.py
```diff
@@ -108,7 +112,9 @@ def neurites(self):
 
     def _load_file(self, f):
         if isinstance(f, neurom.core.morphology.Morphology):
-            return neurom.core.morphology.Morphology(f.to_morphio(), process_subtrees=self.process_subtrees)
+            return neurom.core.morphology.Morphology(
```
eleftherioszisis, line 115: Could we maybe introduce a copy() method?

neurom/features/population.py
```diff
@@ -77,11 +75,7 @@ def sholl_frequency(
     neurite_filter = is_type(neurite_type)
 
     if bins is None:
-        section_iterator = (
-            partial(iter_sections, section_filter=neurite_filter)
-            if use_subtrees
-            else partial(iter_sections, neurite_filter=neurite_filter)
-        )
+        section_iterator = partial(iter_sections, neurite_filter=neurite_filter)
```
eleftherioszisis, line 78: This doesn't look right. In the subtree case the filter does not propagate to the section level if the section_filter is not specified. Right?

tests/data/mixed/expected_morphology_features.json
```diff
@@ -358,7 +435,7 @@
         "neurite_type": 2
       },
       "expected_wout_subtrees": [],
-      "expected_with_subtrees": []
```
eleftherioszisis, line 365: Same here. There should be no trunk value for the axon.

tests/data/mixed/expected_morphology_features.json
```diff
@@ -508,7 +627,7 @@
...
```
eleftherioszisis, line 520: Same here. No trunk angles for the axon.

neurom/core/population.py
```diff
@@ -74,19 +74,22 @@ def __init__(self, files, name='Population', ignored_exceptions=(), cache=False,
...
         self._files = _resolve_if_morphology_paths(files)
 
+        self._process_subtrees = process_subtrees
+
         if cache:
-            self._files = [self._load_file(f) for f in self._files if f is not None]
+            self.reset_cache()
 
-        self._process_subtrees = process_subtrees
+    def reset_cache(self):
```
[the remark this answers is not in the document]
adrien-berchet: As it is currently implemented I don't think so. My initial thinking was that it could be used to reload the files (for example if the user know that the files changed during runtime and he wants to reload them) but once the morphs are loaded they are not reloaded, just copied. So we could force to reload the files in this function but I think it's indeed simpler to just not make it public for now.

neurom/features/morphology.py
adrien-berchet: Ah yeah, I needed it at some point because lambdas do not support `*args, **kwargs` but now it's no more needed. I will change.

neurom/features/neurite.py
adrien-berchet: Not when the user passes a composite type because the current section and its children might not have the same type since sections do not have a composite type.
Another possible implementation could be to add an optional `composite_type` argument to `Section.is_homogeneous_point`. So when it is not passed it just compares the types of the children to the type of the current section and when it is passed it compares to this passed type (and I guess we should also check that the current section is contained in this composite type, otherwise it could lead to weird results). Does this implementation makes sense to you?

neurom/core/morphology.py
adrien-berchet: :smile:

neurom/core/population.py
adrien-berchet: We could but it will need a `process_subtrees` parameter and will return a 'copy' with a different internal state so I'm not sure we should name it 'copy'. What about a method like the following?
```python
def use_process_subtrees(self, process_subtrees=False):
    return neurom.core.morphology.Morphology(
        self.to_morphio(), name=self.name, process_subtrees=process_subtrees
    )
```

eleftherioszisis: Why not copy the morphology and then change its `process_subtrees` attributes? `copy` should return an identical copy of the morphology.

tests/data/mixed/expected_morphology_features.json
```diff
@@ -310,7 +374,7 @@
...
```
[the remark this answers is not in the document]
adrien-berchet: hmmm yeah, you're right... I mixed things up, sorry. I'm going to fix this.

neurom/core/population.py
adrien-berchet: Yeah we can also do this indeed. I will add it.

neurom/features/population.py
adrien-berchet: There is something weird here indeed.
This `section_iterator` is only used to compute the bins if they are not provided by the user. So if we filter the sections here, the bins might not be consistent with the result of `neurom.features.morphology.sholl_crossings` that is called later because this function does not filter the sections.
So for now I prefer making the filters consistent and later we can add section filter if it is relevant.

neurom/core/morphology.py
```diff
@@ -295,21 +293,24 @@ def iter_neurites(
...
+    return (
+        (
+            mapfun(
+                neurite,
+                section_type=filt.type
+                if (filt is not None or filt.type is NeuriteType.all)
+                else None,
+            )
+            if neurite.process_subtrees
+            else mapfun(neurite, section_type=NeuriteType.all)
```
eleftherioszisis, line 310: Does passing the `NeuriteType.all` forces the filter to go through all the sections?

neurom/features/morphology.py
```diff
@@ -71,26 +71,48 @@
...
     )
 
 
-def _map_neurite_root_nodes(function, morph, neurite_type, use_subtrees=False):
-    neurites = iter_neurites(obj=morph, filt=is_type(neurite_type), use_subtrees=use_subtrees)
-    return [function(neurite.root_node) for neurite in neurites]
+def _map_neurite_root_nodes(function, morph, neurite_type):
+    neurite_type = SubtypeCollection(neurite_type)
+    neurites = list(iter_neurites(obj=morph, filt=is_type(neurite_type)))
+    if neurite_type != NeuriteType.all:
```
eleftherioszisis, line 87: Is it possible to check if the neurite type is a composite type instance?

neurom/core/morphology.py
```diff
@@ -453,7 +453,12 @@ def root_node(self):
 
     @cached_property
     def type(self):
-        """The type of the root node."""
+        """The type of the Neurite (which can be composite)."""
+        return NeuriteType(self.subtree_types)
+
+    @cached_property
+    def subtree_types(self):
```
eleftherioszisis, line 460: I would prefer if the subtrees were not visible from the public api. Why not get them from the type?

neurom/core/morphology.py
adrien-berchet: It should not change anything, it's just to have a consistent signature. If we don't do this we can't use lambdas as functions. But we could use inspection instead and pass a section type only when it is needed.

neurom/core/morphology.py
eleftherioszisis: I am not in favor of extending the api unless there is an essential element for the morphology/neurite. If the user wants to analyze the morphology they can do it manually. subtree processing is also not mature enough to expose it directly to the user.

adrien-berchet: What are you afraid of exactly?
Users can already access to this list of sub-tree types with `Neurite.type.to_list()`. The problem is that when a Neurite object has an unregistered composite type, it's not possible. So in this case it's hard for the users to know what composite type they should register to be able to process this Neurite object.

neurom/features/morphology.py
eleftherioszisis: Instead of doing two passes it would be cleaner if the logic determined the correct filter and then do once the iter_neurites with that filter. sth like:

```python
if neurite_type == NeuriteType.all:
    filt = is_type(neurite_type)
else:
    filt = lambda neurite: SubtypeCollection(neurite.type).root_type == neurite_type

return [function(trunk.root_node) for trunk in iter_neurites(obj=morph, filt=filt)]
```

neurom/features/neurite.py
```diff
@@ -74,8 +74,13 @@ def homogeneous_filter(section):
 
     # forking sections cannot be heterogeneous
     if (
-        iterator_type in {Section.ibifurcation_point, Section.iforking_point}
-        and section_type != NeuriteType.all
+        section_type != NeuriteType.all
+        and not SubtypeCollection(section_type).is_composite()
```
eleftherioszisis, line 78: Is this expensive to compute?

neurom/features/population.py
eleftherioszisis: Both `sholl_crossings` and `sholl_frequency` should work on the sections of the subtype in the case of subtrees. By filtering only by neurite type this is lost.

neurom/features/neurite.py
eleftherioszisis: Yeah, I was referring to the ctor. I am also wondering why do we need to check the section_type for being a composite type. It is counter-intuitive for a section type to be a composite imho.

adrien-berchet: In some cases we pass the `neurite_type` as `section_type` in the heterogeneous case of `iter_neurites`, so a composite type can arrive here. It just mean we want to filter the sections according to the subtypes of the composite type, not that the filtered sections must have a composite type. But maybe we should make it more clear in the docs?

eleftherioszisis: In the case of composites, why not pass NeuriteType.all to the section_type if the neurite_type is composite and process_subtrees=True? In the end, as far as section filtering is concerned, when a composite type matches a neurite all the sections are traversed, isn't that the case?

neurom/features/population.py
adrien-berchet: Indeed, good catch.

neurom/features/morphology.py
adrien-berchet: Indeed it's better!

neurom/features/neurite.py
adrien-berchet: The only thing we want here is to keep bifurcation points when a composite type is given with `ibifurcation_point` or
`iforking_point` iterator, so I think it's better to keep it here.

tests/conftest.py
```diff
@@ -0,0 +1,91 @@
...
+    return _load_morph_no_warning(H5_PATH / "Neuron.h5")
+
+
+@pytest.fixture
+def SWC_MORPHOLOGY_3PT(SWC_PATH):
+    return _load_morph_no_warning(SWC_PATH / 'soma' / 'three_pt_soma.swc')
+
+
+@pytest.fixture
+def MORPHOLOGY(SWC_PATH):
```
eleftherioszisis, line 68: Given that these objects make sense only in the context of specific tests, I am not fond of making them global fixtures..

adrien-berchet: There are other tests that actually load the same files, they just don't use the fixture mechanism. I don't want to change all the tests to use fixtures for now, but we should prepare for it. But if you prefer we can move these fixtures to `tests/features` for now.

neurom/core/morphology.py
eleftherioszisis: ok, let's revisit this in the future.

neurom/core/types.py
```diff
@@ -110,43 +118,38 @@ def __ne__(self, other):
...
+    n_digits = math.ceil(len(str(index)) / (len(str(base)) - 1))
+
+    int_types = np.unravel_index(index, shape=(base,) * n_digits)
+
+    if _ALL_SUBTYPE in int_types and len(int_types) > 1:
+        raise NeuroMError(
+            f"A subtype containing the value {_ALL_SUBTYPE} must contain only one element "
+            f"(current elements: {int_types})."
+        )
+    return int_types
```
adrien-berchet, line 150: I think these should be static methods of `SubtypeCollection` since they are only related to this class.

neurom/core/types.py
```diff
@@ -56,22 +59,27 @@ class SubtypeCollection(int):
...
+            value = int(value)
 
         obj = super().__new__(cls, value)
 
-        if isinstance(value, SubtypeCollection):
-            obj._subtypes = list(value._subtypes)
-        else:
-            obj._subtypes = SubtypeCollection.to_list(obj)
+        obj._subtypes = tuple(
+            SectionType(int_type) for int_type in _index_to_ids(int(obj), cls._BASE)
```
adrien-berchet, line 80: Why casting them to `SectionType`?

neurom/core/types.py
adrien-berchet, line 150: I'm not sure this implementation is better than the previous one from the `to_list()` method. Do you think it's faster? I guess `np.unravel_index` is fast but since you have to count the number of digits before calling it I'm not convinced.

neurom/core/types.py
```diff
@@ -110,43 +118,38 @@ def __ne__(self, other):
...
-        if not digits:
-            digits = [0]
-        res = digits[::-1]
-        if _ALL_SUBTYPE in res and len(res) > 1:
-            raise NeuroMError(
-                f"A subtype containing the value {_ALL_SUBTYPE} must contain only one element "
-                f"(current elements: {res})."
-            )
-        return res
-
```
adrien-berchet: I think it would be interesting to have a public API to retrieve the subtypes. Since the `_subtypes` attribute is now a Tuple (so immutable), maybe we could just rename it into `subtypes` to make it public?

neurom/core/types.py
eleftherioszisis: Because these are the subtypes. Integers are not subtypes, just a number of convention to represent them. The integers may (unlikely) but the types won't.

neurom/core/types.py
eleftherioszisis: I didn't have performance only in mind, but also maintenance. Using existing and thoroughly tested implementations is always preferable instead of rolling out our own.

neurom/core/types.py
eleftherioszisis: Sure

neurom/core/types.py
eleftherioszisis: I don't have a strong preference although I always prefer to decouple. But in this case you are right that they are not used anywhere else. I will move them inside as static methods.

neurom/core/types.py
adrien-berchet: Ok

neurom/core/types.py
adrien-berchet: Yeah but in this case you use functions that are not really made for that purpose and so you also have to add manual stuffs to compute `n_digits`, and in the end it does not always work (for example if `base < 10`, though I think we could fix this with `max(1, len(str(base)) - 1)`).

neurom/core/types.py
```diff
@@ -225,29 +225,25 @@ def _enum_accept_undefined(cls, value):
...
-            return cls._member_map_[value_str]
+    if isinstance(value, NeuriteType) and value.name in cls._member_map_:
+        return cls._member_map_[value.name]
 
     # Name given as string
-    elif isinstance(value, str):
-        if value in cls._member_map_:
-            return cls._member_map_[value]
+    if isinstance(value, str) and value in cls._member_map_:
+        return cls._member_map_[value]
```
adrien-berchet: This is not the same as the previous version. In your version the value is tested multi times while we already know the result of the if statement.
For example, if a `NeuriteType` is given but the name is not in `cls._member_map_`, then now we also check that the value is a `str,` which we already know.

eleftherioszisis: Indeed. Revert!

neurom/core/types.py
```diff
@@ -155,107 +156,141 @@ def __ne__(self, other):
...
     # Invalid value
     raise ValueError(f"{value} is not a valid registered NeuriteType")
 
 
-NeuriteType.__new__ = _enum_accept_undefined
+def _empty_init(self, *args, **kwargs):  # pylint: disable=unused-argument
+    """Replace default __init__ to do nothing more than the __new__ method."""
+
+
+class MutableEnumMeta(type):
```
eleftherioszisis, line 208: Minor: MutableEnumType would be a better choice given that EnumMeta has been renamed to EnumType.
https://github.com/python/cpython/blob/3e97c001711ab68e3d54d65e264cb5c37fefbec0/Lib/enum.py#L1046

adrien-berchet: Good point, I changed

adrien-berchet (commented):
Interesting!
So if I'm correct you discarded the following features:
1. get neurite types by name using the ctor (e.g. `NeuriteType("axon")`)
2. represent unknown types (since there is no dedicated subtype)
3. pass several types to `neurom.features.neurite._map_sections`

Did I miss anything?

tests/test_mixed.py
```diff
@@ -23,303 +25,172 @@
...
-        )
+@pytest.mark.parametrize(
+    "value, expected",
+    [
+        (0, "<NeuriteType.undefined: 0>"),
+        (32, "<NeuriteType.all: 32>"),
+        ((3, 2), "<NeuriteType.axon_carrying_dendrite: (3, 2)>"),
+    ],
+)
+def test_neurite_type__repr(value, expected):
```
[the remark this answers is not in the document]
eleftherioszisis: Not so much, no. I prefer a flat listing of the tests, but this is a matter of preference mostly.

eleftherioszisis (approved):
Let's do it.

PR #1073, opened 2023-05-15
Check that soma is not empty when features need it

### LINKED ISSUE #1072: Raise more informative error when a morph has no soma
Some morphology files have no soma, which NeuroM is able to load. But when one compute features on such morphology, one may get obscure errors. For example:
```python
nm.get("trunk_angles", neuron, neurite_type=nm.core.types.NeuriteType.basal_dendrite)
```
will raise the following error on a morphology without soma:
```python
File neurom/features/__init__.py:150, in get(feature_name, obj, **kwargs)
    135 def get(feature_name, obj, **kwargs):
    136     """Obtain a feature from a set of morphology objects.
    137 
    138     Features can be either Neurite, Morphology or Population features. For Neurite features see
   (...)
    148         List|Number: feature value as a list or a single number.
    149     """
--> 150     return _get_feature_value_and_func(feature_name, obj, **kwargs)[0]

File neurom/features/__init__.py:110, in _get_feature_value_and_func(feature_name, obj, **kwargs)
    108 if feature_name in _MORPHOLOGY_FEATURES:
    109     feature_ = _MORPHOLOGY_FEATURES[feature_name]
--> 110     res = feature_(obj, **kwargs)
    111 elif feature_name in _NEURITE_FEATURES:
    112     feature_ = _NEURITE_FEATURES[feature_name]

File neurom/features/morphology.py:202, in trunk_angles(morph, neurite_type, coords_only, sort_along, consecutive_only)
    172 @feature(shape=(...,))
    173 def trunk_angles(
    174     morph,
   (...)
    178     consecutive_only=True,
    179 ):
    180     """Calculate the angles between all the trunks of the morph.
    181 
    182     By default, the angles are defined on the x-y plane and the trees are sorted from the y axis
   (...)
    200             only the angle with the next trunk is returned for each trunk.
    201     """
--> 202     vectors = np.array(trunk_vectors(morph, neurite_type=neurite_type))
    203     # In order to avoid the failure of the process in case the neurite_type does not exist
    204     if len(vectors) == 0:

File neurom/features/morphology.py:169, in trunk_vectors(morph, neurite_type)
    166 def vector_to_root_node(neurite):
    167     return morphmath.vector(neurite.root_node.points[0], morph.soma.center)
--> 169 return _map_neurites(vector_to_root_node, morph, neurite_type)

File neurom/features/morphology.py:67, in _map_neurites(function, morph, neurite_type)
     66 def _map_neurites(function, morph, neurite_type):
---> 67     return list(
     68         iter_neurites(morph, mapfun=function, filt=is_type(neurite_type))
     69     )

File neurom/features/morphology.py:167, in trunk_vectors.<locals>.vector_to_root_node(neurite)
    166 def vector_to_root_node(neurite):
--> 167     return morphmath.vector(neurite.root_node.points[0], morph.soma.center)

File neurom/morphmath.py:60, in vector(p1, p2)
     50 def vector(p1, p2):
     51     """Compute vector between two 3D points.
     52 
     53     Args:
   (...)
     58         3-vector from p1 - p2
     59     """
---> 60     return np.subtract(p1[COLS.XYZ], p2[COLS.XYZ])

TypeError: 'NoneType' object is not subscriptable
```
which is not really helpful for users.
I suggest we add some checkers to raise more understandable errors.

adrien-berchet:
Fixes #1072

### DIFF
--- a/neurom/features/morphology.py
+++ b/neurom/features/morphology.py
@@ -63,6 +63,13 @@ from neurom.morphmath import convex_hull
 feature = partial(feature, namespace=NameSpace.NEURON)
 
 
+def _assert_soma_center(morph, feature_name):
+    if morph.soma.center is None:
+        raise NeuroMError(
+            f"The given morphology has no soma so the feature '{feature_name}' can not be computed."
+        )
+
+
 def _map_neurites(function, morph, neurite_type):
     return list(
         iter_neurites(morph, mapfun=function, filt=is_type(neurite_type))
@@ -132,6 +139,8 @@ def trunk_origin_azimuths(morph, neurite_type=NeuriteType.all):
 
     The range of the azimuth angle [-pi, pi] radians
     """
+    _assert_soma_center(morph, "trunk_origin_azimuths")
+
     def azimuth(neurite):
         """Azimuth of a neurite trunk."""
         return morphmath.azimuth_from_vector(
@@ -151,6 +160,8 @@ def trunk_origin_elevations(morph, neurite_type=NeuriteType.all):
 
     The range of the elevation angle [-pi/2, pi/2] radians
     """
+    _assert_soma_center(morph, "trunk_origin_elevations")
+
     def elevation(neurite):
         """Elevation of a section."""
         return morphmath.elevation_from_vector(
@@ -163,6 +174,8 @@ def trunk_origin_elevations(morph, neurite_type=NeuriteType.all):
 @feature(shape=(...,))
 def trunk_vectors(morph, neurite_type=NeuriteType.all):
     """Calculate the vectors between all the trunks of the morphology and the soma center."""
+    _assert_soma_center(morph, "trunk_vectors")
+
     def vector_to_root_node(neurite):
         return morphmath.vector(neurite.root_node.points[0], morph.soma.center)
 
@@ -496,6 +509,7 @@ def sholl_crossings(morph, neurite_type=NeuriteType.all, center=None, radii=None
             '`sholl_crossings` input error. If `center` or `radii` is not set then `morph` is ' \
             'expected to be an instance of Morphology and have a soma.'
         if center is None:
+            _assert_soma_center(morph, "sholl_crossings")
             center = morph.soma.center
         if radii is None:
             radii = [morph.soma.radius]
@@ -525,6 +539,7 @@ def sholl_frequency(morph, neurite_type=NeuriteType.all, step_size=10, bins=None
         If a `neurite_type` is specified and there are no trees corresponding to it, an empty
         list will be returned.
     """
+    _assert_soma_center(morph, "sholl_frequency")
     neurite_filter = is_type(neurite_type)
 
     if bins is None:
@@ -701,6 +716,7 @@ def length_fraction_above_soma(morph, neurite_type=NeuriteType.all, up="Y"):
     Returns:
         The fraction of neurite length that lies on the right of the soma along the given axis.
     """
+    _assert_soma_center(morph, "length_fraction_above_soma")
     axis = up.upper()
 
     if axis not in {"X", "Y", "Z"}:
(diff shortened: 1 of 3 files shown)


neurom/features/morphology.py
```diff
@@ -163,6 +174,8 @@ def elevation(neurite):
 @feature(shape=(...,))
 def trunk_vectors(morph, neurite_type=NeuriteType.all):
     """Calculate the vectors between all the trunks of the morphology and the soma center."""
+    _assert_soma_center(morph, "trunk_vectors")
```
eleftherioszisis, line 177: Instead of increasing the logic inside the function, the soma check could be a decorator, checking that morph has a soma and raising using the function's name if not. Would that work?

adrien-berchet: Yeah I thought about something like this but the `sholl_crossings` feature can actually work with empty soma if `center` and `radii` are provided. So a decorator would not work in all cases. But we could create a decorator for other features and just something specific for this one.

eleftherioszisis: Is it however a realistic scenario that the feature can work without a soma? Couldn't we make it strict even if it isn't always needed?

adrien-berchet: I think it is since it just counts the number of intersections between neurites and a set of circles. The centers of these circles are usually the center of the soma but I guess it could be anything, though I don't really see the point of such feature :smile: 
And in fact, this function was not a feature until `Neurom == 3`, it was just an internal function. Now that it is a feature, I think it should not accept a `center` parameter any more and it should always use the soma center. WDYT?

adrien-berchet: (ofc if we change the signature we will have to target the v4 branch)

eleftherioszisis: No, you are right. The sholl crossings can also take a list of neurites. It is fine the way you have it.

neurom/features/population.py
```diff
@@ -78,6 +79,9 @@ def sholl_frequency(morphs, neurite_type=NeuriteType.all, step_size=10, bins=Non
                         for n in m.neurites if neurite_filter(n))
         bins = np.arange(min_soma_edge, min_soma_edge + max_radii, step_size)
 
+    for morph in morphs:
+        _assert_soma_center(morph, "sholl_frequency")
```
eleftherioszisis, line 83: I would opt for making this lazy, i.e. specifying an internal func:
```python
def func(morph):
    _assert_soma_center(morph, "sholl_frequency")
   return sholl_crossings(...)
```

adrien-berchet: good idea, I changed.

neurom/features/morphology.py
adrien-berchet: ok

neurom/features/population.py
adrien-berchet: I also added the morph name in the exception

neurom/features/morphology.py
```diff
@@ -63,6 +63,19 @@
 feature = partial(feature, namespace=NameSpace.NEURON)
 
 
+def _assert_soma_center(morph, feature_name):
+    if morph.soma.center is None:
+        morph_name = getattr(morph, "name", "")
+        if not morph_name:
+            morph_name = " "
+        else:
+            morph_name = f" '{morph_name}' "
```
eleftherioszisis, line 72: neurom Morphology always has a name, doesn't it?

adrien-berchet: Yeah for the `getattr` I'm not sure, if we are sure that `morph` is a `neurom.core.Morphology` object, then we don't need the `getattr` indeed.
But for the value, it can be anything, like empty string or even None if the user assigned it. So in this case I thought we don't have to print it, but after more thoughts it can actually be an interesting information so I will always print it...

adrien-berchet: But the `name` attribute is not read-only, so it can be anything if the user wants to change it.

eleftherioszisis: If the user explicitly mutates the name with None or "", then this is what they are getting in the error. :)

eleftherioszisis (approved):

PR #1080, opened 2023-09-04
Fix neurom.app.morph_stats.extract_dataframe for Population objects with several workers

adrien-berchet:


neurom/apps/morph_stats.py
```diff
@@ -95,11 +95,15 @@ def extract_dataframe(morphs, config, n_workers=1):
     """
     if isinstance(morphs, Morphology):
         morphs = [morphs]
+    elif isinstance(morphs, Population):
+        morphs = morphs._files  # pylint: disable=protected-access
```
eleftherioszisis, line 99: Wouldn't `morphs.morphologies` do the same?

eleftherioszisis: Actually doesn't the `list(map(func, morphs))` below work automatically on the population?
https://github.com/BlueBrain/NeuroM/blob/b2c99d936fabed0fb792e644561e9ccd483d6f5a/neurom/core/population.py#L95

adrien-berchet: As far as I understand it would load the files. The point here is that we need file paths, not morphologies, because `Morphology` objects are not picklable so multiprocessing fails.

adrien-berchet: > Actually doesn't the `list(map(func, morphs))` below work automatically on the population?
> 
> https://github.com/BlueBrain/NeuroM/blob/b2c99d936fabed0fb792e644561e9ccd483d6f5a/neurom/core/population.py#L95

It works properly in the serial case but not in parallel case.

eleftherioszisis: Ah I see.

adrien-berchet: Nevertheless, we can move it to the multiproc case only if you prefer

eleftherioszisis: No, it's fine like that if it works.

adrien-berchet: Ok!

PR #1083, opened 2023-09-15
Remove a few useless shebang lines

musicinmybrain:
Remove shebang lines from a few files that do not need them.

A shebang line is only meaningful in a file with the executable bit set in its permissions. These files are already correctly non-executable, as they have no “main routine” or interesting side effects and are purely importable modules.

Slightly more detail is available in the commit messages.


mgeplf:
Looks good, thanks for the contribution.

PR #1084, opened 2023-09-15
Replace PyPI mock test dependency with unittest.mock

musicinmybrain:
The [`unittest.mock` module](https://docs.python.org/3/library/unittest.mock.html) is part of the standard library since Python 3.3, so there is no longer any benefit to carrying the PyPI package as a test dependency.

See https://fedoraproject.org/wiki/Changes/DeprecatePythonMock for further context.

### DIFF
--- a/doc/source/developer.rst
+++ b/doc/source/developer.rst
@@ -68,7 +68,7 @@ all required dependencies in your virtual environment:
 
 .. code-block:: bash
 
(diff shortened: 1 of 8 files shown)


eleftherioszisis (approved):
LGTM

musicinmybrain:
Thanks!

PR #1088, opened 2023-12-08
modernize to pyproject.toml

mgeplf:


GianlucaFicarelli (approved):

pyproject.toml
```diff
@@ -1,6 +1,64 @@
...
 requires = [
-    "setuptools>=42",
-    "wheel",
+    "setuptools>=64",
+    "setuptools_scm>=8"
 ]
 build-backend = "setuptools.build_meta"
+
+[project]
+name = "neurom"
```
eleftherioszisis, line 9: Doesn't that have to be NeuroM. It's project name, not package name. Right?

mgeplf: not sure; I've set it to `NeuroM` now.

mgeplf: Note that the original has:
``
name='neurom',
```

mgeplf: I guess we'll go back to the original

PR #1092, opened 2024-01-24
Remove pyXX prefix for lint, docs, and coverage

eleftherioszisis:
Also, incorporate coverage into tests that are already executed.


adrien-berchet:
I suggest moving the `docs` job to `py39` to make the CI faster

adrien-berchet (approved):

PR #1093, opened 2024-01-25
Fix warnings in tests

eleftherioszisis:


neurom/apps/morph_stats.py
```diff
@@ -234,7 +232,7 @@ def extract_stats(morphs, config):
     return dict(stats)
 
 
-extract_stats.__doc__ += str(EXAMPLE_CONFIG)
+extract_stats.__doc__ += str(EXAMPLE_STATS_CONFIG)
```
adrien-berchet, line 235: What is the point of giving this file path? Shouldn't we add the example config in JSON format as an example?

tests/test_import.py
```diff
@@ -29,6 +29,5 @@
 def test_import_neurom():
     try:
         import neurom
-        return True
     except Exception:
-        return False
+        print("neurom failed to be imported")
```
adrien-berchet, line 33: This test will always succeed, is it expected?

neurom/apps/morph_stats.py
eleftherioszisis: This PR was for removing the warnings, not changing the logic. I assume the json is too big to dump in the docstring.

adrien-berchet: But giving a raw path is weird IMO.
But at least the docstring of `extract_dataframe` is a bit more explicit than the one of `extract_stats`, maybe we should just make it more clear. Also, note that indent is missing so the path is not included in the Note block. Maybe it could be something like:
```python
"""
     blablabla...
     Note:
        An example config can be found in the `CLI -> neurom stats` page of the documentation or in
        the following file:
"""
extract_stats.__doc__ += " " * 4 + str(EXAMPLE_STATS_CONFIG)
```
which would give:
```pycon
>>> print(neurom.apps.morph_stats.extract_stats.__doc__)
     blablabla...
     Note:
        An example config can be found in the `CLI -> neurom stats` page of the documentation or in
        the following file:
        <absolute_path_to_NeuroM>/neurom/apps/config/morph_stats.yaml
```

And similar changes should be made for `extract_dataframe`.

adrien-berchet: But ofc this can go in another PR if you prefer

adrien-berchet (approved):

PR #1096, opened 2024-01-26
Add testing of example scripts (#1041)

eleftherioszisis:
Cherry-picked from v4 branch


tests/test_examples.py
```diff
@@ -0,0 +1,27 @@
...
+    spec = importlib.util.spec_from_file_location(filepath.stem, filepath)
+    module = spec.loader.load_module()
+
+    with tempfile.TemporaryDirectory() as tempdir:
+
+        # change directory to avoid creating files in the root folder
+        try:
+            cwd = os.getcwd()
+            os.chdir(tempdir)
+            module.main()
```
adrien-berchet, line 25: Isn't it better to call a subprocess here to be sure that it works in a clean python engine?

adrien-berchet: Forget about this :smile:

eleftherioszisis: Nah, I prefer it this way. It is also easier to debug when it fails. It is highly unlikely it won't work in a subprocess if it works like this.

adrien-berchet (approved):

tests/test_examples.py
adrien-berchet: Yeah I agree :)

PR #1111, opened 2024-04-04
V4

mgeplf:


eleftherioszisis:
The register/unregister in NeuriteType neum in this branch is quite fragile because it tries to bend the immutability of the python enums:
```python
    @classmethod
    def register(cls, name, value):
        """Register a new value in the Enum class."""
        value = _int_or_tuple(value)

        if hasattr(cls, name):
            existing = getattr(cls, name)
            raise ValueError(f"NeuriteType '{name}' is already registered as {repr(existing)}")

        try:
            existing = cls(value)
        except ValueError:
            existing = None

        if existing:
            raise ValueError(f"NeuriteType '{name}' is already registered as {repr(existing)}")

        obj = _create_neurite_type(cls, value, name=name)

        cls._value2member_map_[value] = obj
        cls._member_map_[name] = obj
        cls._member_names_.append(name)

        return obj

    @classmethod
    def unregister(cls, name):
        """Unregister a value in the Enum class."""
        if name not in cls._member_names_:
            raise ValueError(
                f"The NeuriteType '{name}' is not registered so it can not be unregistered"
            )

        value = cls._member_map_[name].value
        del cls._value2member_map_[value]
        del cls._member_map_[name]
        cls._member_names_.remove(name)
```
Therefore, it is no surprise that some of the functionality we relied upon is now broken in 3.12 because they tried to make them safer:
https://docs.python.org/3/whatsnew/3.12.html#enum

In addition, we rely on protected variables to make it work, which is also not great.

In the end, the use cases for which registration is needed are not so many. To make an int type usable, it has to also exist in MorphiIO as a SectionType and all section types are already available as neurite types.

That leaves us with registering new composite types that combine existing section types. As we don't have use cases where other type combinations are needed apart from axon-on-dendrite, which is already defined in NeuriteType, I suggest we remove the registration of neurite types and revisit them in the future when solid cases for this functionality appear.

mgeplf (commented):
I've read a lot of it, and it all looks good to me.

If converting our other tools hasn't produced any large `WAT` moments, I think we should go with this.

PR #1113, opened 2024-04-12
Add has_no_back_tracking function

adrien-berchet:
Following #1112 : add a morphology checker (return all back-tracking points of the morphology when the check fails).


eleftherioszisis (approved):

adrien-berchet:
Thanks!

PR #1114, opened 2024-04-12
Make has_no_back_tracking() compatible with generate_annotation()

adrien-berchet:
The `has_no_back_tracking()` function could not be used by the `generate_annotation()` function because the output was not formatted as expected. This PR fixes this issue.


adrien-berchet:
Thanks!

PR #1117, opened 2024-04-23
Add __version__ attribute

adrien-berchet:


neurom/__init__.py
```diff
@@ -68,3 +70,5 @@
 AXON = NeuriteType.axon
 SOMA = NeuriteType.soma
 ANY_NEURITE = NeuriteType.all
+
+__version__ = version(__package__)
```
eleftherioszisis, line 74: I think you should add that at the top before the other imports:
https://peps.python.org/pep-0008/#module-level-dunder-names

adrien-berchet: Ok, done

eleftherioszisis (approved):

PR #1118, opened 2024-04-24
Move soma methods to functions

eleftherioszisis:
* Expose soma methods as free functions that work on both morphio an neurom soma objects.
* Keep backward compatibility with the soma methods.
* Disallow storing soma values, such as radius or area, as attributes because the underlying morphio soma can be mutated.

### DIFF
--- a/neurom/core/soma.py
+++ b/neurom/core/soma.py
@@ -89,12 +89,7 @@ class Soma:
 
     def overlaps(self, points, exclude_boundary=False):
         """Check that the given points are located inside the soma."""
-        points = np.atleast_2d(np.asarray(points, dtype=np.float64))
-        if exclude_boundary:
-            mask = np.linalg.norm(points - self.center, axis=1) < self.radius
-        else:
-            mask = np.linalg.norm(points - self.center, axis=1) <= self.radius
-        return mask
+        return soma_overlaps(self, points, exclude_boundary=exclude_boundary)
 
 
 class SomaSinglePoint(Soma):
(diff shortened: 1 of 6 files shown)


neurom/core/soma.py
```diff
@@ -267,20 +202,238 @@ def __str__(self):
...
+        return 2. * math.pi * soma_radius(morphio_soma)**3
+
+
+    soma_algo = {
+        SomaType.SOMA_UNDEFINED: _soma_undefined_volume,
+        SomaType.SOMA_SINGLE_POINT: _soma_single_point_volume,
+        SomaType.SOMA_CYLINDERS: _soma_cylinders_volume,
+        SomaType.SOMA_NEUROMORPHO_THREE_POINT_CYLINDERS: _soma_three_point_cylinders_volume,
+        SomaType.SOMA_SIMPLE_CONTOUR: _soma_undefined_volume,
+    }[morphio_soma.type]
```
adrien-berchet, line 361: Isn't it better to define these functions and dicts outside the function?

neurom/core/soma.py
```diff
@@ -443,6 +363,127 @@ def _soma_three_point_cylinders_volume(morphio_soma):
     return soma_algo(morphio_soma)
 
 
+def soma_overlaps(soma, points, exclude_boundary=False):
+    """Check if soma overlaps with points."""
+
+    if hasattr(soma, "to_morphio"):
+        morphio_soma = soma.to_morphio()
+    else:
+        morphio_soma = soma
```
adrien-berchet, line 372: Maybe we could also accept morphologies as input?
```suggestion
    if hasattr(soma, "soma"):
        input_soma = soma.soma()
    else:
        input_soma = soma
    if hasattr(input_soma, "to_morphio"):
        morphio_soma = input_soma.to_morphio()
    else:
        morphio_soma = input_soma
```

eleftherioszisis: Why give that additional responsibility to this function? This is what the respective feature is for.

adrien-berchet: I think it's interesting to make our codes simpler. Especially with #1119 it possible to write something like `soma_center(morph)` which makes sense IMO.

neurom/core/soma.py
eleftherioszisis: Moved the functions outside. I however left the dicts inside.

neurom/core/soma.py
eleftherioszisis: I still don't think the soma module should be responsible for handling the morphology. Especially, when passing the soma is as simple as `soma_radius(morph.soma)`

adrien-berchet: I guess that what bothers me is that the names are over explicit. If the functions of the `soma` module can only work on soma objects, why naming the functions `soma.soma_*`? For me, if you specify the object in the name it's because it is not already explicit, which is the case if the function can accept a morphology. But if it only accepts a soma it's already explicit. For example, writing `soma.get_radius(morph.soma)` would make more sense to me than `soma.soma_radius(morph.soma)`. WDYT?

eleftherioszisis: No issue with renaming. `soma_*` changed to `get_*`

adrien-berchet: Ok!

adrien-berchet (approved):

PR #1119, opened 2024-04-25
Some proposals

adrien-berchet:


adrien-berchet:
I will fix coverage later if we agree on the rest.

neurom/core/population.py
```diff
@@ -85,6 +89,16 @@ def _reset_cache(self):
...
+    @property
+    def mutable(self):
+        """Enable mixed tree processing if set to True."""
+        return self._mutable
+
+    @mutable.setter
+    def mutable(self, value):
+        self._mutable = value
+        self._reset_cache()
+
```
eleftherioszisis, line 101: This is not a responsibility of the wrapper class. Whether the inputs are mut/immut it should only depend on how the population is initialized.

neurom/core/soma.py
```diff
@@ -203,13 +203,24 @@ def __str__(self):
...
 
-    if hasattr(soma, "to_morphio"):
-        morphio_soma = soma.to_morphio()
+    if hasattr(input_soma, "to_morphio"):
+        morphio_soma = input_soma.to_morphio()
     else:
-        morphio_soma = soma
+        morphio_soma = input_soma
+
+    return morphio_soma
```
eleftherioszisis, line 218: This is not needed. The functions work with both somas. Passing a morphology here is out of the scope of this module.

neurom/core/morphology.py
```diff
@@ -538,7 +538,7 @@ def __repr__(self):
 class Morphology:
     """Class representing a simple morphology."""
 
-    def __init__(self, filename, name=None, process_subtrees=False):
+    def __init__(self, filename, name=None, *, mutable=None, process_subtrees=False):
```
eleftherioszisis, line 541: While I can see the point of adding the option to load_morphology, adding it here it will only complicate things. Frankly, The Morphology wrapper shouldn't allow passing a file path to begin with, given that the entry point of loading a morphology are the io functions...

adrien-berchet: Indeed if we can't pass a file path it's not needed. But since it is allowed I think we should allow to choose whether the loaded morph is mutable or not.

neurom/core/soma.py
adrien-berchet: I still think it makes sense to write something like `soma_volume(morph)`, don't you? We can move the functions to another module and keep only the Soma class here if that bothers you.

neurom/core/population.py
adrien-berchet: Since we can change the `process_subtrees` status of the morphologies after the `Population` is created, I don't see why we could not also change the `mutable` status of these morphologies?

neurom/core/morphology.py
eleftherioszisis: How about cleaning it up by allowing to pass only morphio morph instances and let io deal with loading from file. This way io and wrapper will be decoupled instead of passing here options that are for io.

adrien-berchet: I think it makes sense!

eleftherioszisis (approved):

